PR #10, opened 2018-04-05
moving obsolete files to the obsolete folder

slayoo:


olebole:
Why not just remove them? They are kept in the repository.

slayoo:
I thought of removing them only after migrating any valid content into other README files, etc (which I plan to do for the documentation-related files soon)

GillesDuvert (approved):
Hi, good idea, as long as they are still available but are not uploaded by default.

Issue #1, opened 2018-04-05
automatic builds on Windows


slayoo:
This repo seems to have it already implemented by @pjb7687 : https://github.com/pjb7687/GNU-Data-Language-0.9.6

slayoo:
@pjb7687 would you have time to try to import the Windows build scripts for GDL to this repo? Thanks!

pjb7687:
@slayoo It once worked with Drone CI but unfortunately it is discontinued, so it doesn't work. Maybe I can try similar configuration with Travis CI instead. I have no idea whether it supports building in Docker container with msys32 ??

pjb7687:
I will try it in this weekend.

slayoo:
thank you!
There is also the AppVeyor platform. I do have some experience with it, but I have no experience buidling GDL in Windows, hence I didn't go too far trying with it, but in case this could be of some help, here's the appveyor.yml file I've tried with:
```
platform:
  - x86
  - x64

init:
  - ps: |
        If ($env:platform -Match "x86") {
          $env:CMAKE_GEN_PLAT="Win32"
          $env:PYTHON_ROOT="C:\Python27"
        } Else {
          $env:CMAKE_GEN_PLAT="x64"
          $env:PYTHON_ROOT="C:\Python27-x64"
        }
  - cmd: C:/"Program Files (x86)"/"Microsoft Visual Studio 12.0"/VC/vcvarsall.bat

install:
  # <TEMPORARY>
  - ps: wget https://github.com/gnudatalanguage/gdl/archive/master.zip -OutFile master.zip
  - 7z x master.zip
  # </TEMPORARY>

  #- choco upgrade cmake # so newer Boost versions are supported
  #- set path=C:\Program Files\CMake\bin;%path%

  #- ps: wget http://bitbucket.org/eigen/eigen/get/3.3.2.zip -OutFile eigen.zip
  #- 7z x eigen.zip -o"C:\projects" -y > nul

  #- set path=%PYTHON_ROOT%;%path%
  #- python -m pip install numpy

before_build:
  - del "C:\Program Files (x86)\MSBuild\14.0\Microsoft.Common.targets\ImportAfter\Xamarin.Common.targets"

build_script:
  # <TEMPORARY>
  - cd gdl-master/gdl
  # </TEMPORARY>
[...]

PR #32, opened 2018-04-07
correcting to new declaration of WShow(int, bool,int); CmakeLists.txt policy 0012 set to old to reduce warnings; Enhance obj_destroy to allow multiple parameters; Provide PathSeparator() as static routine instead of lib::

maynardGK:
(devicewin.hpp) will probably allow me to proceed with MingW compilation of git code.
CmakeLists.txt policy 0012 set to old to reduce warnings from FindopenMP.cmake (using cmake 10)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,7 @@
 # tested with 2.4 and doesn't seem to run.
 # SA: 2.6 supports simplified if constructs, e.g. endif()
 cmake_minimum_required(VERSION 2.6 FATAL_ERROR)
+cmake_policy(SET CMP0012 OLD)
 
 project(GDL)
 
@@ -28,7 +29,6 @@ include(FindPkgConfig)
 include(FindPackageHandleStandardArgs)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR}/CMakeModules)
-
 #### Default cached values
 set(PYTHON_MODULE OFF CACHE BOOL "GDL: Build GDL as a Python module ?")
 
--- a/src/basic_pro.cpp
+++ b/src/basic_pro.cpp
@@ -344,21 +344,17 @@ namespace lib {
   void obj_destroy(EnvT* e) {
     StackGuard<EnvStackT> guard(e->Interpreter()->CallStack());
 
-    int nParam = e->NParam();
-    if (nParam == 0) return;
+      int n_Param=e->NParam();
+      if( n_Param == 0) return;
 
-    BaseGDL* p = e->GetParDefined(0);
-
-    if (p->Type() != GDL_OBJ)
-      e->Throw("Parameter must be an object in"
-      " this context: " +
-      e->GetParString(0));
-    DObjGDL* op = static_cast<DObjGDL*> (p);
-
+	  for( SizeT ipar=0; ipar<n_Param; ipar++) {
+		BaseGDL*& par=e->GetPar( ipar);
+		if( par == NULL or
+			par->Type() != GDL_OBJ) continue;
+		DObjGDL* op= static_cast<DObjGDL*>(par);
     SizeT nEl = op->N_Elements();
-    for (SizeT i = 0; i < nEl; i++) {
-      DObj actID = (*op)[i];
-      e->ObjCleanup(actID);
+		for( SizeT i=0; i<nEl; i++)	
+			e->ObjCleanup( (*op)[i]);
     }
   }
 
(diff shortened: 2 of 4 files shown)


slayoo:
@maynardGK Thank you!

Shall we try to split these changes so each PR (and hence each branch) correspond to separate update (i.e. WShow vs. obj_destroy).

Why is the change to CMakeLists.txt needed?

I though of also waiting with merging any win32 #ifdefed changes till we have continuous integration set up for Windows (see https://github.com/gnudatalanguage/gdl/issues/1)

GillesDuvert:
Hello, @maynardGK ! 
concerning devicewin.cpp:
Could you try first to remove wshow from devicewin and test? Indeed, "normally", the win device "wshow" should be inherited from GraphicsMultiDevice::WShow and does not need to be overloaded locally. This to ensure a common behaviour for the 3 variants of "multi-windows", X, wxWidgets and Win.
If this is asking too much, there is no problem in accepting this patch of course.

GillesDuvert:
@slayoo 
the three patches are unrelated. I am not familiar with Git, but I suspect it would be better to make 3 successive changes.

GillesDuvert:
I forgot to mention the basic_pro.cpp patches are OK. Makes code simpler, also!

slayoo:
Then I highly recommend splitting them into three PRs... or at least doing so in the future :)

maynardGK:
Hi guys,

devicewin: I just tried it with WShow removed, window behavior is different and maybe worse
but I can't tell as the new gdlstream functions are just in and plotting is not yet fixed.

cmake_policy change is there only by accident it tagged along with devicewin.  I got a long
cmake diatribe in cmake.out about a line of code in FindOpenMP (because if(TRUE) happened) and 
this is the cause.

obj_destroy was separately committed and pushed but got merged when I did the pull request.  It seems to be automatically done .

slayoo:
Github's pull requests are updated whenever something is pushed to the branch the PR is based on.
To prevent it, and to allow separate simultaneous PRs the changes need to ge to separate branches.
HTH

GillesDuvert:
@maynardGK : I do not see why file.cpp would not continue to be the unique location where PathSeparator() is defined? It is advantageous to factorize code and maintain only at one place. As dialog.cpp include file.hpp there is no need for this patch?.

maynardGK:
Ok this is confusing.   Now I've pushed another commit on trivia branch (it is more trivia) and without further action (no PR) it has appeared in these comments (under slayoo's CMake warnings comment.) 
as if it was already appended to the pull requests. I'll let the situation settle a bit  here before pushing more trivia.  Is there a way to provide separated multiple commits in a single branch?

@Gilles PathSeparator is so simple I believe it is better kept local.

slayoo:
@maynardGK Indeed, the commit-branch-clone-fork-pull-request logic is often confusing.

Basically, the prerequisite is to have a separate branch per every feature (and separate from the master branch). The pull requests are bound to a specific branch and not a set of commits. As a consequence, if there is a correction needed to be done within a given PR, it is enough to commit something to the given branch, then push, and then the changes will appear here and any continuous integration actions will be triggered. 

On the other hand, any non-related development needs to be done in separate branches, and then a new pull request can be opened pointing to this separate branch. Note that creation of these separate branches generally needs to be done based on a "clean" master to avoid containing changes from on PR in the other from the beginning of the branch existence.

HTH

GillesDuvert:
@maynardGK : factorization is a key for simpler code maintainance. irrespective of the (at the moment) simple aspect of PathSeparator, which is factorized, what is the need to redefine it locally?

maynardGK:
@GillesDuvert : PathSeparator() will not be used by file.cpp after it has completed its modifications.
The routine is dangling there in file.cpp and its only purpose is to be there for Dialog.cpp.  
More often, file.cpp routines are  checking for either separator.  _WIN32 runs can use the forward slash as easily as the backwards - to me the forwards is preferable since the filesystems are shared with a Linux. (A backwards slash is needed for the raw CMD.exe which uses forward-slash for command switch.)

maynardGK:
@slayoo Separate branches is ok but there are many cases at the moment with a lot of
edits and some cross-fertilization.  For instance, in file.cpp my mods for file_search, file_move,
each will use a subroutine related to path search.  So I think it makes sense to incrementally
change that file in multiple commits on the same branch. 
  In a similar vein, changes to basic_pro.cpp are ready in addition to the trivia here and the heap_refcount
branch.  heap_refcount would be useful to have nailed down for testing/usage, and also to un-complicate the differencing view.  Heap_refcount is pretty self-contained and not difficult, but more complicated examples can be anticipated.

maynardGK:
I will close this pull request soon:
WShow issue will become a single branch
obj_destroy will be a commit in a series of patches for basic_pro.
The CMake change is only relevant to the use of late-model CMake (I am using 3.10 here) and
I can live without it.
Path_separator() change to dialog() will move to where file.cpp is changed. 
Probably have file.cpp host PathSeparator().

maynardGK:
Closing this PR. obj_destroy will be committed in a series of patches for basic_pro.

maynardGK:
Now I'm really closing

PR #112, opened 2018-04-09
Implement HEAP_REFCOUNT to show reference count of pointers and objects.

maynardGK:
Done here before an important change to PTR_VALID() that changes
the reference increment when a pointer is cast.  The structure
{GDL_CONTAINER} is also defined in objects.cpp, identical to IDL_CONTAINER.

GDL uses two heaps, one for pointers and one for objects.  Each time
one of these is created or needed, a reference count is incremented to
signify its relevance; when the reference count goes to zero, the variable
can/should be deleted: "automatic garbage collection".
keywords: DISABLE, ENABLE, IS_ENABLED regarding the state of garbage collecion,
either globally or per-variable basis.

Stage1 - heap_refcount
Greg Jung (aka maynardGK)
gvjung@gmail.com


slayoo:
Thank you!
Let's try to add a test routine to the same PR and include it in make-check logic before merging

slayoo:
Thanks for the test routine!
I'll create a pull request into your branch to fix the Makefile.am conflict right away.

slayoo:
Here it is: https://github.com/maynardGK/gdl/pull/1
If you merge this into your branch, this PR should contain all the changes that went already ingot master + a resolution of conflict in Makefile.am
HTH

src/GDLInterpreter.hpp
```diff
@@ -223,7 +224,8 @@ class CUSTOM_API GDLInterpreter : public antlr::TreeParser, public GDLInterprete
 //             }
 //         return false;
 //     }
-
+ //   static bool IsEnabledGC() { return enable_GC;}
+	static bool IsEnabledGC() { return true;}
```
slayoo, line 228: is this intentionally set to "return true"?

src/basic_fun.cpp
```diff
@@ -676,6 +676,117 @@ namespace lib {
...
+    else {
+      if( pType == GDL_PTR) {
+      
+	  DPtrGDL* pPtr = static_cast<DPtrGDL*>( p);
+	  for( SizeT i=0; i<nEl; ++i)
+		  (*ret)[ i] = interpreter->RefCountHeap( (*pPtr)[ i]);
+	  if( e->KeywordSet(DISABLEIx) or
+	      e->KeywordSet(ENABLEIx) ) {
+		bool set = e->KeywordSet(ENABLEIx) ? true: false;
+//	if(trace_me) cout <<" GC set? "<<set<<endl;
```
slayoo, line 732: I guess worth cleaning up?

src/basic_fun.cpp
```diff
@@ -676,6 +676,117 @@ namespace lib {
...
+	  }
+	  else
+	  {
+	    pL = static_cast<DLongGDL*>(p);
+	  }
+	  for( SizeT i=0; i<nEl; ++i)
+		  (*ret)[ i] = interpreter->RefCountHeap( (*pL)[ i]);
+	}
+      }
+// #if 1
```
slayoo, line 751: cleanup?

src/basic_fun.cpp
```diff
@@ -676,6 +676,117 @@ namespace lib {
...
+	  }
+	  for( SizeT i=0; i<nEl; ++i)
+		  (*ret)[ i] = interpreter->RefCountHeap( (*pL)[ i]);
+	}
+      }
+// #if 1 
+    if(e->KeywordPresent(IS_ENABLEDIx)) {
+      DByteGDL* enabled;
+//     	if(trace_me) 
+//	cout << "  heap_refcount( prm, KeywordPresent(IS_ENABLEDIx)) "<< endl;
```
slayoo, line 755: ditto?

src/basic_fun.cpp
```diff
@@ -676,6 +676,117 @@ namespace lib {
...
+      if(pType == GDL_OBJ) {
+	enabled = interpreter->IsEnabledGCObj(static_cast<DObjGDL*>( p));
+	}
+      else {
+      if( pType == GDL_PTR) {
+	  enabled = interpreter->IsEnabledGC(static_cast<DPtrGDL*>( p));
+	} else {
+//	if(trace_me) 
+//	cout << " heap_refcount(prm=lonarr, KeywordPresent(IS_ENABLEDIx) "<< endl;
```
slayoo, line 764: ditto

src/basic_fun.cpp
```diff
@@ -676,6 +676,117 @@ namespace lib {
...
+	  Guard<DPtrGDL> ptr_guard(ptr);
+	  for( SizeT i=0; i<nEl; ++i)
+	        (*ptr)[ i] = (*pL)[ i];
+
+	  enabled = interpreter->IsEnabledGC(ptr);
+	}
+      }
+      e->SetKW( IS_ENABLEDIx, enabled);
+    }
+// #endif
```
slayoo, line 786: ditto

src/objects.cpp
```diff
@@ -203,6 +210,16 @@ void InitStructs()
...
   structDesc::GDL_CONTAINER_NODE = gdlContainerNode;
 
+  DStructDesc* gdlContainer = new DStructDesc( "GDL_CONTAINER");
+  gdlContainer->AddTag("GDL_CONTAINER_TOP", &aLong64);
+  gdlContainer->AddTag("GDLCONTAINERVERSION", &aInt);
+  gdlContainer->AddTag("PHEAD", &aPtrRef);
+  gdlContainer->AddTag("PTAIL", &aPtrRef);
+  gdlContainer->AddTag("NLIST", &aLong);
+  gdlContainer->AddTag("GDL_CONTAINER_BOTTOM", &aLong64);
+//  gdlContainer->AddParent(gdl_object);// no operator overloading
```
slayoo, line 220: ?

src/basic_fun.cpp
maynardGK: trace_me is a debug flag that can easily be activated for future development, the outputs are 
where I was interested at one time and would be again when more effort is given.

src/GDLInterpreter.hpp
maynardGK: Yes.  This means, globally, EnabledGC().
  I don't have reason to set this false and am not sure I had a strong logic one way or the other,
so in order to insert feature and let someone else worry about ramifications, I confine it to true.

An operable IsEnabledGC() is kept in objects.cpp that can be happily toggled to no effect.

maynardGK:
@slayoo The object {GDL_CONTAINER} is prototype for the IDL built-in {IDL_CONTAINER} object.
To try and mimic the IDL_CONTAINER features, I use script code.
see src/pro/idl_container__define.pro
 Once I get the rest of the object support code I have, I will drop test_container.pro 
into testsuite.  (it says success but the test are really easy).
`GDL> #test_container
Compiled PROCEDURE : TEST_CONTAINER,TEST,NO_EXIT,SHORT,VERBOSE,HELP
GDL> test_container,/verb
 oc0,1,2 heap_refcount():       1       1       1
GDL_CONTAINER::ADD (internal):  Mixed pointers/Objects attempted
 heap_refcount(oc):       1
% TEST_CONTAINER: =======================================================
% TEST_CONTAINER: =                                                     =
% TEST_CONTAINER: =  NO errors encountered during TEST_CONTAINER tests  =
% TEST_CONTAINER: =                                                     =
% TEST_CONTAINER: =======================================================`

maynardGK:
@GillesDuvert in the last commit you provided the wrong answer to get  test_heap_refcount to work without error.  So testsuite/test_heap_refcount.pro will need to be 
regressed  when reference count is properly done when ptr_valid(/cast) is
called, in the basic_fun branch.

GillesDuvert:
@maynardGK Hi, thanks for the wrning, but could you be more precise?  test_heap_refcount passes with current gdl (after this PR was merged) which is mandatory, but, more important, it passes with IDL8. As GDL behaviour should be the same as in IDL (as it is now), there should be no changes of tested values in future releases, no?

maynardGK:
You're right, I mis-counted. I was thinking of the -other- reference counting error which 
will happen when (not updated) ptr_valid(/cast) is used.  I will make a test_ptr_valid.pro
and add to the basic_fun branch.

maynardGK:
I was wrong again, the modification to basic_fun involving .IncRef had no effect on the results here.
I've commited a test_ptr_valid.pro in the basic_fun branch.
The .IncRef modification line only applies when one is converting from pointer index (long) to actual pointer using /CAST keyword - i.e. only in emergencies and debug situations.

Now, I've done a "git pull origin maynardGK:heap_refcount" from the previous master and I am unsure
of the situation - should I "git checkout master" will that get me back to the common origin?

PR #114, opened 2018-04-09
Implement DELVAR procedure, .RNEW (includes the HEAP_REFCOUNT changes)

maynardGK:
delvar_pro is found in basic_pro.cpp;
methods to GDL internals are added to facilitate this in envt and in dpro.
The .RNEW command is activated in dinterpreter.cpp
compile option flags STATIC and NOSAVE are in place, although they are not in use.

findvar_pro may be useful in a development context - normally it is not compiled.
DELVAR is to be used only from the $MAIN program, as per IDL docs; if it isn't, an
exception is raised with a message.

Tested since 2016.  Smoke test made for current git build.
Stage 1
Greg Jung (aka maynardGK)
gvjung@gmail.com

### DIFF
--- a/src/GDLInterpreter.hpp
+++ b/src/GDLInterpreter.hpp
@@ -34,7 +34,8 @@
 //#define GDL_DEBUG
 //#undef GDL_DEBUG
 //#define GDL_DEBUG_HEAP
-
+bool IsEnabledGC(); // defined in GDLInterpreter.hpp with EnableGC(bool);
+void EnableGC(bool);
 
 class CUSTOM_API GDLInterpreter : public antlr::TreeParser, public GDLInterpreterTokenTypes
 {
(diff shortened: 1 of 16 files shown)


slayoo:
Is this intentional that this PR includes the HEAP_REFCOUNT changes?

slayoo:
A test routine (to include in make-check) would be great!
And big thanks!

slayoo:
Was addition of the src/gdlc.i-1523252260.g file intentional?

maynardGK:
It's not intentional that both commits are in same PR - the are from different branches.  Somehow
the commits were there but no pull requests for them. 
 The file src/gdlc.i-15<etc>.g is also
accidental, a backup file from winmerge when I staged with a wildcard glob. I tried to edit this out just
now, but I can change the file not delete it in Github's system.

maynardGK:
I'm guessing the best/only way for me to "edit" this is to drop it and re-create it on a better separated branch,
so I will close, and be back with "delvar-rnew" branch which will not have a previous commit.

PR #129, opened 2018-04-10
post-git-migration svn r4096 (escape_special_char.pro and file_move.pro)

slayoo:
changes by @alaingdl

### DIFF
--- a/src/pro/escape_special_char.pro
+++ b/src/pro/escape_special_char.pro
@@ -1,84 +1,108 @@
 ;+
 ; NAME: ESCAPE_SPECIAL_CHAR
-; 
+;
+; important Nota Bene : AC 2018 : it would be great to be able to *remove* this
+; code in the future ... All "file_*" functions should be able to
+; manage by them-selves the special chars.
+;
 ; ESCAPE_SPECIAL_CHAR function escape special characteres present
 ; in input string argument and returns the result.
 ; Special characteres can be given by Keyword : list_of_special_char
 ; or default list is used.
 ;
-; MODIFICATION HISTORY:
-;   - June 2010: created by Lea Noreskal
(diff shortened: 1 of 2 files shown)


slayoo:
@alaingdl could you please comment if this has anything to do with this issue reported by @olebole:
https://sourceforge.net/p/gnudatalanguage/patches/97/
Thanks!

alaingdl:
I would say no, but I need to check again

the problem in "test_file_move.pro" is we use a file with a '$' which may be consider
as an ENV (printenv) in Bash shell ...

slayoo:
BTW, are these changes covered in the testsuite?

src/pro/escape_special_char.pro
```diff
@@ -1,84 +1,108 @@
...
+if N_ELEMENTS(str_in) GT 1 then MESSAGE, 'input string should not be an array'
+;
 if KEYWORD_SET(verbose) then begin
-    MESSAGE, /continue, '/verbose keyword not available, please contribute'
+   MESSAGE, /continue, '/verbose keyword not available, please contribute'
 endif
 ;
 ; list of special characteres can be modified if it is not complete
-;if( not KEYWORD_SET(list_of_special_char)) then list_of_special_char ='[*.^$,&\]'
+;if( not KEYWORD_SET(list_of_special_char)) then list_of_special_char ='[*^$,&\]'
```
slayoo, line 60: let's write a test for this change!

alaingdl: those lines are commented ...
the real case in l47 vs l64 : removing "*" & ".", adding space " "

alaingdl: as I said in the comments, the goal would be to *remove* this old code, and having in the C-code
using primitives (without "spawn")

slayoo:
Merging as it anyhow is in the SF repo

PR #152, opened 2018-04-11
Parallel minmax

GillesDuvert:
New try, from a cleaner set of files.


GillesDuvert:
again this pull request has not the "good" files. I remove it.

Issue #127, opened 2018-04-10
post-git-migration commits from SVN


slayoo:
Based on: https://sourceforge.net/p/gnudatalanguage/svn/log/

- [x] [r4137] by gilles-duvert
   ```correctly throws an error for widget_info(nonexisting_widget,/valid) instead of returning zero.```
- [x] [r4136] by gilles-duvert
   ```should remove GH #251. Please confirm.```
- [x] [r4135] by gilles-duvert
   ```conformant error message.```
- [x] [r4134] by gilles-duvert
   ```align with GH version today```
- [x] [r4133] by gilles-duvert
   ```align with GH version today```
- [x] [r4132] by gilles-duvert
   ```align with GH version today```
- [x] [r4131] by gilles-duvert
   ```align with GH version today```
- [x] [r4130] by gilles-duvert
   ```not valid anymore```
- [x] [r4129] by gilles-duvert
   ```replaced by formats.IDL```
- [x] [r4128] by gilles-duvert
- [x] [r4127] by gilles-duvert 
- [x] [r4126] by gilles-duvert
   ```align with GH repository```
- [x] [r4125] by alaingdl (https://github.com/gnudatalanguage/gdl/pull/292):
   ```these lines are mandatory on Debian with gcc 6.3.0```
- [x] [r4124] by alaingdl:
   ```update of "minimal_script4gdl.sh" to manage Cmake versions above 3.2 ... (needed on OSX)```
- [x] [r4123] by alaingdl (https://github.com/gnudatalanguage/gdl/pull/291):
   ```missing include <omp.h> in a special case```
- [x] [r4122] by alaingdl (https://github.com/gnudatalanguage/gdl/pull/293):
   ```as for CONTOUR, SURFACE should accept "degenerated" 3D equivalent to 2D array```
- [x] [r4121] by alaingdl:
   ```WordExp() crashed when called with a null string ... as a result, openr was broken ...```
- [x] [r4120] by alaingdl:
   ```update "minimum_script4gdl.sh" to have 0.9.8 & 0.9.8 svn```
- [x] [r4119] by gilles-duvert:
   ```align with GH repository (minor changes)```
- [ ] [r4118] by gilles-duvert:
   ```closed two formatting bugs. see on github, pr #197```
- [x] [r4117] by gilles-duvert:
   ```Various small changes to avoid most compilation warnings, especially on travis+icc and osx/clang```
- [x] [r4116] by gilles-duvert:
[...]

PR #156, opened 2018-04-11
new try at parallel minmax

GillesDuvert:
New try again. Working with branches is clearly not my cup of tea.


GillesDuvert:
OK leaving the icc petty problem, all in all it works. To be merged.

slayoo:
Any idea why test_bytscl.pro and test_modulo.pro both fail after merging it? Leaving this as is will make all CI runs report "red" in pull requests. If the failures are to be ignored, please add these two files to the list of travis-ingored scripts in .travis.yml (sed one-liners in before_install section). Thanks

GillesDuvert:
no idea, and it is not expected that all tests will pass anyway. Sometimes these test do not even pass with IDL8. I suspect a problem with the test procedure, not the test_xxx.pro files themselves. I will not look into this mess. Switching to python-scripted tests seems more elegant, too. And we need more non-regression tests comparing listings with reference listings.
How to make the test on travis and at the same time report "green"?

slayoo:
Here is a pull request disabling these two tests: https://github.com/gnudatalanguage/gdl/pull/158
Could you please also mark which of the post-migration SVN commits are covered with this PR: https://github.com/gnudatalanguage/gdl/issues/127
Thanks!
S.

GillesDuvert:
As GH gdl trunk is always a lot of commits behind (for good cause), I still try to have SF's SVN up-to-date in my (clumsy) manner. Up-to-date with my most recent developments and those of others that I have tested (on my machine), be they for good or not.
AFAIK this PR: #127 covers r 4095,4094,4093,4092

PR #157, opened 2018-04-11
Speedup builds by splitting huge template file

GillesDuvert:
Removed smooth, minmax, where and convol from datatypes.[c|h]pp to let them live their own life. This improves an awful lot the speed of compilation on multicores and reduce the individual memory imprint.


GillesDuvert:
hmm... better if I start again from a cleaner branch. removing this PR

PR #158, opened 2018-04-11
Temporarily disable test_bytscl.pro and test_modulo.pro ??? (as of now the master branch CI fails!)

slayoo:


slayoo:
@GillesDuvert Please note that on Travis the tests are run only for OSX and ICC. On OSX we use Clang, hence the icc build is the only one supporting OpenMP for which tests are run. Since the minmax changes are OpenMP-related, I'm a bit worried that these two failing tests should get more attention?

slayoo:
This PR is meant just as a temporary kludge to allow regression tests on Travis.
Need for a proper fix reported here: https://github.com/gnudatalanguage/gdl/issues/159

GillesDuvert:
@slayoo  you forgot gcc? Travis  run both ICC and GCC. both have OpenMP.
The OSX Clang tests are a must and show useful information for the case when openmp is disabled.
(Meaning that GDL's version on OSX is forcibly slow? I cannot imagine).
I see that openmp is part of clang (see http://www.llvm.org/) and so should be available on travis (just by adding flag -fopenmp ?). a TODO?

That said, you're right, tests may fail because of parallelism problems. GDL speedup is done in many ways, using EIGEN of course, but also by using different code parts depending on the presence of one or many threads, and sometimes on the presence of some hardware acceleration (SIMD). Use of new features of OPENMP4 is also contemplated. So test files should ideally be run with 1, more than 1 (max number of)  threads and  without openmp at all.

slayoo:
to clarify, on travis we have as of now:
- osx/clang: compilation + tests
- linux/icc: compilation + tests
- linux/gcc: compilation *only*

GillesDuvert:
Let's do it. we know we have to test the test files anyway!

GillesDuvert:
according to listings, on travis gcc performs tests too. 
`if [[ $TRAVIS_OS_NAME != 'gcc' ]]; then make check; fi`

slayoo:
Apologies for false alarm! Thank you for pointing that out, will remove the bogus condition right away

slayoo:
https://github.com/gnudatalanguage/gdl/pull/168

slayoo:
For the record, the problem of tests not running on Linux was apparently solved with this commit:
https://github.com/gnudatalanguage/gdl/pull/109/files#diff-d75e53c1a4658d8c288686195aa2ac75R90

PR #187, opened 2018-04-16
 Patch to avoid accidental assert failures from EnvT::KeywordIx

maynardGK:
Converts a hard exit/crash to a harmless diagnostic for new imperfect software. (see comments i n commit). Also removal of a modification from 2014 that was suppose to prevent duplicate keywords.

Greg Jung 
gvjung@gmail.com

### DIFF
--- a/src/envt.cpp
+++ b/src/envt.cpp
@@ -1168,10 +1168,12 @@ int EnvT::KeywordIx( const std::string& k)
   assert( pro != NULL);
   int val=pro->FindKey( k);
   if( val == -1) {		//  assert( val != -1);
-    cout << pro->ObjectName() << "  Key: " << k << endl;
-    cout << "Invalid Keyword lookup (EnvT::KeywordIx) ! " << endl
-		<< " Returning the wrong (but a valid) key index of zero" << endl;
-		val = 0;
+    cout << "Invalid Keyword lookup (EnvT::KeywordIx) ! "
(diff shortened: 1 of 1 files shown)


maynardGK:
The error that stopped the test was an internal server error. The rest look ok.

slayoo:
I've just restarted the failed job

GillesDuvert:
Hi @maynardGK , I do understand the intent, I had once wanted to get rid of these annoying asserts.
But consider: this particular assert serves to find a severe programming bug that should be cured asap. **As asserts are not trapped in Releases**, the presence of such programming errors is invisible for normal users (all the GetKw() functions return NULL as if the Kw was absent). It is thus mandatory to keep a way to inform the authors that they forgot to enter the keyword in the dictionary. We as authors must test our changes in Debug mode first and with these asserts we are bound to find our mistakes, sooner or later.
Some heavily used programs trap such errors and ask users to report by sending an automatic bug report. We are not prepared to do that I believe.

maynardGK:
@GillesDuvert 
That's ok, but prior to exit the message should be made - or the bug remains undiagnosed.
Setting the index to zero - thereby invoking the first possible keyword - may be too lax, I agree.
But the message is helpful, and a message with exception, preventing execution of the subroutine,
may be enough of a "heads up".  I'll commit on top of this with an exception, then.

maynardGK:
I am closing this as the latest modification is used in basic_fun branch to restart testing.

Issue #178, opened 2018-04-16
PM: TITLE keyword incompatible with the FORMAT keyword


slayoo:
As reported on SF.net (https://sourceforge.net/p/gnudatalanguage/bugs/366/):

```
GDL> pm, [1,2], format='(I2)', title='PM'
% Type conversion error: Unable to convert given STRING: 'PM' to LONG.
0
1
2
```

GillesDuvert:
strangely this is occuring only if the procedure is called 'pm'. renaming it to 'mp' or anything else works.

slayoo:
Gilles, do you mean renaming in libinit.cpp?
```
 961   const string pmKey[] = {"FORMAT", "TITLE", KLISTEND };
 962   new DLibPro(lib::pm, string("PM"), -1, pmKey);
```

GillesDuvert:
Ah I see. lib::pm wrongly uses the format *with* the title.

alaingdl:
I am lost here ! Some working parts of the code were  removed. And the test stopped working ...
I am on the way to revert that after some extra tests !

GillesDuvert:
Two important points:
- PM : the doc does not mention any option, only "PM array, array2 ..." . So the code should be cleaned and the error will not appear anymore
- the doc mentions "RM" also that does not exist in GDL!

alaingdl:
The IDL doc is wrong. Those codes are *pro in recent IDL distros, then interfaces and keywords are clear.
I can write to NV5 to mention that but for me it is not important at all ! Just a side effect when I wrote
the tests for SHMMAP, using /proc/self instead of ../src ...

PR #190, opened 2018-04-17
Pm bug

GillesDuvert:
solves issue #178


slayoo:
Thanks Gilles! 

BTW, if you include "Closes #NN." string in a commit message, the relevant issue will be automatically closed upon merging of a PR

PR #240, opened 2018-04-21
CI for Windows (by Jeongbin - thank you!)

slayoo:


GillesDuvert:
I suggest to drop completely PSLIB which is badly handled (probably our fault) in gdl. One can live without, as the limitations of the plplot postscript driver we depend on are not terribly annoying. Just create an enhancement request stating exactly what should be improved in the raw PS output (withouts pslib) and it will be solved easily.

slayoo:
let me close this one and bring relevant appveyor setup updates step by step in separate PRs

Issue #15, opened 2018-04-05
launchtest linking problem on Linux/Intel compiler


slayoo:
As depicted here: https://travis-ci.org/slayoo/gdl/builds/362621901
in some cases execution of tests seems to be impossible on Travis Linux machines due to linkage problem:
```
CMakeFiles/launchtest.dir/launchtest.c.o: In function `main':
launchtest.c:(.text+0x3b): undefined reference to `__kmpc_begin'
launchtest.c:(.text+0x4d): undefined reference to `__kmpc_end'
launchtest.c:(.text+0xef): undefined reference to `__kmpc_end'
launchtest.c:(.text+0x10d): undefined reference to `__kmpc_end'
```

PR #252, opened 2018-04-30
Restore compatible operation of X and WIN plots

maynardGK:
These modifications are several slight edits, nost of which are regressions, that restore proper operation
of plot windows - principally under WIN.  
 Objects.cpp is a routine called upon initailization of GDL.  Historically the widget initialization was done via a call to GDLWidget::Init(), from this routine.  Lately it was moved to graphicsdevce.  For whatever reason,  GDLWidget::Init() is again called from InitObjects routine and SetWxStarted is added to the Init() procedure located in gdlwidget.cpp.
 The option NO_WIDGET_DRAW is added to prevent disruption when widgets are attempted while the 
deviceWIN plotting device is in use - it doesn't go well!  
The option USE_WINGDI_NOT_WINGCC can be used to use a new device driver in PLPLOT version 13 or above.  Still, a patch is needed in this device for proper operation of the UnsetFocus() routine.

Finally, UnsetFocus() is re-activated from GraphicsMultiDevice::WShow


maynardGK:
Bad luck? twice, same error, on test that creates a window:
 15/180 Test  #15: test_bug_2949487.pro ...............   Passed    0.11 sec
        Start  16: test_bug_2974380.pro
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself

Close and re-open

maynardGK:
This is again generating an error in 0.5, 0.10 from test_bug_2974380.pro !!
I am going to close this and slooowly reimpose the edits on another PR.  At this point
the only changes to note are the UnsetFocus() call from MultiDevice and GDLWidget::Init().
Operation on linux is fine, BTW, it is only clang/OSX that is bothered.

PR #257, opened 2018-05-03
Remove or invalidate references to qhull library as we still do not not use it.

GillesDuvert:


slayoo:
there are also mentions of QHULL in the readme.md

PR #258, opened 2018-05-04
Merge ~ treatment for Win32 into WordExp()

maynardGK:
GDL uses a function "WordExp()" which will call unix-based "wordexp()" which has
    some quirks and, in the case of _WIN32, is not present (nor, we assume, in the case of FreeBSD).
    This cleans up the "#if 0 ... #endif" blocks of the code and inserts the _WIN32 case for tilde expansion.

    Greg Jung (aka maynardGK)
    gvjung@gmail.com


maynardGK:
Except for a time-limit, all tests passed.  Closing this PR; the str.cpp code is included in the file_etc branch PR.

PR #262, opened 2018-05-06
Slowly find where OSX/clang test fails.

maynardGK:
This is the baseline build of code for win_plots, where invalid windows plots were
fixed and the keyboard focus was correctly restored for Linux.  The Travis builds
xxx.5, and xxx.10 failed via test_bug_2974380 this is for clang compiled on OSX
systems.

This first commit only submits the repaired devicewin.cpp, .hpp files (which are not used in OSX)
If this passes the next step will be to activate the UnsetFocus() call in WShow.


maynardGK:
So, it is passing when UnsetFocus was re-inserted.  Now however we have another significant insertion,
  GDLWidget::Init();       was not called if GDL_USE_WX was not set to "YES".  Next then, we add 
this call after creation of the WX plot device; in earlier failures this call was made before the deviceWX was accessed.

maynardGK:
And now the errors have returned once there is a wxWidget initialization call that actually
gets executed (previously the iitialization call came only when GDL_USE_WX was "YES" - a condition
not set, I assume, in the test).

slayoo:
@maynardGK should we add a "WIP" prefix to name of this PR?

maynardGK:
This is work finished, but it is very ugly stuff!  basically I found out that OSX/Clang test will fail if we initialize widgets.  I located the init() and uninit() routines to more functionally-appropriate locations and made widgets.cpp survive upon getting errors on a draw widget.  For my own use I would tend to use only text aspect of widgets but want the best plot possible; in other applications the integrated Wx
widget is needed.
I don't see much improvement in the situation, WX plotting won't be as good as the X driver, and running a widget_draw with "X" or "WIN" doesn't work right.    So I would just assign to Gilles whether to accept or delete this,
 or try it out for a while:
The "UnsetFocus" aspect is a noticeable improvement, I still get widgets even though GDLWidget::Init() isn't called - but I'm guessing there are other ill effects to that.

maynardGK:
Actually omitting the GDLWidget::Init() does crush the widget capability, so I'll just close it as it stands.

slayoo:
ok, thanks

PR #266, opened 2018-05-08
 Restore src/pro/cw_bgroup.pro and revise testsuite/demo_widgets.pro

maynardGK:
In some manner the compound widget procedure CW_BGROUP was lost to the GDL
distribution; the last time it appeared was in 2014 during 0.9.5. Bring this back
to the pro/ subdirectory with a small correction.
 A modified testsuite/demo_widgets.pro is tested to run with or without a viable widget/plot cohension. It will also function under the programming given with the NO_WIDGET_DRAW option.

Greg Jung
gvjung@gmail.com

### DIFF
--- a/src/pro/cw_bgroup.pro
+++ b/src/pro/cw_bgroup.pro
@@ -1,0 +1,214 @@
+; part of GNU Data Language - GDL 
+;-------------------------------------------------------------
+;+
+; NAME:
+;       CW_BGROUP
+; PURPOSE:
+;       Create a compound checkbox, radiobox or regular button group
+;
+; CATEGORY:
+;       GUI widgets
+;
+; CALLING SEQUENCE:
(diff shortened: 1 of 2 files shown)


GillesDuvert:
@maynardGK the cw_bgroup was removed because it was somehow 'false'. In the sense that it worked at some point based on the very preliminary C++ code for widgets, largeley amended since. `cw_bgroup` is used in, e.g., atv.pro and unfortunately 'our' cw_bgroup does not work: see what happens in demo_widgets. IDL's cw_bgroup works well instead.  Putting it back would cause severe regression in `atv,` `look` and many other widgets applications, that people succed to use with idl's one.
I suggest to rename this cw_bgroup to cw_bgroup_donotuseyet (for example) as it just need to be revised in order to work in all conditions.

Issue #19, opened 2018-04-05
Support [start:end:interval] array generation


slayoo:
(moved from https://sourceforge.net/p/gnudatalanguage/bugs/731/)
As reported on SF by Gilles:

A new (IDL8) feature:

```
IDL> x = [0:10:0.01]
IDL> help,x
X               FLOAT     = Array[1001]
```
```
GDL> x = [0:10:0.01]
% Lexer/Parser syntax error: expecting RSQUARE, found ':'
```

Issue #89, opened 2018-04-07
new keywords START= (and INCREMENT) in MAKE_ARRAY() and *INDGEN()


slayoo:
As reported on SF.net: https://sourceforge.net/p/gnudatalanguage/feature-requests/156/ and https://sourceforge.net/p/gnudatalanguage/feature-requests/155/

A keyword START has been added in IDL since 8.2.1 in MAKE_ARRAY()
Then a INCREMENT keyword (related to START !) was add in 8.3

A keyword START has been added in IDL since 8.2.1 for all the *INDGEN() functions.

rexso:
In addition to #272, #316 finalises this issue, no?

slayoo:
correct, thanks!

PR #272, opened 2018-05-15
Added support for START and INCREMENT keywords in INDGEN functions

rexso:
Partly solves #89 by adding support for START and INCREMENT keywords to all INDGEN functions.
Will create a separate PR after adding these keywords to MAKE_ARRAY as well.

START and INCREMENT were added in IDL 8.2.1 and 8.3 respectively.

Hope this is useful. I am new to GDL hacking, so comments are welcome and appreciated.
I would love to write tests, but I lack experience with IDL and have no IDL licence,
so for now I have just verified the results based on the examples found on this page:
https://www.harrisgeospatial.com/docs/Creating_Arrays.html

### DIFF
--- a/src/basic_fun.cpp
+++ b/src/basic_fun.cpp
@@ -791,24 +791,25 @@ namespace lib {
   BaseGDL* bindgen( EnvT* e)
   {
     dimension dim;
+    DDouble off = 0, inc = 1;
     //     try{
     arr( e, dim); 
     if (dim[0] == 0)
       throw GDLException( "Array dimensions must be greater than 0");
 
-    return new DByteGDL(dim, BaseGDL::INDGEN);
+    e->AssureDoubleScalarKWIfPresent("START", off);
+    e->AssureDoubleScalarKWIfPresent("INCREMENT", inc);
+    return new DByteGDL(dim, BaseGDL::INDGEN, off, inc);
     /* }
        catch( GDLException& ex)
        {
        e->Throw( "BINDGEN: "+ex.getMessage());
        }
     */ }
-  // keywords not supported yet
   BaseGDL* indgen( EnvT* e)
   {
     dimension dim;
-
-    // Defaulting to GDL_INT
+    DDouble off = 0, inc = 1;
     DType type = GDL_INT;
 
     static int kwIx1 = e->KeywordIx("BYTE");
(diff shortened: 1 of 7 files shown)


slayoo:
big big thanks for this contribution!

a test simply exemplifying the new keywords would be great and simple to add, let's say:
- a minimal (even 3-line) file with a test procedure added to the testsuite directory
- a new one-line entry in the testsuite/Makefile.am file which will trigger execution of this test during "make check" and hence also during continuous integration on Travis

The test procedure could even look like:
```
pro test_indgen_start
  print, indgen(10, start=1, increment=2)
end
```
Of course covering more cases in the test file would be better.

Thanks again!

slayoo:
Here is an example of a minimal test addition (for an unrelated issue): https://github.com/gnudatalanguage/gdl/pull/273/files

rexso:
I have now added some basic tests for the INDGEN functions which also include the START and INCREMENT keywords. Not sure if there are some naming conventions in use for the test suite, or if my approach is wrong, but the tests are passing at least :smile:

These tests are my first lines of IDL code ever, so please forgive my ignorance if I have done any big mistakes.

slayoo:
Thank you!

One more question/comment: as of now, regardless of the type of array to be returned, the start and increment values are treated as doubles. Performance-wise this might have some impact I guess due to unneeded conversions, but most importantly I'm wondering if there would be no surprise situations where user-supplied double would be silently converted into an integer?

rexso:
@slayoo fair question. The rationale behind using double regardless of type, is that this seems to be the way IDL does it, according to this statement in the [INDGEN documentation](https://www.harrisgeospatial.com/docs/INDGEN.html):

> Set this keyword to a double-precision number giving the spacing between values in the array. The default is 1. Setting this keyword is equivalent to multiplying each array element by this value and then converting to the result type.

The way I interpreted this, is that a double is used regardless, but converted when needed. There is indeed some (neglectable?) performance impact that could be solved by using templates instead. However, I am also a bit concerned about surprises if there is some implicit conversion, so I thought the best choice would be to emulate IDL and rely on IEEE 754 rules for rounding. Please note that I have not reverse engineered IDL, so my conclusion is solely based on the above mentioned statement from the documentation. At least from my tests the result produced is the same as expected in IDL.

slayoo:
@rexso thank you for clarification
@GillesDuvert @alaingdl any opinion on the double issue?

slayoo:
@olebole @opoplawski  any opinion on the double issue?

opoplawski:
I don't have an opinion.

GillesDuvert:
Hi,
This looks great! Special thanks for your careful reading of the IDL documentation!
The  double precision is unavoidable and I cannot imagine why the IEEE 754 rounding would not be OK (anyway that's what test procedures are for).
What I would do, as we do in several places, to speed code, is to triplicate the loops, one for each case: no off and inc, no inc, off and inc present. It may be worthwile to reintroduce the parallelizing in the 2 latter cases, but that's always tricky. Better leave it to a further change in the future.

Also, to be complete it certainly needs some additional work at the language level, in the ANTLR files, as users may also expect a=[0:10:0.1] to work. This, arguably, can be done in a second time however.

rexso:
@GillesDuvert thanks for your comment. I totally agree regarding the double precision, and also think that tests should be used to catch unexpected behaviour. The proposal for how to potentially speed up the code sounds reasonable, and I think this might be suitable for a future refactor task. I might look into it after getting more familiarised with the GDL codebase.

Running into some IDL code using the colon operator [start:finish:increment] is actually the reason for adding the START and INCREMENT keywords to INDGEN, as this seems to be the foundation for this short-hand syntax. I am still new to both IDL and GDL, so I really want to dig more into the GDL codebase before undertaking an ANTLR task to improve GDL syntax. At least supporting START and INCREMENT in INDGEN works as a temporary substitute for the colon operator in my case :smile_cat:

GillesDuvert:
I suggest we merge your PR as is. The small speed improvements can be done afterwards. I would advice to stay away from the ANTLR part as it is too specialized. I'll try to patch the antlr interpreter in the coming weeks to finally close #19.

slayoo:
Merged, thank you!

slayoo:
Submitted a relevant new issue for the speed improvement: https://github.com/gnudatalanguage/gdl/issues/279

PR #273, opened 2018-05-15
adding a test for the wordexp problem fixed in 3b4ecba

slayoo:


slayoo:
sorry for the messy commits list - this one is definitely intended for "squash and merge". After that I'll recreate my fork to clean this up.

Issue #45, opened 2018-04-07
option to regenerate antlr files with system antlr


slayoo:
#As discussed at SF.net:
- https://sourceforge.net/p/gnudatalanguage/patches/39/ with @sfabbro
- https://sourceforge.net/p/gnudatalanguage/patches/100/ with @olebole and @GillesDuvert 
the idea of shipping all ANTLR files within GDL does not have fans among packagers, for good reasons. 

As pointed out in the SF issues, the solution which was introduced e.g. in:
- Gentoo (https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-lang/gdl/files/0.9.5-antlr.patch) 
- Debian (https://sources.debian.org/patches/gnudatalanguage/0.9.8-2/Build-with-Debian-s-antlr-instead-of-the-embedded-copy-in.patch/)
to link with system antlr has the problem of creating version mismatch between the generated files used in compilation and the system files used.

I believe, we should offer via CMake the option to use system antlr (as in the Gentoo patch) but then trigger a regeneration of all antlr-generated files, i.e.:
```
$ cd gdl/src
$ for i in *.g; do antlr $i; done;
```

Another issue is that reportedly the GDL-shipped antlrheaders are modified with respect to the original (https://sourceforge.net/p/gnudatalanguage/patches/100/).

slayoo:
See also: https://github.com/olebole/gnudatalanguage/pull/4

slayoo:
For the record, the regeneration is done e.g. in the Gentoo build spec:
https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-lang/gdl/gdl-0.9.6-r2.ebuild#n79

slayoo:
Two points:
- as introduced in #529, the regeneration is now done on Travis and the the output files are checked for compliance with the ones from the repo
- as discussed in #198, there is also the question if to remove at all the ANTLR-generated files and the ANTLR headers from the repository and tarballs?

GillesDuvert:
Making GDL compile easier for everybody is incompatible with requiring those same people to install yet another dependency---moreover an old version of, ANTLR, no?

olebole:
Why an old antlr? On Debian, we always use the latest version (currently 2.7.7).
[...]

PR #274, opened 2018-05-21
Use size_t for string positions

opoplawski:


slayoo:
the change in the .g antlr input file is not reflected in a generated file

slayoo:
Alain, I suggest reverting as this one includes changes to generated files only, and not to the .g grammar files.

slayoo:
Sorry, vice versa - change in .g file is not reflected in a relevant generated file, right?

opoplawski:
I guess I would argue that generated files should not be in git but should be generated for releases only.

GillesDuvert:
Hm, we have many contributors that change GDL\*pp files (and FMT\*pp files) , and that is good, but the bottleneck is to insure these improvements are backported to the .g files at some point. OTH, it is just too easy to forget that .g files **may** have changed on GH... Not having the .g files on GH, or, rather, having these files elsewhere on GH is probably a beginning of solution, especially if there would be some way of making gnudatalanguage/gdl (here) depend from gnudatalanguage/antlr (there).

The other related topic is that the GDL\*pp files should be, eventually, recreated from the .g files by the packagers (#45) . And that "stock antlr" and not the content of the antlr subdirectory should be used by them. This points to an evolution of our CMake. Volunteers?

And, besides, as you wanted to ask, yes, there seems to be a way to have gdl parsing speed multiplied by a huge amount... And gdl files are antlr 2, not even antlr 3... So there **will** be some work on .g files in the future.

slayoo:
Why would not having .g files in the repo help in any way? Orion suggested not to keep the generated files. I would then add suggestion not to store the antlr header files either... (but I'm almost certain Marc mentioned once that they are actually a modified version).

GillesDuvert:
they are a modified version, but I've played with antlr a lot, especially with stock antlr and that did not seem to make any difference at the time...
And, yes, I misread @opoplawski comment, but I fail to see how removing the GDL\*files is going to make life easier for contributors who at the very least should have antlr2.7 on their machine... If any, the current situation with its drawbacks is bearable. But someone can make an experiment on a clone of gdl and tell us.

slayoo:
btw, the antlr website mentions "ANTLR v2 to v3 converter (in progress and only marginally useful)": http://www.antlr3.org/download/v2v3-0.2.tar.gz

GillesDuvert:
Hi, @slayoo , thanks for the tip. I'll give a try.

PR #275, opened 2018-05-21
Quiet possible uninitialized variable warning

opoplawski:


slayoo:
Thanks Orion! 
GDLInterpreter.cpp is a generated file, hence the changes need also to be reflected in gdlc.i.g

opoplawski:
Hmm, no idea to to make this change in gdlc.i.g.

GillesDuvert:
Yes, l'll make this small change in one of the future updates of the .g files
Being in sync between .g and .cpp/.hpp files is very important, and all commits should be checked for that before merging. They should be tagged "todo-generated-files" also.

GillesDuvert:
Should be present now that #308 is done.

PR #276, opened 2018-05-21
Use size_t, unique_prt

opoplawski:


slayoo:
here the change is only in the antlr input file, not in the generated files

GillesDuvert:
done in #303 (a copy of this PR in fact)

Issue #153, opened 2018-04-11
test_parse_url is failing


opoplawski:
```
144: Test command: /export/home/orion/fedora/gdl/gdl-0.9.8/build/testsuite/launchtest "test_parse_url.pro"
144: Test timeout computed to be: 9.99988e+06
144: % Compiled module: TEST_PARSE_URL.
144: % PARSE_URL: URL fragment left out: #etc
144: % Compiled module: ERRORS_ADD.
144: % TEST_PARSE_URL_IPV6: Error on operation : { http user @pass host  path argument?value}
144: Case :        0
144: expected  : { http user @pass host  path argument?value}
144: processed : { http user @pass host  /path argument?value}
144: % TEST_PARSE_URL_IPV6: Error on operation : { http   [FEDC:BA98:7654:3210:FEDC:BA98:7654:3210] 80 index.html }
144: Case :        1
144: expected  : { http   [FEDC:BA98:7654:3210:FEDC:BA98:7654:3210] 80 index.html }
144: processed : { http   [FEDC:BA98:7654:3210:FEDC:BA98:7654:3210] 80 /index.html }
144: % TEST_PARSE_URL_IPV6: Error on operation : { http   [1080:0:0:0:8:800:200C:417A]  index.html }
144: Case :        2
144: expected  : { http   [1080:0:0:0:8:800:200C:417A]  index.html }
144: processed : { http   [1080:0:0:0:8:800:200C:417A]  /index.html }
144: % TEST_PARSE_URL_IPV6: Error on operation : { http   [1080::8:800:200C:417A]  foo }
144: Case :        4
144: expected  : { http   [1080::8:800:200C:417A]  foo }
144: processed : { http   [1080::8:800:200C:417A]  /foo }
144: % TEST_PARSE_URL_IPV6: Error on operation : { http   [::192.9.5.5]  ipng }
144: Case :        5
144: expected  : { http   [::192.9.5.5]  ipng }
144: processed : { http   [::192.9.5.5]  /ipng }
144: % TEST_PARSE_URL_IPV6: Error on operation : { http   [::FFFF:129.144.52.38] 80 index.html }
144: Case :        6
144: expected  : { http   [::FFFF:129.144.52.38] 80 index.html }
144: processed : { http   [::FFFF:129.144.52.38] 80 /index.html }
144: % Compiled module: BANNER_FOR_TESTSUITE.
144: % Compiled module: GDL_IDL_FL.
[...]

PR #283, opened 2018-05-28
Changing old parse_url() to parse_url_php() and adding a new parse_url() compatible with IDL

sflinois:
This change was made to address the issue : test_parse_url is failing #153
The parse_url() that was implemented was based on the php parse_url() which has some advantages. I kept it and renamed it parse_url_php(). I also renamed test_parse_url.pro into test_parse_url_php.pro.

I implemented a new parse_url() based on the IDL version of the function and created some tests in test_parse_url.pro

It is my first contribution to GDL, every feedback is appreciated !


sflinois:
I should have asked before making the pull request, but i have some questions about codecov and how functions are implemented in GDL.

I'm not really sure how codecov work, I guess it uses the testsuite from GDL and check if the tests cover well the code. I don't really know.

I also did not put the test_parse_url_php.pro in the testsuite, since it was not compatible with IDL i thought it might cause problems. 
I also put the parse_url_php() function in basic_fun.cpp, i also don't know if there is any other place i should have put it.
Maybe the old parse_url should have juste been replaced and not changed to a different name

slayoo:
@sflinois big thanks for the PR and apologies for delayed response.

Re codecov: one of the travis builds has the CODECOV env var set which triggers addition of additional code-usage-tracking flags for the compiler. Then, after execution of tests, the statistics are uploaded to codecov.io to generate the reports. For more details, please have a look at this commit in which the service was introduced: https://github.com/gnudatalanguage/gdl/commit/a437102b2d922cd8c6e437e9d633c9096151ebe5

I haven't had a look at your changes yet, but one key thing would be to re-enable execution of the url test file in the .travis.yml file. It is disabled there temporarily: https://github.com/gnudatalanguage/gdl/blob/master/.travis.yml#L52 (in order to enable Travis help us catch regressions - since this test was failing at "commit zero" on github, executing it would give always a failure for the whole CI run).

HTH, thanks again!
Sylwester

slayoo:
Thanks!
+ I vote for removing the old PHP-based version

slayoo:
Just one last comment: I suggest removing the comment block beginning with SA (which was meant to mean my initials ;)): https://github.com/gnudatalanguage/gdl/pull/283/files#diff-86f6cdbb810c527276cef16f6b0be541R8022
In particular, the code can only be based on IDL behaviour not on IDL code - and I hope this is the case here!
Thanks, S.

sflinois:
Alright, i will remove the comment block with SA, sorry about that ^^
I found cases where PARSE_URL throws exceptions :
> GDL> parse_url("http://example/dir?query:port")
> InterpreterLoop: Exception: basic_string::_M_create
> GDL> parse_url("http://example/dir:port?query")
> InterpreterLoop: Exception: basic_string::_M_create
> GDL> parse_url("http://example:port?query/dir")
> InterpreterLoop: Exception: basic_string::_M_create

I'll rewrite the function and add the cases to the testsuite :)

slayoo:
Thanks again!

PR #286, opened 2018-05-30
 Combined merger of 5 inter-related branches

maynardGK:
combo = legacy_list + basic_fun + basic_pro + gdlhelp_fin + legacy_hash

      This PR combines five inter-related change branches that have been waiting since 
before gdl0.9.5 in essentially the same form as now.  Still waiting to merge, once this resides in 
master, are the in-place insertion and extraction of data held in LIST and HASH objects.
      Because of the extensive editing changes made to LIST and HASH, this intermediate state can still
be termed "legacy" in that the general behavior should be the same except for a few enhancements.
I was, however,  able to add the "TOARRAY" method without a lot of trauma.
      Internally, there are few small changes  with one exception: the N_Elements() function now always returns the N_Elements() of the GDL variable  referenced to it; LIST and HASH objects do not get an exceptional treatment until the IDL convention  that n_elements( <a list>) will return the count() of that list (or hash).
        The built-in object GDL_CONTAINER is also new, and equivalent to IDL_CONTAINER.  In fact the terms are  rendered equivalent in the code, rather than IDL_CONTAINER being coded in script; this can be easily undone if there are difficulties.
  Each of these PRs were individually coherent and the merger proceeded easily with only a few
trivial cases of conflict resolution.  The merged commits include gdlc.g and gdlc.i.g files that
appear to be compatible with the GDLParser, GDLLexer, GDLInterpreter etc. changes.
A final commit was needed to bring the proper uniqueptr/auto_ptr code to GDLParser.hpp

The result was built with gcc on opensuse and a make check appears normal for the state
of GDL today:

97% tests passed, 7 tests failed out of 211

Total Test time (real) = 140.14 sec

The following tests FAILED:
         94 - test_fft_leak.pro (Failed)
         97 - test_file_delete.pro (Failed)
        142 - test_n_tags.pro (Failed)
        152 - test_parse_url.pro (Failed)
        175 - test_resolve_routine.pro (Failed)
        199 - test_total.pro (Failed)

Greg Jung 
gvjung@gmail.com


GillesDuvert:
Hi @maynardGK , many thanks. As there are changes in antlr-produced files I have to report them in the .g files before PR.  Especially as I'm changing the .g files now. Let me review this next week.

maynardGK:
I inspected the new stuff and found the n_elements() component of size() in
basic_fun_jmg may be wrong for LIST and HASH cases.  I have it as taking LIST_count() and
HASH_count() values; I don't know what the result is for IDL, so I can't tell if a correction was made or if I neglected this in the PRs.  If it is incorrect as it stands I can make another PR to fix it, once I get a new master to branch from.

GillesDuvert:
@maynardGK OK, let's do like that.

PR #287, opened 2018-05-30
avoid using uninitialized values that should normally be set inside a…

GillesDuvert:
…n openMP parallel whose actual number of threads is insecure!


GillesDuvert:
Bug was found using 'atv'. It is due to one of the parallell loop not executed because the remaining 'chunk' was of size 0. Proper initialization of the variables solves the problem.

PR #288, opened 2018-05-31
Optional libtiff dependency for better TIFF support

rexso:
The last few days I have been working on redoing the TIFF support in GDL, because we use GeoTIFF images a lot in our company, and it seems like there is no GeoTIFF support in GDL yet. Since the current TIFF support is backed by GraphicsMagick/ImageMagick, no good TIFF support can be provided, and GeoTIFF specifics cannot be handled at all.

My solution is to use **libtiff** and **libgeotiff** to provide native support for both TIFF and GeoTIFF files. Naturally, people who are not interested in supporting this, might not be eager to depend on these libraries, therefore both these libraries are completely optional. If the build is not compiled with this new TIFF-support, basic manipulation is still provided through the old *GM/IM* bindings. If only improved TIFF support is wanted without support for GeoTIFF, this is also possible.

For now, **QUERY_TIFF** and **READ_TIFF** support is added, and **WRITE_TIFF** will be added later in a different PR if this one is accepted. This new version supports all general query routine info structure fields and the TIFF specific fields ([reference](https://www.harrisgeospatial.com/docs/query___routines.html)), as well as the GeoTIFF structure fields ([reference](https://www.harrisgeospatial.com/docs/READ_TIFF.html)) if compiled with the optional GeoTIFF support. The only exception is the **SBIT_VALUES** from the general info structure, because I have no idea what this field is supposed to be. Apparently the *GM/IM* implementation does not support this field either, but adding support in the future should be trivial.

Please note that this implementation depends on the **C++11** standard. Refer to [this internal discussion](https://github.com/orgs/gnudatalanguage/teams/gdlteam/discussions/2) regarding this topic. I have named the header files and source code dependent on C++11 with the **.hxx** and **.cxx** extensions respectively, to make it easier to find these files (for whatever reason).

In addition, I have also created a **DStructFactory** class for easy *DStructGDL* creation, to aid in potential future additions of TIFF tags. I do not know if the GDL codebase already includes something simlar, if so, please let me know and I will port away this redundancy. For me at least, this class turned out to be very helpful when working with an arbitrary number of struct fields of various types.

Comments or questions are welcome :smile:
 
---

*Update:* **READ_TIFF** has now been added, in addition to the initial **QUERY_TIFF** support.


rexso:
The one failing CI check that fails seems to be failing on C++11 features.

Some input on how to omit C++11 specifics for non-complying compilers would be nice, since I took it from the discussion that C++11 should be okay? At least for something non-essential as this PR I suppose?

slayoo:
Big thanks! 
BTW, here's a list of previously reported TIFF-related issues: https://github.com/gnudatalanguage/gdl/labels/tiff

slayoo:
Aren't the builds failing because of lack of a C++11 compiler flag (like -std=c++11 for gcc) flag?
Perhaps this would bring some solutions: https://stackoverflow.com/questions/10851247/how-to-activate-c-11-in-cmake

Also, there seem to be something fishy happening in CMake as when it fails to detect the new libraries, it still displays the TIFF setting as ON in the summary message, e.g.:
https://travis-ci.org/gnudatalanguage/gdl/jobs/386141502#L1006

Also, the .travis.yml and appveyor.yml files would need to be modified to enable installation of the new dependencies.

HTH

rexso:
@slayoo might be, as this is on **gcc 4.8.4** which has not C++11 enabled by default. This version should be feature-complete though. If this PR is interesting, I am up for discussing how to solve this, but I do not think that I would be the right person to update neither the cmake nor the CI configurations.

slayoo:
A CMake version requirement will need to be updated as well - the C++11 feature list support seems to have been added in CMake 3.8 (https://cmake.org/cmake/help/v3.8/release/3.8.html#c-c), while we currently require 2.8: https://github.com/gnudatalanguage/gdl/blob/master/CMakeLists.txt#L12

slayoo:
And yes, of course "this PR is interesting"! :)
We'll help with updates to .travis.yml, CMakeLists.txt and appveyor,yml

rexso:
For reference, this in a comparison of outputs after running **QUERY_TIFF** on a random GeoTIFF file:
*(Note that the columns have been adjusted for readability, but the values are not modified).*

**INFO** structure:
---
- [IDL](https://gist.github.com/rexso/a7ae20e4ad7136fcc2eed53a7dec4e9c)
- [GDL (GraphicsMagick)](https://gist.github.com/rexso/38eb6fc348ac8a21efbf18cac466e89a)
- [GDL (libtiff)](https://gist.github.com/rexso/0a9ef7e6e0f0aaf099e89d727201bcb0)

**GEOTIFF** structure:
---
- [IDL](https://gist.github.com/rexso/b4c8d5e77de94b369220ea423b51039c)
- [GDL (libgeotiff)](https://gist.github.com/rexso/f5fcc70da9656de5d25909d89707bd8c)




I notice that the order of the fields is not the same, and I do not know if this could be an issue. Someone familiar with IDL would have to comment on that. The order of my fields is in the exact order as stated in the [documentation](https://www.harrisgeospatial.com/docs/query___routines.html) whereas this is not the case for the IDL output.

maynardGK:
This type of thing is ideally suited for a LINKIMAGE application; have you looked into that?  The interface
would be just as fast as compiling in. We have a very simple example that succeeds but no other instances
of it I know of.  Especially if the procedure set are all new additions, it makes sense to do things this way.

slayoo:
What would be the advantage of moving toward using LINKIMAGE for features that are part of IDL? This would break the rule of mimicking IDL whenever possible. It would also create quite a lot of work for packagers.

rexso:
@maynardGK I had not looked into LINKIMAGE before now, but after briefly reading about it [here](https://www.harrisgeospatial.com/docs/using_idl_LinkImage.html), I still do not understand how it relates to TIFF support in GDL. Please elaborate.

maynardGK:
In LINKIMAGE, one creates a shareable image containing a) the library and b) the GDL interface routines into the library.  I'm pretty sure this is how IDL manages it's add-on packages such as ENVI ...
Integration into the program is a small procedure that 1) calls linkimage and 2) defines the procedures of 
the interface.
  Such an architecture is probably not suited for aspects that are integral across the GDL code base, such as the plot stream, or for intrinsic improvements (i.e., LAMBDA) that need straight access to the internals; but  components such as MagicK, CDF, HDF are ideally suited.
  As for integration, I doubt it would look  very complicated once we have more experience with it.

GillesDuvert:
Hi, this looks like a nice step forward GDL being useful for geophysics.
The structure returned by read_tiff in presence of the GEOTIFF option (described [here](https://www.harrisgeospatial.com/docs/read_tiff.html)) indicates that the content will vary, so an automaton to make structures will be handy. I doubt his need however to be in c++11, see, e.g., the code of RESTORE for how to create arbitray structures on the fly (but agreed this can be adjusted afterwards).
As the structure tags are not fixed for GEOTIFF, the order of these tags is irrelevant. However for the structure returned in query_tiff, the order may be important since people may have written code using  the tag number and not the tag name as in `GDL> help,!x.(6)`
Concerning the discussion around LINKIMAGE, adding DLM support to GDL would be even better!

rexso:
@GillesDuvert Thanks for your comment.

My only concern regarding the INFO structure returned by the QUERY functions, is that the order is not stated to be relevant in the [documentation](https://www.harrisgeospatial.com/docs/query___routines.html). If the order was relevant, i.e. implying that users could rely on using tag numbers, the correct order should have been presented in the documentation. I do indeed return the values ordered the same way as shown, but this does not match with the order returned by IDL. I have only tried this with QUERY_TIFF in IDL, so other QUERY functions might vary. I think it is fair to say that users who in this case rely on numbers instead of field names, have faulty code.

Regarding C++11, I too believe that this implementation could be done without relying on C++11. I just think that some features of modern C++ is more pleasant to work with, so until there is a definitive conclusion regarding whether or not this should be acceptable in GDL, I would prefer not to backport the code to an older standard. We already use this PR in my company, and for me it is much easier to maintain the code if is kept this way. However, if there ever is some conclusive answer on which standard GDL should be relying on, I would be more than *happy to the porting*. After all, I do want to contribute to this great project.

rexso:
Latest commit include the implementation of **READ_TIFF**, supporting both tiled and non-tiled TIFF and GeoTIFF images of any pixel format supported by IDL, including 64-bit LONG, ULONG and DOUBLE not explicitly stated as supported in the [documentation](https://www.harrisgeospatial.com/docs/query___routines.html). Only **BITS_PER_SAMPLE** values of 1, 4, 8, 16 and 32 are explicitly stated to be supported, but since 64 is an acceptable value in *libtiff* and there are coherent types in IDL/GDL, I decided to include support for such types as well.

I have also added support for the **SUB_RECT** keyword, both for tiled and non-tiled images.

Since I currently have no TIFF images with the relevant **DOT_RANGE**, **ICC_PROFILE** and **PHOTOSHOP** tags for testing, I still have not implemented support for these keywords. Instead, all of these keywords return 0, as documented to be the default value if not present. Support should be easy to add in the future, and I have already added some comments in the code regarding the expected value for these keywords. The existing GM/IM implementation of READ_TIFF does not support these keywords either, and no one seems to be complaining about it right now. For completeness, I certainly would like to support these keywords though, but first I need some compliant TIFF images to test with.

According to the TIFF standard, planar configurations other than contiguously stored samples are neither widespread nor recommended, and since I do not have any TIFF images of such configuration, I have not prioritised to add support for this yet:
> PlanarConfiguration=2 is not currently in widespread use and it is not recom-
> mended for general interchange. It is used as an extension and Baseline TIFF
> readers are not required to support it.

Again, I would like to add support for completeness, but I think it is fair to allow for such support to be implemented later. My main focus is to support TIFF images of a widely used configuration. Regardless, **INTERLEAVE** is supported in *read_tiff.pro* (thus independent from the C++ implementation), but only for the aforementioned planar configuration of contiguously stored samples.

As a quick summary, these are the keywords currently not supported (ordered by priority):

- **DOT_RANGE**
- **ICC_PROFILE**
- **PHOTOSHOP**

Update: Added support for colour maps (palettes).
Update: Added support for CHANNELS keyword.

GillesDuvert:
Is there a c++11 flag problem on Travis?
Whatever the case, I guess adding specific compile properties to the C++11 dependent code in src/CMakeFiles.txt, such as:
`set_source_files_properties(file.cxx PROPERTIES COMPILE_FLAGS "-std=c++11")`
will be necessary ?

GillesDuvert:
And, wow! :+1: 
You do not expect people do median filtering on 3x16 bit images? Because one could make a 16 bit version of the constant-time median  filtering algo located in medianfilter.cpp

rexso:
@GillesDuvert I tried adding the suggested statements to cmake, but it does not seem to satisfy Travis, which still complains about C++11 specifics. I believe Travis has to be reconfigured as suggested by @slayoo, but I do not have the sufficient experience with neither Travis nor GDL to fix this problem myself unfortunately. I would highly appreciate it if someone else could look at this issue.

As a long-term solution, manually adding the C++11 flags to CmakeFiles probably is not the best solution anyway? I believe it would be better to either add C++11 as the required standard for building the whole project, or add some checks to leave out C++11 specific code if the compiler does not support this version. After all, TIFF support is not really critical, thus only needed for those who want to read and manipulate TIFF and GeoTIFF.

slayoo:
I'm aiming at submitting today a PR with a suggestion how we could enable C++11 in CMake

slayoo:
Tried (https://github.com/slayoo/gdl/pull/1/files) with set(CMAKE_CXX_STANDARD 11) and compilation works on all platforms/compilers but there are multiple test failures with "terminate called after throwing an instance of 'GDLException'" (https://travis-ci.org/slayoo/gdl/builds/390074105). Any obvious hints?

GillesDuvert:
No clue.. :-1:

rexso:
Support for three channel palettes (colour maps) added in latest commit, and accordingly the **R**, **G** and **B** keywords for such images.

rexso:
Support for **CHANNELS** keyword added in latest commit by moving the logics from C++ to *read_tiff.pro*.

This means that the new READ_TIFF implementation now supports all existing functionality found in the GM/IM implementation, in addition to the newly added TIFF/GeoTIFF related stuff. **Wohoo** :1st_place_medal:

slayoo:
closing and reopening to trigger tests run on appveyor

rexso:
@slayoo  Seems like your cmake additions took care of the C++11 issues in Travis :+1:

slayoo:
Somehow github claims there are over four hundreds of files changed - could you please try pulling/pushing/rebasing so to make the diff appear properly? Thanks

Also, codecov results suggests that the new features are not covered in the CI.

rexso:
Since this branch has diverted beyond trivial rebase, I created a new PR with a cleaned up version: #375

PR #293, opened 2018-06-03
as for CONTOUR, SURFACE should accept "degenerated" 3D equivalent to 2D array (Alain's rev. 4122 from SVN)

slayoo:
https://sourceforge.net/p/gnudatalanguage/svn/4122/

### DIFF
--- a/src/plotting_surface.cpp
+++ b/src/plotting_surface.cpp
@@ -75,9 +75,13 @@ namespace lib
 
       if ( nParam ( )==1 )
       {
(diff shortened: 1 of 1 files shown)


GillesDuvert:
For some reason the patch proposed is not in sync with SF content, which use a EquivalentRank() method not present in this file. Error message is also wrong. I'll open a new PR with the 'good' files.

slayoo:
Thanks (I just added here what was in the SF diff - not looked at other commits that modified the lines around afterwards)

PR #298, opened 2018-06-06
 make CONTOUR, SURFACE and SHADE_SURF accept "degenerated" N-D equival…  …ents to 2D array

GillesDuvert:


GillesDuvert:
Seems codecov is not working... these red crosses are painful.

slayoo:
I believe it shows correctly that the two new tests are not added to Makefile.am!

GillesDuvert:
You are very perceptive... I'm not familiar with codecov and I do not see where is the hint about the two test files (test_convol and test_triangulate) in codecov reports.
Making and saving a test file is something, enabling it in Makefile.am is another thing. Several of our test files do not even pass the bona fide idl test...

slayoo:
Gilles, in my understanding, the key point is that a change (either removed or added line) must have a corresponding test coverage to be "green" - either the removed code should have been executed in previous tests or new code is executed in current tests (but that is my assumption how it works, not checked!). My note about the new tests was just a quick guess - I've seen new tests added but no new entry in the test list, sorry for being not precise.

PR #303, opened 2018-06-07
Minor changes

GillesDuvert:
applied fully (cpp an g files) patch suggested by @opoplawski about unique_pointers and some SizeT.


maynardGK:
I think the Appveyor fail happens with the use of std=gnu++11 and isnan.

PR #305, opened 2018-06-08
adding OpenMP multithread support in the finite function

sflinois:
Using multithread on huge for loops to improve the finite() function speed on big arrays

### DIFF
--- a/src/math_fun_jmg.cpp
+++ b/src/math_fun_jmg.cpp
@@ -197,39 +197,58 @@ namespace lib {
   // general template
    template< typename T, bool> struct finite_helper
    {
-     inline static BaseGDL* do_it(T* src, bool kwNaN, bool kwInfinity)
-     {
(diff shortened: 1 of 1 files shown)


GillesDuvert:
@sflinois , looks good!

alaingdl:
Hey, sorry for the request but I do not understand why we have a red cross here ...
(in fact, two, due to codecov)

I would like to validate this code ! (and make more tests)

slayoo:
Alain, one red cross was for reduction of overal coverage for the project, and one for low coverage within this patch. These crosses are meant to encourage adding tests for each PR. Of course, GDL is a tricky case as the code to be covered spans both the interpreter and library routines, and our current coverage is very low. Consequently, codecov reports sometimes give  quite surprising information that adding some new library feature (and a test) you increase coverage in the intepreter, etc.

Anyhow, red cross for codecov (or codacy) is not a no-go by any means! - just an encouragement to add more tests.

PR #313, opened 2018-06-12
OpenMP not found and info on brew package for clang supporting OpenMP on OSX

sflinois:


sflinois:
I am not very familiar with Cmake yet
If you have any idea how to find the LLVM package to use their Clang instead of the default Clang on OSX, fell free to share.

alaingdl:
OK, we have to go ahead to help end-users (no openmp by default on OSX)
but we need to have a better solution (detection of Llvm in Cellar if any)

slayoo:
@acoulais @sflinois this causes clang builds to fail on Travis, we have a failing master now

slayoo:
thanks for reverting!
this should perhaps help: https://github.com/sflinois/gdl/pull/1

slayoo:
Re " but we need to have a better solution (detection of Llvm in Cellar if any)" - good point!
Why not include a copy-paste ready solution for OSX users in the error message (i.e. including the CMake flags needed to pick the alternative compilers)

slayoo:
a lot of "libc++abi.dylib: terminating with uncaught exception of type GDLException" errors :( (in sflinois#1)
Just a guess, but perhaps there is a libstdc++ issue (the gcc one used instead of the one from clang?)

sflinois:
I am working on it and trying to find a solution ! :)

slayoo:
Merci!

sflinois:
I tried finding info on how to automate the search for different compiler on OSX and Linux in CMake without much success. The solution is usually to manually change the compiler.

We can still force the CMake to use the LLVM Clang with something like that:
(homebrew llvm should always have the latest clang in /usr/local/opt/llvm/bin/)
(you can get the path with brew --prefix llvm)
```
if (APPLE)
 if(EXISTS "/usr/local/opt/llvm/bin/clang++")
   set (CMAKE_CXX_COMPILER "/usr/local/opt/llvm/bin/clang++")
   set (CMAKE_C_COMPILER "/usr/local/opt/llvm/bin/clang")
 endif ()
endif(APPLE)
```
note that this need to be implemented before the 'project(GDL)' line in the CMakeLists.txt
this should allow us to remove the line 
`if [[ $DEPS == 'FULL' && $TRAVIS_OS_NAME == 'osx' ]]; then export CMAKE_ARGS="${CMAKE_ARGS} -DCMAKE_CXX_COMPILER=/usr/local/Cellar/llvm/6.0.0/bin/clang++ -DCMAKE_C_COMPILER=/usr/local/Cellar/llvm/6.0.0/bin/clang"; fi 
`
from .travis.yml

alaingdl:
thanks for the very important trick "before" ! I was testing yesterday without success !!!

slayoo:
@sflinois @acoulais I sincerely oppose hardcoding compiler paths in CMake :)
This will only strike back at us in the future as it breaks the least surprise principle - no one expects it to do so. Moreover, I do not understand why would it be needed. With the added diagnostic message, a user will have clear choice between not using OpenMP or using OpenMP by changing to another compiler.

Please also note that all the discussion is in no way APPLE-limited - the same concerns clang on Linux and Windows. All the more, I suggest not to differentiate the behaviour between different OSes.

slayoo:
and the lines in Travis should stay in my opinion also - because we do want to test both OpenMP=ON and OpenMP=OFF

sflinois:
I understand
changing your CXX and CC environment var also seems works
Should we inform the user that they can change their var using
```
export CXX=/usr/local/opt/llvm/bin/clang++
export CC=/usr/local/opt/llvm/bin/clang 
```
if they want to permanently use another compiler ?


also
> and the lines in Travis should stay in my opinion also - because we do want to test both OpenMP=ON and OpenMP=OFF 

The idea was to not have to specify the compiler and to find it when OpenMP was ON
this line was still there
`    - if [[ $DEPS == 'FULL' && $TRAVIS_OS_NAME == 'osx' ]]; then export BREW_PACKAGES="${BREW_PACKAGES} llvm"; fi `

But I agree that trying to change the compiler in the CMake is probably a bad idea

slayoo:
I would suggest sticking to -DCMAKE_... settings in the diagnostic hints - just for the sake of not interfering with other packages (imagine plplot suggests export CXX=...)

sflinois:
Alright, I understand :)

PR #319, opened 2018-06-12
patch for size(): n_elements(<list>)

maynardGK:
For l a list or hash, it is IDL convention that n_elements() return, for scalar lists or hashes, the count of the object instead of the actual # of elements in the variable passed. This was
implemented in n_elements but was neglected in size();

GDL> l=list(indgen(5),/ex)
GDL> print,n_elements(l)
           5
GDL> print,size(l)
           1           5          11           5
GDL> l=list(indgen(5),/ex) & l2=replicate(l,2)
GDL> print,size(l2)
           1           2          11           2
GDL> help,l,l2
L               LIST      <ID=1  N_ELEMENTS=5>
L2              OBJREF    = Array[2]

### DIFF
--- a/src/basic_fun_jmg.cpp
+++ b/src/basic_fun_jmg.cpp
@@ -39,6 +39,8 @@ namespace lib {
 
   using namespace std;
   using namespace antlr;
+  SizeT HASH_count( DStructGDL* oStructGDL);
+  SizeT LIST_count( DStructGDL* oStructGDL);
  
   BaseGDL* isa_fun( EnvT* e) 
   {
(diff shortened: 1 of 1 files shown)


alaingdl:
I do accept this useful patch but I would happy to have some tests added into "test_size.pro"
(just to avoid any future unwanted regression !)

couls you consider to add some tests (even very basic ones !) in "test_size.pro" ? thanks !

maynardGK:
@GillesDuvert this patch, and another for container), became reverted when I merged file_etc.

PR #322, opened 2018-06-13
OpenMP option in CMake

sflinois:
removed the incorrect information on Debian/Ubuntu compiler package
and changed the homebrew clang path in .travis.yml


sflinois:
I'm not sure why Travis build failed.
Do I need to close and re-open the PR to trigger the tests or am i missing something ?

slayoo:
The only two builds affected (those using the new clang) have 20 test failures:
```
The following tests FAILED:
	 16 - test_bug_2974380.pro (Failed)
	 41 - test_bug_3275334.pro (Failed)
	 79 - test_congrid.pro (Failed)
	 85 - test_device.pro (Failed)
	 88 - test_execute.pro (Failed)
	 89 - test_extra_keywords.pro (Failed)
	 95 - test_file_lines.pro (Failed)
	107 - test_hash.pro (Failed)
	108 - test_hdf5.pro (Failed)
	115 - test_indgen.pro (Failed)
	123 - test_list.pro (Failed)
	130 - test_moment.pro (Failed)
	141 - test_plot_oo.pro (Failed)
	144 - test_pmulti_basic.pro (Failed)
	148 - test_ps_decomposed.pro (Failed)
	155 - test_qhull.pro (Failed)
	158 - test_random.pro (Failed)
	169 - test_spawn_unit.pro (Failed)
	182 - test_tv.pro (Failed)
	186 - test_window_background.pro (Failed)
```

It seems all of them fail with:
```
libc++abi.dylib: terminating with uncaught exception of type GDLException
```

Do you have access to OSX where this could be reproduced?

sflinois:
We can't reproduce the cases be we have ideas where this is coming from.
We will work on it

slayoo:
Thank you!

Actually, I've just noticed that the set of errors failing matches those that fail when GDL is compiled with -std=c++11 !!!

See:
https://github.com/slayoo/gdl/pull/1
https://travis-ci.org/slayoo/gdl/jobs/390074110
```
The following tests FAILED:
	 16 - test_bug_2974380.pro (Failed)
	 41 - test_bug_3275334.pro (Failed)
	 79 - test_congrid.pro (Failed)
	 84 - test_device.pro (Failed)
	 87 - test_execute.pro (Failed)
	 88 - test_extra_keywords.pro (Failed)
	 94 - test_file_lines.pro (Failed)
	106 - test_hdf5.pro (Failed)
	112 - test_indgen.pro (Failed)
	126 - test_moment.pro (Failed)
	135 - test_plot_oo.pro (Failed)
	138 - test_pmulti_basic.pro (Failed)
	142 - test_ps_decomposed.pro (Failed)
	148 - test_qhull.pro (Failed)
	151 - test_random.pro (Failed)
	162 - test_spawn_unit.pro (Failed)
	175 - test_tv.pro (Failed)
	179 - test_window_background.pro (Failed)
```

Moreover, there the problem exists not only for clang, but also for gcc and icc!

slayoo:
Wait, that might be the point!!! New clang's default setting is probably C++11 (or newer) instead of C++98. 
Please try adding this to the main CMakeLists.txt:
```
cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
cmake_policy(SET CMP0026 OLD) # LOCATION property
set(CMAKE_CXX_STANDARD 98)
```

slayoo:
Interesting, test_binfmt.pro started failing on Windows...

slayoo:
I've restarted the Appveyor build to check if it's deterministic

slayoo:
yes, deterministic :(
Let me suggest disabling the binfmt test for now (see the before_build: section in appveyor.yml).

Also, perhaps we could add to the CMake diagnostic message some hint that disabling OpenMP will incur significant performance penalty?

sflinois:
I added a small message in CMakeLists.txt and try disabling the test on 32bit architecture (we believe this is the reason the test failed)

slayoo:
it's certainly not the reason - this test is passing on appveyor without the -std=c++98 flag plus it passes OK on 32 bit IDL

sflinois:
Huum, we (alain and I) tried the binfmt test on a 32-bit Debian and we found out the tests were failing too, see [https://github.com/gnudatalanguage/gdl/issues/332](https://github.com/gnudatalanguage/gdl/issues/332)

slayoo:
still, if it runs ok on IDL there is no reason to disable it from within the test, but rather for CI controls (i.e. in the appveyor file). The fact that only adding -std=c++98 makes this test to fail on some platforms might be a hint for fixing also...

sflinois:
Ok, i'll remove the changes to test_binfmt.pro
I don't know how to properly remove the test from appveyor doing guess something like 
```
before_build:
  - md c:\projects\gdl\build
  - sed -i 'test_binfmt.pro/d' c:\projects\gdl\testsuite\Makefile.am
```
would work but it is not very clean

slayoo:
thanks
there are already more than 100 tests disabled for appveyor in the appveyor.yml file - please just copy paste one of those lines and change the test name
thanks again

sflinois:
Ooh, my appveyor.yml was not up to date and i was confused, sorry about that

slayoo:
merci!

sflinois:
I just have few questions about those lines :
```
cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
cmake_policy(SET CMP0026 OLD) # LOCATION property
set(CMAKE_CXX_STANDARD 98)
```
The first line require cmake 3.1 so that we can add the next line
`cmake_policy(SET CMP0026 OLD) # LOCATION property`
and this line allow us to use the LOCATION properties
but I just don't know where and how this LOCATION propertie is used
If there is an easy way to explain it, i'm willing to learn ! :)

I assume that C++11 and above are problematic in some cases with GDL so we force the compiler to use C++98 with :
`set(CMAKE_CXX_STANDARD 98)`

slayoo:
Sylvain, the chain of requirements is as follows:
- wihle using newer clang C++11 is enabled by default
- apparently GDL source is not compatible with C++11 (problems at runtime) 
- we need to enforce C++98 in a way that would work for all compilers and OSes -> we want to use CMAKE_CXX_STANDARD
- CMAKE_CXX_STANDARD was introduced in CMake 3.1 -> we want to require it
- requireing CMake 3.1 makes CMake force us to decide if CMP0026 behaviour is set to OLD or NEW
- we set it to OLD with cmake_policy()

The LOCATION thing is completely unrelated, IIUC it is a CMake target property that we read within the CMake scripts to better support out-of-tree builds.

alaingdl:
This change in CMakeLists.txt broke the compilation on my laptop, where I have ubuntu 14.04.5 LTS which is maintained till mid 2019. I don't plan to upgrade.

cmake --version : 
cmake version 2.8.12.2

current git GDL compile without trouble if I change these  lines

> cmake_minimum_required(VERSION 3.1 FATAL_ERROR)
> cmake_policy(SET CMP0026 OLD) # LOCATION property
> set(CMAKE_CXX_STANDARD 98)

 by

cmake_minimum_required(VERSION 2.8 FATAL_ERROR)

slayoo:
Ubuntu 14 has the cmake3 package - should help!

alaingdl:
Ha ! OK ! thanks !
tested with success !

PR #328, opened 2018-06-14
Patch help, includefirst.hpp for WIN32

maynardGK:


slayoo:
could any of the failing tests be uncommented thanks to it?

PR #331, opened 2018-06-14
 Interim solution with widget_draw option.

maynardGK:
Current state of affairs is that "WX" was supposed to unify plotting and widgets, but
    the plplot driver for WX is not quite as capable as a plain X driver or Win driver for the
    simple plots.  However incorporating a plot  or image into a widget group requires this. To do so,
    set GDL_USE_WX=YES in the environment.

      If you avoid plotting or images, you can have your widgets and your plots by building GDL
    with "NO_WIDGET_DRAW" macro set.  If this is so, "X" or "WIN" plot drivers and widgets will coexist,
    but widget_DRAW will return a widget ID of -1; the widget routine is prepared for this.
    Under batch build conditions neither of these will be set.  OSX/Clang build fails a simple window
    creation test when wx initialization occurs, and so the initialization, which is imperative for widget operation, is hidden in these two conditions.

      The UnsetFocus() is restored and plotting can be tested by running testsuite/testfocus
    which will put up 5 plots and iconize two of them.
    demo_widgets.pro and test_widgets.pro are both conditioned to expect a non-working widget_draw.

### DIFF
--- a/src/devicewin.cpp
+++ b/src/devicewin.cpp
@@ -412,11 +412,10 @@ bool DeviceWIN::WShow(int ix, bool show, int iconic)
 	if (ix >= wLSize || ix < 0 || winList[ix] == NULL) return false;
 
   if (iconic!=-1) { //iconic asked. do nothing else.
-    if (iconic==1) winList[ix]->Iconic();  else winList[ix]->DeIconic();
-    return true;
+		if (iconic==1) IconicWin(ix); else DeIconicWin(ix);
+	} else {
+		if (show) RaiseWin(ix);  else LowerWin(ix);
   }
-	if (show) winList[ix]->Raise();      else winList[ix]->Lower();
-
 	UnsetFocus();
 
 	return true;
--- a/src/devicewin.hpp
+++ b/src/devicewin.hpp
@@ -69,7 +69,7 @@ private:
 	bool WState(int);
 	bool WSize(int, int*, int*);
 	bool WSet(int);
-	bool WShow(int, bool, bool);
+	bool WShow(int, bool, int);
 	int WAdd();
 	DIntGDL* GetWindowPosition();
 	DLong GetVisualDepth();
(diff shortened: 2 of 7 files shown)


GillesDuvert:
@maynardGK Reviewing this, I see that you say "_However incorporating a plot  or image into a widget group requires this. To do so, set GDL_USE_WX=YES in the environment_."
This is not so. 
Indeed, "canvas" widgets use WX, as the widget support in GDL is exclusively done with WxWidgets. But it is not necessary to "_set GDL_USE_WX=YES in the environment_" for that. GDL lives with 2 graphic outputs if widgets are present: (WX and X11) or( WX and WIN). WIN or X11 make the GDL graphic windows (command WINDOW), WX the graphic canvas of WIDGET_DRAW. So there should be no provision for absent WxWidgets, as all the WIDGET_xxx functions are deactivated in this case. Conversely if GDL has been linked with WxWidgets, ALL the widgets are present, including WIDGET_DRAW, which knows to plot images.

**If this is not the case on a particular implementation, it is a bug and deserves a bug report and proper patch.**

Now, about defining GDL_USE_WX=YES
This is to turn on an experimental feature that makes WINDOW commands to use WxWidgets too, so "normal" plots are in a WxWidget window. This, if it was devoid of bugs---or plplot driver features---would have the advantage of 1)  having only one driver and only one device to support and 2) all the nice WxWidgets features such as antialiasing, alpha channel, TrueType fonts...
Conversely, if WxWidgets was able to create an "empty" drawing region in a widget, a pixmap where we could plot using the x11 driver (as IDL does), it would also make our life easier. But AFAIK this is not possible.

maynardGK:
@GillesDuvert for X11 the situation is much better, but it is my experience that use of the widget_draw while the plot device is "WIN', does not work well - hardly at all.  When the plplot device driver is also WX,
however, the widget behavior is as good as it gets, but the window plotting suffers.  So, the patch allows the best behavior of widgets and windows to co-exist, at the sacrifice of the widget_draw routine.

The only bug I found here was, that OSX with clang compiler does not tolerate wxWidget initialization.

Issue #278, opened 2018-05-21
test_save_restore fails with GCC 8


opoplawski:
With gcc 8 I get:
```
test 157
        Start 157: test_save_restore.pro
157: Test command: /builddir/build/BUILD/gdl-0.9.8/build/testsuite/launchtest "test_save_restore.pro"
157: Test timeout computed to be: 10000000
157: % Compiled module: TEST_SAVE_RESTORE.
157: % Compiled module: GDL_IDL_FL.
157: % Compiled module: GIVE_LIST_NUMERIC.
157: % Undefined item not saved: LONG64_S
157: % Undefined item not saved: LONG64_S_A1D
157: % Undefined item not saved: LONG64_S_A2D
157: % Undefined item not saved: ULONG64_S
157: % Undefined item not saved: ULONG64_S_A1D
157: % Undefined item not saved: ULONG64_S_A2D
157: TEST EXITED FROM SIGNAL 11
157/185 Test #157: test_save_restore.pro ..............***Failed    0.20 sec
```
I believe that this is caused by not having a return() statement as noted in the warning here:
```
/builddir/build/BUILD/gdl-0.9.8/src/saverestore.cpp:166:14: warning: unused variable 'cur' [-Wunused-variable]
     uint32_t cur=writeNewRecordHeader(xdrs, 6);
              ^~~
/builddir/build/BUILD/gdl-0.9.8/src/saverestore.cpp:167:3: warning: no return statement in function returning non-void [-Wreturn-type]
   }
   ^
```
This is the patch I came up with:
```
--- gdl-0.9.8/src/saverestore.cpp.return        2018-03-26 03:57:27.000000000 -0600
+++ gdl-0.9.8/src/saverestore.cpp       2018-04-11 16:33:39.129920693 -0600
@@ -164,6 +164,7 @@ namespace lib {

   uint32_t writeEnd(XDR *xdrs) {
     uint32_t cur=writeNewRecordHeader(xdrs, 6);
+    return cur;
   }

   int getVersion(XDR* xdrs) {
```
Which resolves the segfault.  However, with this patch, test_save_restore never returns on s390x.

alaingdl:
asking for a VM with Fedora 28 in my lab. :)

alaingdl:
in fact, I can't compile on F28, because of change in XDR/RPC (seehttps://github.com/gnudatalanguage/gdl/issues/306)

GillesDuvert:
Still there?

[...]

PR #338, opened 2018-06-14
Added support for tirpc as replacement for SunRPC

rexso:
After updating glibc, GDL suddenly failed linking due to missing xdr references.  Apparently, SunRPC has been removed from glibc since version **2.26**, so newer systems should rely on **tirpc** instead ([reference](https://fedoraproject.org/wiki/Changes/SunRPCRemoval)). SunRPC is *optional* in earlier glibc so the more modern and potentially faster tirpc is preferable anyway.

This patch merely adds the tirpc directory as a prefix during path lookup in cmake, and tries to link the relevant library if found.

Am I the only one who had this problem?

Sure, it worked fine until yesterday since I compiled glibc with rpc built-in, and this might be the case for most distributions, but still it seems like tirpc is preferred. I do not have any Windows or OSX box to test this patch on, but I suppose the CI tests will catch any errors. This fixed the linking errors on my Gentoo system without SunRPC and still builds fine on both newer and older Ubuntus, as well as on Arch at least.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -356,6 +356,22 @@ else(GSL_FOUND)
         "(suitable Fedora package: gsl-devel)")
 endif(GSL_FOUND)
 
+# RPC MANDATORY
+# -DRPCDIR=DIR
+set(CMAKE_PREFIX_PATH ${RPCDIR})
+find_package(RPC QUIET)
+set(HAVE_RPC ${RPC_FOUND})
+if(RPC_FOUND)
+	set(LIBRARIES ${LIBRARIES} ${RPC_LIBRARIES})
+	include_directories(${RPC_INCLUDE_DIR})
+else(RPC_FOUND)
+	message(FATAL_ERROR "RPC support is mandatory.\n"
+	"Note that SunRPC has been removed in glibc-2.26 and later, "
+	"while being optional in earlier versions. Consider using the "
+	"recommended and more modern libtirpc instead.\n"
+	"Use -DRPCDIR=DIR to specify the rpc directory tree.\n")
+endif(RPC_FOUND)
+
 if(WIN32 AND NOT CYGWIN)
 set(CMAKE_PREFIX_PATH ${XDRDIR})
 find_package(Xdr)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -356,46 +356,44 @@ else(GSL_FOUND)
         "(suitable Fedora package: gsl-devel)")
 endif(GSL_FOUND)
 
-# RPC MANDATORY
-# -DRPCDIR=DIR
-set(CMAKE_PREFIX_PATH ${RPCDIR})
-find_package(RPC QUIET)
-set(HAVE_RPC ${RPC_FOUND})
-if(RPC_FOUND)
-	set(LIBRARIES ${LIBRARIES} ${RPC_LIBRARIES})
-	include_directories(${RPC_INCLUDE_DIR})
-else(RPC_FOUND)
-	message(FATAL_ERROR "RPC support is mandatory.\n"
-	"Note that SunRPC has been removed in glibc-2.26 and later, "
-	"while being optional in earlier versions. Consider using the "
-	"recommended and more modern libtirpc instead.\n"
-	"Use -DRPCDIR=DIR to specify the rpc directory tree.\n")
-endif(RPC_FOUND)
+if(UNIX)
+	set(CMAKE_PREFIX_PATH ${RPCDIR})
+	find_package(RPC QUIET)
+	set(HAVE_RPC ${RPC_FOUND})
+	if(RPC_FOUND)
+		set(LIBRARIES ${LIBRARIES} ${RPC_LIBRARIES})
+		include_directories(${RPC_INCLUDE_DIR})
+	else(RPC_FOUND)
+		message(FATAL_ERROR "RPC support is mandatory.\n"
+		"Note that SunRPC has been removed in glibc-2.26 and later, "
+		"while being optional in earlier versions. Consider using the "
+		"recommended and more modern libtirpc instead.\n"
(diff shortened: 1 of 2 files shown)


alaingdl:
thanks ! that is related to #338 and #278

slayoo:
Windows build failed with:
```
CMake Error at CMakeLists.txt:378 (message):
  RPC support is mandatory.
  Note that SunRPC has been removed in glibc-2.26 and later, while being
  optional in earlier versions.  Consider using the recommended and more
  modern libtirpc instead.
  Use -DRPCDIR=DIR to specify the rpc directory tree.
```

rexso:
I suppose this is easily fixed by restricting the RPC check to UNIX systems.
Windows is already covered by xdr check after all. :smile:

slayoo:
Just to make sure, would this change mandate changing the setting for MINI and FULL builds on Travis (i.e. installation of a new package or enable/disable flag?) Asking as it would be best to have as much of CMake code branches covered on Travis.

GillesDuvert:
in my distro, `/usr/include/rpc/xdr.h` is provided by glibc and `/usr/include/tirpc/rpc/xdr.h` by libtirpc-devel. pkg-config works:
```
$ pkg-config --cflags libtirpc
-I/usr/include/tirpc 
$ pkg-config --libs libtirpc
-ltirpc 
```

slayoo:
Ubuntu and Homebrew have tirpc packaged - would be great to add this dependency to the FULL variant in .travis.yml then. Something along the lines of:
```
    # rpc
    - if [[ $DEPS == 'FULL' && $TRAVIS_OS_NAME == 'linux' ]]; then export APT_PACKAGES="${APT_PACKAGES} libtirpc-dev"; fi
    - if [[ $DEPS == 'FULL' && $TRAVIS_OS_NAME == 'osx' ]]; then export BREW_PACKAGES="${BREW_PACKAGES} libtirpc"; fi
    - if [[ $DEPS == 'FULL' && $TRAVIS_OS_NAME == 'osx' ]]; then export CMAKE_ARGS="${CMAKE_ARGS} -DRPCDIR=..."; fi  
    - if [[ $DEPS == 'FULL' && $TRAVIS_OS_NAME == 'linux' ]]; then export CMAKE_ARGS="${CMAKE_ARGS} -DRPCDIR=..."; fi  
```

With the ellipses replaced with the required paths - /usr/include/tirpc/rpc/ for Ubuntu? and /usr/local/... for brew?

The aim being to have both libc and tirpc variants tested on Travis.

rexso:
Indeed many distros still provide glibc with SunRPC, however the trend is to port away this built-in support in favour of libtirpc. If available, tirpc is preferred with glibc as fallback. Regardless, some RPC implementation is required.

@slayoo Ellipses should be replaced with **/usr/include/tirpc/** without the final *rpc/* part. The include statement in the source code is still **<rpc/rpc.h>**. Not sure about brew and OSX directory structure.

slayoo:
https://github.com/Linuxbrew/homebrew-core/blob/master/Formula/libtirpc.rb
Nothing special there, so I guess /usr/local/include/tirpc

slayoo:
oops, actually this is Linuxbrew not homebrew link!

slayoo:
anyhow, for now let's skip OSX - just please add if possible the conditional tirpc setting for the FULL builds on Linux (to have a minimal CI coverage)

alaingdl:
Not sure to understand the discussion !

rpc is removed from core gcc in gcc 8+

with Sylvain, we are testing now how it is going on Fedora28 (which has already gcc8)
and also if side effects ...

alaingdl:
Unfortunately, the link failed on FC28

100% ... 
gdl /usr/bin/ld: CMakeFiles/gdl.dir/CFMTLexer.cpp.o: relocation R_X86_64_32S against symbol `_ZTV9CFMTLexer' can not be used when making a shared object; recompile with -fPIC

[ ... ]

Linking CXX executable /usr/bin/ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
make[2]: *** [src/CMakeFiles/gdl.dir/build.make:2230: src/gdl] Error 1
make[1]: *** [CMakeFiles/Makefile2:125: src/CMakeFiles/gdl.dir/all] Error 2
make: *** [Makefile:141: all] Error 2

rexso:
@alaingdl looks unrelated to this PR. What compiler is that? Seems like it is not generating position independent code by default, and the solution is as reported by the error: to recompile using the -fPIC flag

alaingdl:
@rexso why unrelated ? If I try your files on my Fedora28 I cannot go to the end at compilation.

gcc version 8.1.1 20180502 (Red Hat 8.1.1-1) (GCC)

alaingdl:
success on Fedora28 but need to do : 

cmake .. -DCMAKE_CXX_FLAGS=-fPIC

rexso:
Is it specifically when this patch is applied?

If tirpc is detected, it tries to link the tirpc library, indeed, but the codebase is dynamically linked anyway.

The brief output you provided states linking failure in CFMTLexer.cpp.o and I believe there is no RPC related code there. Therefore I do not think this linking error is related to this PR per se. This patch only sets the correct include paths and adds the tirpc library to the list of libraries to be linked. There is no modification to compiler flags here.

If this patch makes building fail on certain systems, it must be due to either the headers or the library not being found.

rexso:
Yeah, regarding that -fPIC flag, we might have to create another issue so we can look into adding this flag explicitly. Most distributions nowadays make gcc build with this flag by default when dynamically linking, but this is not always the case.

alaingdl:
@rexso I cannot really answer because it is only on Fedora28 up to now

elbarto51:
I have seen that with this patch the problem that I currently have is corrected. but I'm a newbie, so what are the steps to apply this patch? Thank you

Issue #345, opened 2018-06-14
gcc warning: format ‘%x’ expects argument of type ‘unsigned int’


slayoo:
```
envt.cpp:1325:44: warning: format ‘%x’ expects argument of type ‘unsigned int’, but argument 4 has type ‘const void*’ [-Wformat=]
     ix,static_cast <const void *>(env[ix]) );
                                            ^
```

slayoo:
(e.g. here: https://travis-ci.org/slayoo/gdl/jobs/392357355)

rexso:
Fixed in #346

PR #346, opened 2018-06-14
Patching #343 and #345

### LINKED ISSUE #343: warning: 'register' storage class specifier is deprecated and incompatible with C++1z
```
basic_fun.cpp:4063:11: warning: 'register' storage class specifier is deprecated and incompatible with C++1z [-Wdeprecated-register]
          ELEM_SWAP(array[ll], array[hh])
          ^
/Users/travis/build/slayoo/gdl/src/basic_fun.cpp:4026:26: note: expanded from macro 'ELEM_SWAP'
#define ELEM_SWAP(a,b) { register DFloat t=(a);(a)=(b);(b)=t; }
                         ^
```

rexso:
This PR should fix #343 and #345.

### DIFF
--- a/src/basic_fun.cpp
+++ b/src/basic_fun.cpp
@@ -4057,7 +4057,7 @@ namespace lib {
  *  Modified by G. Duvert, 2017, for NaN/INF handling and correction of Nicolas's code
  *  which gave erroneous results when two or more elements were identical. 
  */
-#define ELEM_SWAP(a,b) { register DDouble t=(a);(a)=(b);(b)=t; }
+#define ELEM_SWAP(a,b) { DDouble t=(a);(a)=(b);(b)=t; }
   
   DDouble quick_select_d(DDouble array[], SizeT arraySize, int even) {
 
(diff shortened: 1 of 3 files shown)


slayoo:
@GillesDuvert Gilles, any concerns about removing the register thing?

rexso:
Nothing to worry about :smile_cat: 

According to the C++ standard, section *7.1.1, Storage class specifiers*:
> A register specifier is a hint to the implementation that the variable so declared will be heavily used. [Note: The hint can be ignored and in most implementations it will be ignored if the address of the variable is taken. This use is deprecated (see D.2). —end note]

The reality is that modern C++ compilers completely ignore this hint, and many of them always did. Since C++11 this specifier has been *deprecated*, and it has been *removed* since C++17.

Even if you could find a compiler that actually honours this hint and is not capable of doing such optimisations automatically, worst case scenario would be a time penalty no one would ever notice.

rexso:
Travis job 9 timed out? :open_mouth: 
False positive?

slayoo:
restarted!

rexso:
What is the implication of the failing codecov/patch check?

slayoo:
no implication, just a hint that the code patched is not yet executed anywhere in the testsuite :(

rexso:
Ah, I see.. Thank you for the clarification.
At least the compiler output should be a bit nicer with this patch though :smile_cat:

GillesDuvert:
yes, register is ignored by compilers. No worry.

PR #355, opened 2018-06-15
small patches to help tests.

maynardGK:
test_hash and test_list now need to recognize "git" in the release name in order that
some already-known defects do not cause failures.  file_which(filename) is employed
to get a file when user is not on  its directory.

### DIFF
--- a/src/pro/file_delete.pro
+++ b/src/pro/file_delete.pro
@@ -80,7 +80,7 @@ pro FILE_DELETE, files0,  files1,  files2,  files3,  files4,  files5,  files6,  
                  allow_nonexistent=allow_nonexistent, noexpand_path=noexpand_path, recursive=recursive, $
                  quiet=quiet, verbose=verbose
 
-  if ( strupcase(!version.os_family) eq 'WINDOWS') then  message,"Sorry, FILE_DELETE is not available on Windows (FixMe!)"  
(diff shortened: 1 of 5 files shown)


slayoo:
@maynardGK Why do we need the "isgit" logic? If this gets into a release, the tests will be skipped for no good reason, right?

maynardGK:
@slayoo There is another major edit coming to both hash.cpp and list.cpp, as planned to leave a history of "transitional" files. I haven't spent enough time on it, recently, to say if they are 100 or 99% ready or just 95%.  So the ~isgit allows a pass to what is expected - running with the new stuff or from from IDL will go beyond that and even more rigorous tests are invited.

alaingdl:
@maynardGK sorry but I changed "test_indgen.pro" after you and commit it.
I don't know how to resolve the conflict

maynardGK:
The system will present it as a conflict when the merge is done - easy.

slayoo:
@alaingdl after fixing and committing you can also open a pull request to Greg's branch (maynardGK:fix_tests). This way - after Greg merges it - we will have CI results before merging into master.

maynardGK:
Looks like you're making it to not need an external file, which is fine - my change was only to pick that file up from file_which() rather than assuming it is in current directory.

Issue #344, opened 2018-06-14
runtime problems after compiling with -std=c++11 (pb around DInterpreter::NoReadline() with stdin set to /dev/null)


slayoo:
In an attempt to switch C++11 on by default in GDL, it became apparent that the source code is not actually compatible with the newer standard. Compilation goes fine, while there are runtime crashes in all tested configurations, regardless of the compiler: gcc, clang and icc.

The runtime symptoms are:
- "libc++abi.dylib: terminating with uncaught exception of type GDLException" (on OSX)
- "terminate called after throwing an instance of 'GDLException'" (on Linux)

The same problems were actually appearing without switching c++11 flag, but by using new enough version of the compiler which had C++11 enabled by default. To fix this, currently it is set using CMAKE_CXX_STANDARD to enforce C++98 during compilation.

rexso:
Do you have anything more descriptive than this termination message? A stack trace maybe?
I have never compiled GDL with compilers older than gcc 6.1, where C++14 is the default standard.

slayoo:
got it!

```
(gdb) run -quiet -e test_file_lines < /dev/null
The program being debugged has been started already.
Start it from the beginning? (y or n) y
Starting program: /home/slayoo/devel/gdl/build/src/gdl -quiet -e test_file_lines < /dev/null
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
% Compiled module: TEST_FILE_LINES.
% FILE_LINES: Could not open file for reading 
% Execution halted at: TEST_FILE_LINES      3 /home/slayoo/devel/gdl/testsuite/test_file_lines.pro
%                      $MAIN$          
[New Thread 0x7ffff45cb700 (LWP 18413)]
terminate called after throwing an instance of 'GDLException'

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff52d3231 in __GI_abort () at abort.c:79
#2  0x00007ffff5c8d5a9 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
[...]

PR #365, opened 2018-06-15
making C++11 required

slayoo:


slayoo:
Fixes #344

slayoo:
@rexso - this one should help in #288 
@GillesDuvert @alaingdl any objections against merging? (and making GDL require a C++11 compiler)

alaingdl:
Hum, not happy because I still have two machines in CentOS 5x (5.11), where we do use GDL. One is a more than 1000 cores with half a Po storage. No plan at all to migrate. Quite sure not to have a c++11 compiler here.

I can compile GDL today on both, just by removing one call to OpenMP

slayoo:
Would the devtools package help: https://unix.stackexchange.com/questions/63587/how-to-install-g-4-7-2-c11-on-centos-5-x (never tried!)

rexso:
This is very interesting, good job @slayoo. It would be really nice to have some explicit standard requirements, and I believe that supporing at least C++11 will benefit GDL quite a lot in future code.

Of course, supporting current users is important, but since there has been no explicit C++ standard requirement until now, it is difficult to map the needs of all users. However, it looks like the [internal standard discussion](https://github.com/orgs/gnudatalanguage/teams/gdlteam/discussions/2) speaks in favour of supporting C++11.

What version of GCC are you using on CentOS 5.11, @alaingdl? GCC has been C++11 feature-complete since version 4.8.1, which was released in May 2013, so it had been around for more than a year when CentOS 5.11 was released in September 2014. As a side note, CentOS 5 updates were dropped a year ago, whereas CentOS 6 is on LTS until 2020. CentOS 7 has been available since 2014. Anyway, if the currently available version of GCC does not support C++11, is it not an option to upgrade GCC at least?

I think it is fair to have a standard requirement, even if pre C++11 should be decided. When no standard is explicitly defined, developers do not really have any guidelines of what is acceptable. If my vote counts, I give my vote to support this PR. :+1:

slayoo:
Frankly speaking, I assume that compiling GDL on outdated systems anyhow requires manual compilation of most of the build-time and run-time dependencies, right?. Building the compiler should not add much more trouble then! 

Moreover, this can be automated and made quite easy with Gentoo Prefix (https://wiki.gentoo.org/wiki/Prefix) - an automated workflow for building local Gentoo environment on a user account without root privilages. I did try it with GDL and it does work and solves any outdated software issues.

Let's go for C++11!

GillesDuvert:
@alaingdl with the evolution of gdl, always some old machine will definitely been unable to recompile it. We can suppose that the gdl running ont it will still run as it is until it will be shut down, probably more by lack of hardware an system maintainers than by not being able to run the latest version of this humble software :smile:

GillesDuvert:
This said, the CMAKE minimal version should be 3.0.2 and not 3.1? Works for me with 3.0.2. No need to impose more constraints than necessary...

slayoo:
This depends on the OS, on OSX it should actually be 3.2 for it to work:
https://public.kitware.com/Bug/bug_relationship_graph.php?bug_id=15355

slayoo:
CMake docs:
- 3.0.2: https://cmake.org/cmake/help/v3.0/search.html?q=CMAKE_CXX_STANDARD -> no results
- 3.1: https://cmake.org/cmake/help/v3.1/search.html?q=CMAKE_CXX_STANDARD -> present

slayoo:
yet... if it does work, no objections to just set 3.0 for Linux and 3.2 for other platforms...

GillesDuvert:
Actually with all those changes it dos not pass at all with my mageia 5 gcc 4.9.2.
Putting on hold waiting for further tests.

slayoo:
What are the errors? There are hardly any changes here!

GillesDuvert:
OK, my mistake. Works, except that:
Previous to cmake 3.1, you need to tell cmake to use specifically c++11 by  adding `-DCMAKE_CXX_FLAGS='-std=c++11'` to the cmake command. (i.e., the `set(CMAKE_CXX_STANDARD 11)` stanza is not interpreted until 3.1 ).
It would be nice to have a more "relaxed" way of telling people 

> either use >=3.1 OR add  `-std=c++11` to the `CMAKE_CXX_FLAGS`"

than just the `set(CMAKE_CXX_STANDARD 11)` stanza in the cmakefile...

slayoo:
Thanks! 
Wouldn't it be "saner" simply to require CMake 3.1 as per the docs of CMAKE_CXX_STANDARD? 
The advantage is that if someone tries with Visual Studio e.g., CMake will not use -std flag but something else that VS understands (yes, I know it will not help much in that case :)).
CMake 3.1 it has been released around the same time as gcc 4.9.2 (late 2014) so should be available...

GillesDuvert:
No problem. 
INSTALL.CMake still refers to CMake 2.6
Probably the place to hint about the  -DCMAKE_CXX_FLAGS trick, also.

slayoo:
Merci et voilà (commit message was meant to include [skip ci], not [skip-ci] which would disable Travis and Appveyor checks for README-only modification)

INSTALL.CMake
```diff
@@ -80,5 +80,10 @@ cmake -DCMAKE_BUILD_TYPE=Release -DPYTHON=YES -DCMAKE_CXX_FLAGS_RELEASE=" -Wno-r
 make -j 4
 sudo make install
 
-For a special compilation with all the options for fast code, including targeting the machine's processor type, use:
--DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS_RELEASE="-O3 -fno-math-errno -funsafe-math-optimizations -fno-rounding-math -fno-signaling-nans -fcx-limited-range -march=native -DNDEBUG"
+For a special compilation with all the options for fast code, including targeting the machine's processor type, use for example:
+-DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS_RELEASE="-O3 -fno-math-errno -fno-signaling-nans -msse2 -march=native -DNDEBUG"
```
slayoo, line 84: btw, isnt't it equiv. to -Ofast?

GillesDuvert: I do not think: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-Ofast
These are just examples that worked for me.

GillesDuvert:
reopen to restart travis etc after file_et commits

slayoo:
Apparently closing and reopening confused codecov

slayoo:
(no, apparently codecov is just waiting to update its status till all other CI processes finish)
green.

PR #370, opened 2018-06-17
keep obsolete/udg as a tar-gz file, remove directory

maynardGK:
The presence of udg in the GDL creates a messy !PATH variable for those who
have set "GDL_PATH=+$GDLGIT" - small .pro files are sctattered in the udg directories,
 bringing them into the path.  This commit deletes that directory
and memorializes it in a .tgz file:
$ tar tf obsolete/udg.tgz | head -20
udg/
udg/chapters/
udg/chapters/about.tex
udg/chapters/api.tex
udg/chapters/cmdline.tex
udg/chapters/coding.tex
 etc


slayoo:
thanks!

PR #371, opened 2018-06-17
HDF4 uses rpc, so need to link with it for tests

opoplawski:
The sd_nccreate test fails without this.


slayoo:
Is there any overlap with the recent libtirpc changes?
Could there be a conflict if hdf uses glibc rpc and we use tirpc with the -DRPCDIR=...?

opoplawski:
The system is only going to have one rpc library so I don't see any conflicts there.

slayoo:
thanks!

PR #376, opened 2018-06-19
suppress clang unsupported friend class declaration warning

### LINKED ISSUE #340: clang warning (datatypes.hpp:250:31: warning: dependent nested name specifier ...)
```
/Users/travis/build/slayoo/gdl/src/datatypes.hpp:250:31: warning: dependent nested name specifier 'Data_<Sp2>::' for friend class declaration is not supported; turning off access control for 'Data_' [-Wunsupported-friend]
  friend BaseGDL* Data_<Sp2>::Convert2( DType destTy, 
                  ~~~~~~~~~~~~^
```

slayoo:
(and remove the TEMPLATE_FRIEND_OK_ ghost)
Closes #340

### DIFF
--- a/src/datatypes.hpp
+++ b/src/datatypes.hpp
@@ -244,12 +244,18 @@ static	void operator delete( void *ptr);
 
   // not all compilers can handle template friend member functions
 #ifndef _MSC_VER
-#if defined( TEMPLATE_FRIEND_OK_) || (__GNUC__ >= 4)
+#  if __GNUC__ >= 4 
+#    if defined(__clang__)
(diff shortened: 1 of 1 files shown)


slayoo:
In my understanding, the clang warning says that it does compile the code, yet it does not assure private/public access controls - which seems fine to ignore in this case.

slayoo:
@GillesDuvert does it seem a valid "solution" to you?

GillesDuvert:
I a nutshell, let's do it (it completely pollutes logs for nothing) and let C++0x experts do the right thing at some point.

slayoo:
@GillesDuvert in my understanding there is nothing to be done here on our side, it's a missing feature in clang, right? The point is that clang does compile it correctly but it does not provide access control so you can call a private method from outside - that's it

PR #378, opened 2018-06-19
change to finite complex

sflinois:
TEST_FINITE_BASIC, type=6,/tes failed on a Ubuntu 4.4.0-121-generic and on Debian 4.9.88-1+deb9u1
The failed tests seems to be related to signbit()==1
changing it to signbit()!=0 made the tests pass 
(signbit is supposed to return a boolean and both implementation should be the same in theory, but it is not the case)
I don't have more information about the issue


alaingdl:
same issue on Ubuntu 16.04 with gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.9) 

one stupid question ! maybe this code (loops) will gain to be rewrite ?!

alaingdl:
on the same machine, for the Real and Double types, _large_ (1e8) loops are as fast or even faster in GDL than in IDL, but are ~2 time slower for Complex in GDL ...

slayoo:
Should we then re-enable test_finite on travis:
https://github.com/gnudatalanguage/gdl/blob/master/.travis.yml#L43
?

sflinois:
We can try re-enabling it, can we re-open this pull request or should i make another one ?

slayoo:
Probably easier with a new one, thanks!

PR #390, opened 2018-06-21
Update FindGrib.cmake

jdkloe:
this change allows to use eccodes as replacement for grib_api (which will no longer be maintained by ECMWF).
In case you wish to keep the option for the users to select which grib implementation to use (to support older linux distributions that don't have eccodes yet) you could consider adding a cmake switch to allow this.
Unfortunately I am not very familiar with this cmake build system, so it's better if I don't attempt this myself I think.


slayoo:
Thank you!

slayoo:
Big thanks!

jdkloe:
and a big thanks to you as well! Amazing amount of work for a one line patch....

PR #397, opened 2018-06-22
This revision of randomu adds the management of the random pool. 

GillesDuvert:
Random sequences *AND* random pool are identical to idl at this time. One can in particular save the random pool state and restore it at will. this permits to manange parallell random series.

### DIFF
--- a/src/gsl_fun.cpp
+++ b/src/gsl_fun.cpp
@@ -31,6 +31,9 @@
 #include "gsl_fun.hpp"
 #include "dinterpreter.hpp"
 
+#ifdef _MSC_VER
+#include "gtdhelper.hpp" //for gettimeofday()
+#endif
 // ms: must not be inlcuded here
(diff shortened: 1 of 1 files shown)


slayoo:
from Appveyor log:
```
C:\projects\gdl\src\gsl_fun.cpp: In function 'BaseGDL* lib::random_fun(EnvT*)':
C:\projects\gdl\src\gsl_fun.cpp:1067:32: error: 'gettimeofday' was not declared in this scope
       gettimeofday(&tval,&tzone);
                                ^
src\CMakeFiles\gdl.dir\build.make:930: recipe for target 'src/CMakeFiles/gdl.dir/gsl_fun.cpp.obj' failed
```

GillesDuvert:
apparently gettimeofday() exists now in mingw. So we can get rid of gtdhelper.cpp, hpp and references in the code.

GillesDuvert:
:+1:

PR #409, opened 2018-06-29
WIP: relaxing CMake version requirement (buggy)

slayoo:


alaingdl:
how can I download this in my clone of the project ?! (I did not find how to download the file too !)

slayoo:
you can pull from this PR branch:
```
$ git pull git@github.com:gnudatalanguage/gdl.git slayoo-patch-1
```
HTH

slayoo:
and as for the download, here is the URL:
https://raw.githubusercontent.com/gnudatalanguage/gdl/slayoo-patch-1/CMakeLists.txt

alaingdl:
@slayoo  thanks for the link to direct download, some tricks in GH interfaceare not in my mind now

Since now the flag c++11 is mandatory, why not put it directly in the CMakeList.txt ?
here I saw suggestions
 https://stackoverflow.com/questions/11783932/how-to-add-linker-or-compile-flag-in-cmake-file

and I propose : 

cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
# requirement for CMAKE_CXX_STANDARD (3.2 for Apple, 3.1 for non-Apple: https://public.kitware.com/Bug/bug_relationship_graph.php?bug_id=15355)
if (${CMAKE_VERSION} VERSION_LESS "3.2")
  string(FIND "${CMAKE_CXX_FLAGS}" "-std=c++11" tmp)
  if(tmp LESS 0)
    SET( CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS}" "-std=c++11")
  endif()
else()
  set(CMAKE_CXX_STANDARD 11)
endif()


Do the compiler will try it later ???

(I have no time to test now, sorry)

slayoo:
There are two issues:
1. -std=c++11 works with clang, gcc and icc, but not with MSVC... which isn't a real issue, I know;
2. -std=c++11 is not the only way of enabling c++11 constructs, gnu++11, c++0x, c++17 etc will all do

So my idea was to simply encourage users to upgrade CMake instead of us playing with detecting all the possible cases... in fact I would suggest to use
```
string(FIND "${CMAKE_CXX_FLAGS}" "-std=" tmp) 
```
only and simply ask the user to take responsibility for picking the correct standard (which is still not perfect as current compiler versions have c++11 by default enabled which makes all this warning invalid...)

alaingdl:
OK ! very pertinent remark ! +1 !

slayoo:
Alain, I was just testing this locally, and realised that we would also need to set 
CMAKE_CXX_FLAGS_DEBUG, CMAKE_CXX_FLAGS_RELEASE, CMAKE_CXX_FLAGS_RELWITHDEBINFO and CMAKE_CXX_FLAGS_MINSIZEREL to make it work (due to the logic around line 1085 in CMakeLists.txt). This becomes messy :(

alaingdl:
no time to extend the message (clarify it), but I need the code !

slayoo:
Alain, it's not only about the message - the behaviour now is clearly buggy - it does not work when specifying the build type (Debug or Release) + it does not cover the case when someone specifies -std=c++17 etc, and _most importantly_ when someone has a recent compiler version with C++11 turned on by default. I suggest reverting.

PR #415, opened 2018-07-02
case no display in get_screen_size

sflinois:
When no display is found, get_screen_size return a [0, 0] array instead of throwing an exception

### DIFF
--- a/src/devicewin.cpp
+++ b/src/devicewin.cpp
@@ -447,6 +447,7 @@ DIntGDL* 	DeviceWIN::GetScreenSize(char *disp)
 	MaxXYSize(&xsize, &ysize);
 	DIntGDL* res;
 	res = new DIntGDL(2, BaseGDL::NOZERO);
+    std::cout << ">>HERE<<" << std::endl
 	(*res)[0] = xsize;
 	(*res)[1] = ysize;
 	return res;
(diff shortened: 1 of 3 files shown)


alaingdl:
lest try it in real life !

slayoo:
satisfying codecov here would be quite easy here!

alaingdl:
it is already tested in two places : "test_get_screen_size" and, indirectly but efficiently (how we located the bug) via "test_resolve_routine"

alaingdl:
maybe something is still wrong in "test_resolve_routine", around line 23

if (files EQ '') then files=FILE_SEARCH(path+path_sep()+filter)

should be 

if (files[0] EQ '') then files=FILE_SEARCH(path+path_sep()+filter)

slayoo:
test_resolve_routine is still not included in the travis run (hence not included in codecov report):
https://github.com/gnudatalanguage/gdl/blob/master/.travis.yml#L51

slayoo:
test_get_screen_size.pro is also not captured by codecov as on Travis the test is skipped through "EXIT, status=77" due to the DISPLAY env var not being set:
https://github.com/gnudatalanguage/gdl/blob/master/testsuite/test_get_screen_size.pro#L96

PR #423, opened 2018-07-10
test_resolve_routine should work with files in the current dir. too !

alaingdl:
my initial change was OK when the current dir. does no contain files (then we change the dir)
but it should also work when the current dir. does contain files !

NB: this code can be easily used to compile large set of *pro files (eg : idlastro, coyote & mpfit
packages !) calling : 

GDL> .r test_resolve_routine.pro
GDL> TEST_ALL_TEST_ROUTINES, filter='*pro'


alaingdl:
block in App for other reasons ...

PR #432, opened 2018-07-26
resolving problems with help,/path_cache

SIAMAR:
we changed ifndef _WIN32 with ifndef _MSC_VER so that Windows is able to find files in directories path

this changes is made for the Mingw64 build on windows


SIAMAR:
i forgot the pathsearch 
the BeautifyPath is not usefull since Windows is already case-incensitive 
BeautifyPath is changing the path and pathsearch is not able to find files in the path

Issue #427, opened 2018-07-20
Windows build problems and discussion


sflinois:
Opening a topic about GDL build on Windows

The focus of this discussion is to solve problems to automate GDL build on MSYS.
And, on a longer term, to be able to build GDL on native Windows (With Visual Studio or Clang)

you can find more info on windows build here : 
https://github.com/gnudatalanguage/gdl/issues/242
https://github.com/gnudatalanguage/gdl/pull/416

maynardGK:
I have been working with the Appveyor build effected via the  appveyor.yml file in the git distribution.
#408 
 scripts/mingw-bldgdl.bat and scripts/installgdl.ps1 are copied from the appveyor build.
  To replicate what appveyor does at home, make a C:\projects folder and make a symlink called gdl pointing to the git. This can be done to a separate disk (with lots of space).
Also, make a symlink at top-level C: to mingw-w64 so that C:\mingw-w64\ points to where the mingw-w64 distributions are kept (again, a different disk than C: which may have space constraints). With these file system changes you can cut-and-paste to replicate the build done by appveyor.  Another link is to match the name of the compiler distro in the appveyor build with the one you have at home.  So my d:/mingw-w64/ directory looks like so:

07/01/2018  09:47 PM    \<DIR\>          i686-5.4.0-posix-dwarf-rt_v5-rev0
07/01/2018  09:38 PM        42,692,388 i686-5.4.0-release-posix-dwarf-rt_v5-rev0.7z
07/21/2018  12:07 AM    \<SYMLINKD\>     i686-6.3.0-posix-dwarf-rt_v5-rev1 [i686-6.4.0-posix-dwarf-rt_v5-rev0]
07/02/2018  11:54 AM    \<DIR\>          i686-6.4.0-posix-dwarf-rt_v5-rev0
06/30/2018  05:55 PM        42,648,051 i686-6.4.0-release-posix-dwarf-rt_v5-rev0.7z
07/09/2018  01:30 PM    \<DIR\>          i686-7.3.0-posix-dwarf-rt_v5-rev0
07/07/2018  04:09 PM        44,677,667 i686-7.3.0-release-posix-dwarf-rt_v5-rev0.7z

You can make symlinks in windows from an MSYS shell run "as an administrator" and where you export the environment MSYS=winsymlinks:nativestrict 

Update:
https://ci.appveyor.com/project/slayoo/gdl/build/1.0.564

[...]

PR #440, opened 2018-08-02
adapt home path for history on windows + added script for windows gdl build

SIAMAR:
We made a script to build GDL on Windows 
you can test it and give us your opinion.


slayoo:
Guys, @SIAMAR @sflinois @alaingdl @acoulais @maynardGK,

We need to somehow take care of the users and decide what to suggest them - to follow @maynardGK's Appveyor scripts (or download the pre-built artefact) or to use the script in this PR. Could this be somehow unified?

alaingdl:
unfortunately, this PR mixed 2 aspects:

* bug correction for history management for MSwin (should be add now in master)

* a script "similar" to [minimum_script4gdl.sh]( http://aramis.obspm.fr/~coulais/IDL_et_GDL/minimum_script4gdl.sh)  for MSwin, to be discuss.
(we already delivered a gdl.zip last week)

This script is very useful for me because I don't know how to use what did @maynardGK 
but I can compile change in master with this one (and play with testsuite on MSwin
with that code)

If I understood well, my students did not succeed to reproduce few steps @maynardGK 
did, but succeed to go to the end with this script, for example on "void" MSwin VM.

maynardGK:
- in gdl.bat, you have probably missed making PLPLOT_LIB assignment so that plotting can work.

- Also, since a custom plplot is being built, options can be set to not build dynamic drivers so that lib/plplot-5.13.0/drivers is not necessary. -DENABLE_DYNDRIVERS=OFF othewise you will also need a PLPLOT_DRV_DIR assignment.  

- @set GDL_PATH=+%cd%\share\gnudatalanguage\lib will cover anything under gdllib; it does makes for a long !PATH which is why I make a copy named gdllib

- -DCMAKE_SYSTEM_PREFIX_PATH=<...>/LIBS  so that cmake will find items via <>/LIBS/lib & etc.
then the relevant items in /LIBS/bin can be found by this being in the %PATH%.

sflinois:
The appveyor build good if you want to try out and use a specific version of GDL but it doesn't allow us to change the code and recompile, which make it hard to find bugs and modify the code.
(Another problem I have with the appveyor build is with readline 7.0. The library block the command prompt on 2 Windows 10 machines I tested it on, that is why we used readline5.0.1 in the script)

The script is new and need some (a lot of) improvements, but I think that the script and the appveyor build are complementary. =)

pjb7687:
> (Another problem I have with the appveyor build is with readline 7.0. The library block the command prompt on 2 Windows 10 machines I tested it on, that is why we used readline5.0.1 in the script)

This is exactly the problem that I had with readline years ago. That was one reason I had to avoid MSYS2 build.. (if I remember correctly the other reason was related to WxWidgets) May I ask whether you already tested WinEditline?

sflinois:
We tested WinEditline but without great success (errors about undefine references during the GDL build)
We manage to have what we needed with readline5.0.1 

Edit : We still have a bug with editline, we can't use any character that need AltGr :(

pjb7687:
Okay, then we still need the old Readline lib... Then this script seems to be the best way to build error-free GDL on Windows. I guess this script could be integrated to appveyor build script..? I actually don't understand what you mean by:

> The appveyor build good if you want to try out and use a specific version of GDL but it doesn't allow us to change the code and recompile, which make it hard to find bugs and modify the code.

IMHO, appveyor doesn't restrict us to do anything with the source code...?

maynardGK:
Followup to this on issue #427. 
 
I've always had problems running GDL from an MSYS console window; instead now I directly start
a CMD window where I start /usr/bin/bash.exe -login -i,  from \usr\bin, and reset my path before I begin:

C:/msys64/home/greg/.bashrc invoked with OLDPWD=/msysC/usr/bin

greg@Homerw7  ~
$ pwd
/home/greg

greg@Homerw7  ~
$ echo $PATH
/usr/local/bin:/usr/bin:/bin:/c/Windows:/c/Windows/system32:/c/Windows/system32/Wbem:/usr/bin/site
erl:/usr/bin/vendor_perl:/usr/bin/core_perl

greg@Homerw7  ~
$ . run32
Wed, Jul 25, 2018 10:16:49 PM /d/bld/gdl/mingw32-git/src/gdl <gdlgit>
Sun, Apr 15, 2018 6:10:52 AM /d/bld/gdl/branches/master <master>
Tue, Apr 10, 2018 7:19:16 PM /d/bld/gdl/mingw32-svn/src/gdl <gdlsvn>
<etc>

greg@Homerw7  ~
$ echo $PATH
/c/usr/bin32:/c/usr/bin32/mingw:/c/usr/lib32:/mingw32/bin:/opt32/bin:/usr/local/bin:/usr/bin:/bin:
/Windows/system32:/c/Windows/system32/Wbem

greg@Homerw7  ~
$

I don't update my support libraries much - I am at gcc 5.3 - but I did have a problem with readline 7
which went back to normal using readline 6.3; also, when running the appveyor GDL from a CMD shell
directly it also has trouble, I believe.  Starting it from explorer or from an MSYS shell (as described above) the new readline works OK.

We should get the new appveyor scripts into master so that we are all talking about the same thing.
As a bonus for that, the fully fuctional HASH and LIST functions are in the same PR.
I put the original script in <git>/obsolete/old_appveyor.yml and broke out the wxwidgets build into scripts/wxmake_mingw.bat.  Its much easier in appveyor to go to MSYS mode for more familiar shell commands, but the Appveyor MSYS is not complete enough to compile a new GraphicsMagick.
As I described elsewhere, once a few symbolic links are made then your git distribution can appear 
to be C:\projects\gdl  (also /c/projects/gdl) and the /install, /build, /mingw, and /win32libs directories can be created for local usage (they dont't get deleted on a new checkout); the scripts that appveyor uses can be directly invoked.

For wxWidgets, plplot/wxwidgets was not being built in old_appveyor and the static libraries were being referenced in the GDL build, not the dynamic ones.  I made a patch for plplot's cmake to find wxWidgets in the MSYS case but it stopped working (it will still find wxWidgets ok if they are installed under /mingw32 - I wanted to keep the wx installation separate).  I tried to get the mingw wx to hook into plplot but that failed.  With MSYS2 I found that it is necessary to edit the bin/wx-config file so that
certain options are not present in the link: hence the sed command as found in appveyor_gdl.msys.
After that there are plenty of other problems as discussed in #331

slayoo:
I wholeheartedly concur with @pjb7687: would be great to have one script that would be used for Appveyor and manually by users - otherwise we will have much more work in supporting windows users (and in maintaining two sets of scripts). 

@alaingdl @acoulais please note that you can log in remotely to the Appveyor machines after the build and experiment there - it works (just uncomment this line: https://github.com/gnudatalanguage/gdl/blob/master/appveyor.yml#L267)

alaingdl:
same problem than for PR #442 

blocked by trouble in Travis for OSX ...

what can we do ? the unwanted file was removed, and we need to fix the "bug" described & solved here

and the discussion here is rich for the MSwin build

slayoo:
reopened to trigger a rebuild after travis/homebrew/numpy problem was fixed in master (#444)

alaingdl:
lets be clear : this PR mixed 2 different aspects : a poor management (bug) of HOME/HOMEPATH 
for MSwin (in basic_pro.cpp)  and a script for compilation on MSwin (to be put in scripts/)

now this PR contains a lot of interesting informations for MSwin compilations,
but we need to commit the patch for basic_pro.cpp !

Issue #441, opened 2018-08-06
Bug in POINT_LUN with negative LUN, possible fix included


pstrub:
POINT_LUN, lun, position

 is used to set or get the current position within an open file. When a negative number is used as a lun, the procedure returns the current position inside the file, and should use a 64 bit variable when necessary (the IDL documentation on this reads: "The returned type will be a longword signed integer if the position is small enough to fit, and a 64-bit longword integer otherwise.") 

GDL's version of POINT_LUN however always returns a 32 bit long as a position. This makes it impossible to correctly obtain the current position beyond 2GiByte file size, the returned position gets truncated to 32 bits. 

I found a fix that resolves the issue for me, and to my knowledge should not have any side-effects. As I'm not familiar with the way the development of GDL is organised, nor with the prerequisites of contributing/submitting changes to the repository, I post my suggestion here.

In basic_pro_jmg.cpp, version 5a970cd  on Apr 16, lines 66 and 67 read:

``` cpp
    *retPos = new DLongGDL( actUnit.Tell());
    return;
```
These lines invariably return the position inside the file as a 32-bit integer, truncationg the 64 bit integer returned by actUnit.Tell().
I therefore replaced it with the following code:

``` cpp
    DLong64 pos = actUnit.Tell();
    if (pos > std::numeric_limits<DLong>::max()) {
      *retPos = new DLong64GDL( pos);
    } else {
      *retPos = new DLongGDL( pos);
    }
    return;
```

for numeric_limits to work, make sure to add `#include <limits>` to the top of the file.

Edit: Setting the file position beyond 2 GiByte works fine with the current version of POINT_LUN.

Please let me know if you need any further information.

alaingdl:
Thanks for the bug report and the suggested patch ! (see PR #442)

You are very welcome to participate. I am not sure myself to fully understand the whole frame work
but first may be you have to go here https://github.com/orgs/gnudatalanguage/teams/gdlteam/members
After being registered as a member, you can make pull requests.

best regards
A.

slayoo:
Anyone can make a pull request, but please do join the team so you can get instant notifications on new GDL-related discussions.

pstrub:
Thanks for the invitation! Not sure though how much I'm going to contribute though, but I'll do what I can do! 

On another note, shouldn't we reopen the issue, as the pull request has been cancelled due to a regression?

PR #442, opened 2018-08-06
correction POINT_LUN following bug report #441

alaingdl:
accepted bug report #441 from @pstrub

I tested that on 2 Linux ...

### DIFF
--- a/src/basic_pro_jmg.cpp
+++ b/src/basic_pro_jmg.cpp
@@ -20,6 +20,7 @@
 
 #include <string>
 #include <fstream>
+#include <limits>
 
 #include "envt.hpp"
 #include "dinterpreter.hpp"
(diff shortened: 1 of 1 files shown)


alaingdl:
Travis for OSX seems to be broken :(

pstrub:
Hi @alaingdl! What's the issue/error msg with Travis OSX? I didn't have any problems building it on my OSX  10.13.6 Macbook. 

Anyway, if the pull request is closed because patch leads to regressions, shouldn't we reopen this in the issue tracker?

slayoo:
Let's keep it open! 
The issue is around homebrew numpy installation - it's not a regression, just a coincidence with some recent change in homebrew.

pstrub:
Thanks for the clarification and for reopening, @slayoo ! I'll keep following the thread!

alaingdl:
I am lost ! the code is OK and the problem with Travis is unrelated !

I would like to merge it :)

may be we can open a bug report for the Travis issue ? I have no time now to look at Travis, some codes pending

slayoo:
@pstrub @alaingdl it would also be great to add a test for this one!

slayoo:
I'll look this evening into the homebrew/numpy issue

alaingdl:
1/ we already have a "test_point_lun"

2/ unclear for me if we need to have a 2+ Go file or not
because we can play with the index above the limit
(unclear for me why)

3/ much more concern by the other place(s) we may have not
taken into account the 64b limit ... I made months ago a review
on some /L64 options (and I have computers to test arrays with such size !)

slayoo:
travis/homebrew/numpy problem fixed (#444)

PR #443, opened 2018-08-07
update of minimum_script4gdl.sh for 0.9.8 & the current Git Version

alaingdl:
update of minimum_script4gdl.sh to be able to compile 0.9.8 vanilla + current Git master

majors changes : 
* git version (runing master)
* move to cmake 3+ (if missing, compiles 3.2.3)
* move to plplot 5.12 or 5.13 (pb with 5.12 on OSX ...)
* don't need to have "git" or "svn" around

successfully tested for 098 vanilla + Git on OSX 10.11, 10.12 & 10.13, Ubuntu 14.04 & 16.04, various Debian

not OK for 098 vanilla on Fedora28 (because of tirpc), OK for Git

### DIFF
--- a/scripts/minimum_script4gdl.sh
+++ b/scripts/minimum_script4gdl.sh
@@ -2,6 +2,8 @@
 #
 # Alain Coulais, 3 Mars 2015, script under GNU GPL v3
 #
+# option "5" --> Vanilla; option "6" --> Git Master
+#
 # Changes:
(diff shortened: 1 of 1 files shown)


alaingdl:
we don't need CI & Travis for that !

slayoo:
well, good point! we actually could include it in the travis run!

PR #452, opened 2018-08-28
WIP: adding IMPLIED_PRINT

sflinois:
I'm currently trying to implement the IMPLIED_PRINT (see http://www.harrisgeospatial.com/docs/ImpliedPrint.html)

I don't really know what is the "cleanest" way to implement it.
I implemented the IMPLIED_PRINT behavior for Float and Double in this PR :
```
GDL> print, !dpi
       3.1415927
GDL> !dpi
       3.1415926535897931
GDL> print, 3.1
      3.10000
GDL> 3.1
       3.0999999

IDL> print, !dpi
       3.1415927
IDL> !dpi
       3.1415926535897931
IDL> print, 3.1
      3.10000
IDL> 3.1
       3.0999999
```

Some variables, such as the Structures are more complex to implement and I had some questions about the way I should implement the IMPLIED_PRINT option.

Note that the print function use  ToStream() to format and print a variable, and the ToStream() parameters do not have the EnvT* e variable in it.

1) Adding a ToStreamImplied function for every type :
    It adds a lot of lines in the default_io file, and in some cases (Float and Double for exemple) it adds a lot of duplicate code. (This is what I did in this PR)
2) Adding the Envt *e variable in ToStream prototype :
    It adds an arg to ToStream but we will have less duplicate code in some functions. It might also complexifie some cases (Such as structure) with some special cases added.
3) Adding binary flags, which kinda act like adding the Envt *e, but it also add some const var, and it's never good adding too many

Feel free to askquestions and to discuss the implementention ideas :)

### DIFF
--- a/src/basegdl.cpp
+++ b/src/basegdl.cpp
@@ -136,6 +136,11 @@ std::ostream& BaseGDL::ToStream(std::ostream& o, SizeT width,
 {
   throw GDLException("BaseGDL::ToStream(...) called.");
 }
+std::ostream& BaseGDL::ToStreamImplied(std::ostream& o, SizeT width, 
+				SizeT* actPosPtr )
+{
+  throw GDLException("BaseGDL::ToStreamImplied(...) called.");
+}
 std::istream& BaseGDL::FromStream(std::istream& i)
 {
   i >> *this;
--- a/src/basegdl.hpp
+++ b/src/basegdl.hpp
@@ -508,6 +508,8 @@ public:
 
   virtual std::ostream& ToStream(std::ostream& o, SizeT width = 0, 
 			    SizeT* actPosPtr = NULL);
+  virtual std::ostream& ToStreamImplied(std::ostream& o, SizeT width = 0, 
+			    SizeT* actPosPtr = NULL);
   virtual std::istream& FromStream(std::istream& i);
 
   virtual bool Greater(SizeT i1, SizeT i2) const; // comp 2 elements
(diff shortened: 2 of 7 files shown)


GillesDuvert:
@sflinois implied print is already here, see #254 .
No need to implement it.
Admitedly, there may be format differences.

sflinois:
I figured out that implied print was already implemented, the idea was to implement the IDL format of implied print.
We can't implement it with the current code structure, i believe we need to change the toStream function prototype or we need to implement a new function

GillesDuvert:
implied_print is just trying "print,/implied_print,toto" when "toto" alone raises an error. as written in libinit.cpp: `//At the moment, print accepts silently "IMPLIED_PRINT" as this is used in autoprint feature.`

So, this is just, at the moment, a print.

`/implied_print` is in fact the same JSON formatting that one has in JSON_SERIALIZE (at the moment, a procedure in src/pro). Actually, IDL says "If you want to output prettified JSON (with appropriate indenting and line breaks), you can simply bypass JSON_SERIALIZE and print out your LIST or HASH using Implied Print."

So, one shoud IMHO:
- pass JSON_SERIALIZE in C++.
- this is to be done exactly the same way as SAVE is done (much simpler of course, but can be done almost by removing superfluous lines of lib::save) 
- make it so that it can be called directly from lib::print when option /implied_print is present.
- no need to change toStream (good, as this is rather dangerous) 

To be complete, writing JSON_PARSE (which is also a simpler replicate of lib::restore) along the way seems mandatory.

GillesDuvert:
closing and reopening to check compiles

GillesDuvert:
reopening to check travis and appveyor.

GillesDuvert:
hmm.. looks bad. lets open again with some other PRs accepted.

PR #375, opened 2018-06-18
libtiff support

rexso:
Rebased and cleaned-up version of #288.

PR #454, opened 2018-09-05
libtiff implementation improvements

rexso:
Dereferencing the geotiff structure returned from QUERY_TIFF/READ_TIFF used on non-geotagged TIFF images resulted in a segfault because of empty values. This patch fixes this problem by returning a scalar zero instead of an empty structure, similar to IDL:

>  If no GeoTIFF information is present in the file, the returned variable is set to a scalar zero.

Also, the colour and orientation problem mentioned in https://github.com/gnudatalanguage/gdl/pull/375#issuecomment-398881741 should now be fixed.

### DIFF
--- a/src/dstructfactory.cxx
+++ b/src/dstructfactory.cxx
@@ -30,6 +30,9 @@ DStructFactory::~DStructFactory()
 
 DStructGDL* DStructFactory::Create()
 {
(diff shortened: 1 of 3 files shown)


slayoo:
thank you!

PR #458, opened 2018-09-07
Feature-complete DILATE and ERODE functions

rexso:
The current C++ implementation of DILATE and ERODE only supports binary images, while we (the company I am working for) have written more feature complete versions in IDL supporting greyscale as well. Even though a potentially more performant C++ implementation is preferable, I still think supporting all features stated in the IDL documentation is more important.

A more performant C++ version can be enabled when completed. Unless someone else is working on this right now, we have talked about porting the code of this PR to C++ at a later point, but we are focusing on some other things right now.

We have also written some basic tests based on the IDL documentation for both [ERODE](https://www.harrisgeospatial.com/docs/ERODE.html) and [DILATE](https://www.harrisgeospatial.com/docs/DILATE.html).

### DIFF
--- a/src/libinit_ac.cpp
+++ b/src/libinit_ac.cpp
@@ -85,12 +85,20 @@ void LibInit_ac()
   new DLibFunRetNew(lib::sobel_fun,string("SOBEL"),1,helpKey);
   new DLibFunRetNew(lib::prewitt_fun,string("PREWITT"),1,helpKey);
 
+  /* NOTE: 2018-09-07 by rexso (remi.solaas at edinsights dot no)
+   *
+   * ERODE and DILATE currently implemented in IDL,
+   * available in pro/erode.pro and pro/dilate.pro
+   * with grayscale support. Uncomment when the C++
+   * implementation is complete.
+   *
   const string erodeKey[]={"HELP","GRAY","PRESERVE_TYPE","UINT","ULONG","VALUES",KLISTEND};
   new DLibFunRetNew(lib::erode_fun,string("ERODE"),5,erodeKey);
 
   const string dilateKey[]={"HELP","GRAY","PRESERVE_TYPE","UINT","ULONG","VALUES",
 			    "CONSTRAINED","BACKGROUND",KLISTEND};
   new DLibFunRetNew(lib::dilate_fun,string("DILATE"),5,dilateKey);
+  */
 
   const string matrix_multiplyKey[]={"ATRANSPOSE","BTRANSPOSE",KLISTEND};
   new DLibFunRetNew(lib::matrix_multiply,string("MATRIX_MULTIPLY"),2,matrix_multiplyKey);
--- a/src/pro/dilate.pro
+++ b/src/pro/dilate.pro
@@ -1,0 +1,467 @@
+;
+;  GDL replacement to DILATE in IDL
+;  Bin Wu <bin.wu (at) edinsights.no>
+;  Aug. 2018
+;
+;  Arguments (from IDL): Image Structure X0 Y0 Z0
+;  Keywords (from IDL): 
+;  	BACKGROUND 
+;		Set this keyword to the pixel value that is to be considered the background when 
+;		dilation is being performed in constrained mode. The default value is 0.
+; 	CONSTRAINED (a boolean keyword)
+;		If this keyword is set and grayscale dilation has been selected, the dilation 
+;		algorithm will operate in constrained mode. In this mode, a pixel is set to the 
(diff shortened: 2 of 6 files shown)


rexso:
**test_call_external.pro** fails on MacOS, not related to this PR I suppose?

slayoo:
Clearly unrelated... but worth fixing :)

At first sight there seem to be something fishy with libfile extension (.so vs. .dylib):
```
Scanning dependencies of target test_ce
[ 99%]Building CXX object testsuite/CMakeFiles/test_ce.dir/libtest_ce.cpp.o
[100%]Linking CXX shared library libtest_ce.dylib
[100%] Built target test_ce
```
...
```
piled module: TEST_CALL_EXTERNAL.
% LOOKING_FOR_CE_LIB: library >>libtest_ce.so<< not found or not compiled !
<end of output>
```

slayoo:
This seems to have stopped working:
https://github.com/gnudatalanguage/gdl/commit/9fd3f4abaa96eff8ab48629305a4b595a2efc8e8#diff-f343f1d9fa89e2bb9cbcb468270a7917R22

slayoo:
Travis has changed the default OSX image last month:
https://changelog.travis-ci.com/default-macos-image-is-now-xcode-9-4-67738

slayoo:
@acoulais @alaingdl this commit has broken it:
https://github.com/gnudatalanguage/gdl/commit/77587c27104f40d4a9b8998966e07421b43456a9

slayoo:
in src/initsysvar.cpp:
```c++
    // AC 2018-sep-07
    if (strcmp(SysName,"darwin") ==0) SysName="Mac OS X";
```
but in many other places we have:
```idl
if (!version.os_name EQ 'darwin') then begin
```
Here's a list:
- testsuite/test_file_which.pro
- testsuite/benchmark/benchmark_generate_cpuinfo.pro
- testsuite/test_python_module_0.pro
- testsuite/test_python_module_1.pro
- testsuite/test_wait.pro
- src/pro/make_dll.pro
- src/pro/zenity_check.pro
- src/pro/online_help.pro
- testsuite/test_read_standard_images.pro
- testsuite/test_call_external.pro

slayoo:
This was introduced in this PR which has 100% CI failures on OSX: https://github.com/gnudatalanguage/gdl/pull/457 :(

rexso:
Thanks for looking into the test issue @slayoo

slayoo:
I'll merge this one and open a new issue for fixing the darwin issue

alaingdl:
Sorry for that, I made it in an hurry, expecting we will forgot to correct it if I did not do it at that time
(I never access to IDL on OSX before !)

IDL> !version
{
    "ARCH": "x86_64",
    "OS": "darwin",
    "OS_FAMILY": "unix",
    "OS_NAME": "Mac OS X",
    "RELEASE": "8.5",
    "BUILD_DATE": "Jul  7 2015",
    "MEMORY_BITS": 64,
    "FILE_OFFSET_BITS": 64
}

alaingdl:
You are too fast for me ! I would like to suggest we rename the C++ codes into erode2 & dilate2 (or any other convenient name) to allow easy comparison (values & bench)
then if the c++ versions are really faster, we call also call them from the *.pro

Amar already tested them and said to me it was OK

slayoo:
added a ticket for it: https://github.com/gnudatalanguage/gdl/issues/461

alaingdl:
I gave today a look into "test_erode" and "test_dilate"
For me, it it a problem to be that close to IDL examples (copyrighted material)

Up to now, we try to used test cases not based on IDL doc.
(in the past we used few examples posted in idl-pvwave news group !)

(but for sure matrix inversion remained inversion, and such E or D operations also :)

A.

slayoo:
@alaingdl @rexso please suggest how to proceed - fix the copyright issue in a new PR or first reverse & then fix it?

alaingdl:
Amar, who defended with success his M1 degree yesterday, should visit me next week to fix several points including inside Erore/Dilate. I would say : wait till next week to fix it.

@rexso  And nevertheless thanks to have prepared test cases

rexso:
I must say that I don't share the concern of copyright infringement. Even though both the implementation and the test cases of both DILATE and ERODE are based on the documentation as clearly stated in the code, no copyrighted code is used in neither case. I could find no references to any texts stating that code in the examples are copyrighted, nor is any code from the examples used directly anyway.

Some values used in the IDL examples are indeed used in the tests, but is this really what you are concerned about? Please elaborate because it seems like I am missing something? If you are concerned about using the exact same numbers as used in the examples, we could certainly change the numbers.

PR #470, opened 2018-09-24
Added LINFIT implementation

rexso:
LINFIT implemented in IDL.
Might port to C++ in the future.

### DIFF
--- a/src/pro/linfit.pro
+++ b/src/pro/linfit.pro
@@ -1,0 +1,104 @@
+;
+;  GDL replacement to LINFIT in IDL
+;  Bin Wu <bin.wu (at) edinsights.no>
+;  Sep. 2018
+;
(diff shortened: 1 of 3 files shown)


slayoo:
Thanks!

Quick comments:
- the summary comments in linfit.pro are copy-pasted from IDL docs 
- most of the keywords are not covered with the unit test
- should we propagate the DOUBLE kw down into procedures called from within LINFIT (e.g., TOTAL) - if so, let's try to cover it in the unit test as well

rexso:
@slayoo thanks for your comments.

I removed the summary copied from the IDL docs used during implementation, but kept the link for future reference. Will look into adding some more tests to cover all keywords.

slayoo:
thank you

PR #473, opened 2018-09-26
Create gdl.desktop

alexmyczko:
allow launching from menu system, should be installed in /usr/share/applications, and icon should also be added and put in /usr/share/icons called gdl as svg or png preferably.

### DIFF
--- a/doc/gdl.desktop
+++ b/doc/gdl.desktop
@@ -1,0 +1,9 @@
+[Desktop Entry]
+Name=GNU Data Language
+Comment=Free IDL compatible incremental compiler
+Exec=gdl
+Icon=gdl
+Terminal=true
+Type=Application
+Categories=Science;
+Keywords=gdl;idl;


alaingdl:
Hi

Could you consider to put in the CMakeLists.txt what is requered to 
propagate this in packages (in order they  can add this file in the right place) ?
Tanks !

(It would be great too to add the testsuite/ (+bench) in the package !)

GillesDuvert:
Welcome addition, we now need a good icon and somehow have a script to get this .desktop put  at the good location in the (freedesktop.org?) system. Probably our fellows packagers can tell what is needed?

slayoo:
@gnudatalanguage/gdlpackagers ?

olebole:
In Debian, we put it under `/usr/share/applications/`, but there is a freedesktop.org specification for it (however it seems to be not simple go get the canonical definition). From [this](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-1.1.html#desktop-file-id), I would think that it goes to an `application` subdir of one of the paths in `$XDG_DATA_DIRS`.

GillesDuvert:
This PR is dormant as:
1) writing a .desktop for GDL is easy but having it put at the right (freedesktop compliant) place by the installer seems to be another piece of cake --- certainly better done by one of the GDL packagers (sorry).
2) and we need an icon.

alexmyczko:
Please see https://github.com/gnudatalanguage/gdl/issues/644 for some example icons

PR #476, opened 2018-09-28
found the culprit for significant speed regression

GillesDuvert:
Such basic "inside" operators should be treated with extreme precaution, as conditional clauses are bound to slow the code. Asserts are valid as they are not used in optimized code, but debug lines should always rendered inactive by defining debug=0 at star and then making debug code conditional as in
`if (debug) do_something;`

### DIFF
--- a/src/gdlarray.hpp
+++ b/src/gdlarray.hpp
@@ -206,14 +206,11 @@ public:
 
   T& operator[]( SizeT ix) throw()
   {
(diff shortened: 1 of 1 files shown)


alaingdl:
successfully tested on two machines : a 8 cores (debian, gcc 6) & a 16 cores (OSX, clang + openmp)

GDL and IDL equal for RANDOMU, COS, SIN, TAN & ALOG (+EXP)

SQRT() is in serious trouble : 3 times slower on OSX , 2 times on Linux

PR #477, opened 2018-09-29
brings back sqrt() speed on par with other atomic operations

GillesDuvert:


GillesDuvert:
on my laptop, 4 cores, 8 threads, 4 years old intel I7, gcc 4.9.2. GDL compiled with -O3 -march=native.
top lines are for /double option of basic_benchmarks.

image

PR #478, opened 2018-10-06
WIP: Track failure of windows' build (appveyor) on hashlist branch

maynardGK:
Puzzle! The merge from master in commit c8a4268
resulted in a branch which no longer links in windows, ( in the 64-bit case ) and also has problems in one case of travis tests. Unfortunately there are a lot of modifications that the merge brought in so where to look? This commit begins as a reversion of that PR to the last, best working repo, 4127137.
  The merges following this all seem pretty harmless; the suspected change in CMakeLists.txt was already tested #431 to be non-effective.

PR #431, opened 2018-07-26
Hash/List functions in finished form & Appveyor build beef up (incl. publishing downloadable builds)

maynardGK:
HASH and LIST improvements were to be a separate PR but they were tacked into the Windows' build.
 Mostly all of the properties of HASH and LIST are now implemented with this change, which has radical changes to the list__overloadleftbracket and rightbracket routines. In hash.cpp, the /recursive option of the TOSTRUCT method is not working. "ordered hash" has a property flag but is not implemented. 

Win_build:

   GDL is built in appveyor using mingw-w64 compiler and support libraries built with this
compiler.   The  **scripts/appveyor_1.msys** sequence obtains those, pre-built, needed from the msys repository **http://repo.msys2.org/mingw/i686/**.    Some of these distributions have been found to not be compatible with other elements of the appveyor build. So, "version upgrades" need to be tested methodically.

  **wxwidgets** and **graphicsmagick** are the next two behemoth support libraries to be integrated.

The previous **appveyor.yml** included a build of wxWidgets which uses mingw32-make (not the msys make) and results in library and include files suitable for use by the "Windows" option of CMAKE.  Due to incomplete development of this branch in the cmake and plplot scripts, it wasn't getting included in the plplot build.  This build is preserved in  **scripts/wxmake_mingw**.bat (a  CMD script).

To create a custom plplot library used with GDL we use the cmake generator "MSYS Makefiles" in **scripts/appveyor_plplot.msys** and build with MSYS "make: instead of building with "mingw32-make".  This script can also be used "in place" from a home system configured (using symbolic links) to look like your git directory is /c/project/gdl.  For this custom build of plplot, three minor patches are applied from this script using the files of **/c/projects/gdl/scripts/patch**.  Two of these patches involve the WIN device driver, a) correct the keyboard focus in both wingcc and wingdi, and b) align structures in wingdi so that it can be used in GDL identically to wingcc.  In the GDL cmake command, an entry for the plplot-driver selector USE_WINGDI_NOT_WINGCC swaps the (patched) wingdi.dll in place of wingcc. Improvements in axes and labelling can be noticed, but there is still no anti-aliasing of the plot line.

  Using MSYS,  cmake scripting uses "unix-style" to find wxWidgets via extensive use of the /bin/wx-config script.  This usage is improved in the plplot build with a minor patch creating a convenient way to locate the library outside of its usual (as pointed to by wx-config)  location. 
[...]

PR #480, opened 2018-10-12
Try speedup where()

GillesDuvert:
New version, on par with IDL speed.


maynardGK (commented):
This looks fine except for the compilation errors in the gcc or clang travis builds.
I merged and test-built a windows version in order to prove that it works as expected.

The other travis error is the same encountered by the #478
 (#478 is the same as #431)
 The command "if [ -v CODECOV ]; then make codecov; else make check; fi" exited with 2.

Since #478 has extensive changes to list.cpp it would be much simpler to incorporate #478 into master prior to bringing in the changes here, which affect list.cpp only modestly.

maynardGK:
Note: PR #478 is the same as #431

alaingdl:
Good news : I made some tests in performance (3 machines) with pretty good results, except on OSX.

I also made tests with large inputs arrays, assuming its triggering the OpenMP version ...
I got consistent results (initial array and re-ordered copys of the original array)
[to be put in testsuite/ ASAP]

alaingdl:
@GillesDuvert 
Bad news : in CI I can reproduce [the problem](https://travis-ci.org/gnudatalanguage/gdl/jobs/440830880) when Eigen3 is not activated. 

in _includefirst.hpp_ adding 
```
#include <cstddef>
#include <cstdlib>
```
help to remove some messages, but I found no solution to remove : 

/tmp/gnudatalanguage/gdl/src/includefirst.hpp: In function ‘void* gdlAlignedRealloc(void*, std::size_t, std::size_t)’:
/tmp/gnudatalanguage/gdl/src/includefirst.hpp:90:27: error: ‘newsize’ was not declared in this scope
  return std::realloc(ptr, newsize);

alaingdl:
@GillesDuvert I got it, it is just a typo !! s/newsize/new_size/

GillesDuvert:
Well, as Travis is here to test what I cannot/wantnot test locally.
Hope we can merge this PR once this (petty) problem is solved.
I have noted the comment by @maynardGK and will make global tests.

alaingdl:
@maynardGK  I am very sorry but I would prefer *not* to wait for your modifications on HASH() & LIST() because at least two teams are _waiting_ for improvements on WHERE()

GillesDuvert:
Actually, gdAlinedXXXX is not aligning anything at all when not using Eigen::. This is bad, as we need aligned allocs even in the absence of Eigen:: (?? who would want a GDL without Eigen:: ?? you tell me!).
Several solutions are available, none of them are bound to work on every machine/os. Having  machine/os indpendent  gdlAlignedMalloc and Realloc woudl be a plus.

alaingdl:
@GillesDuvert 

when OpenMP is desactivated, I have problem due to

```
int thread_id = omp_get_thread_num(); //MINMAX_THREAD_NUM;
```
(multiples occurence, cannot compile)

in similar code _minmax_include.cpp_, you put 
```
int thread_id = MINMAX_THREAD_NUM;
```

Should I do it like that ?!

GillesDuvert:
yes, I forgot this issue. You could keep the same #define logic and use some `WHERE_THREAD_NUM=1`
But the case "1 thread" is explicitely written outside of the threaded part, so putting all the theraded code between `#ifdef _OPENMP` and `#endif` should do the job cleanly.

maynardGK:
@alaingdl There is no wait, hash and list are done. #431 
As bonus, it will also create a usable win32 distro.

alaingdl:
I need to go further to test some cases on OSX & Linux, not clear why AppVeyor build failed since compilation i OK

alaingdl:
this version should be OK with and without OpenMP

PR #489, opened 2018-10-23
avoid unaligned free() in Where() in a rare case

GillesDuvert:


GillesDuvert:
Test_where fails on windows for reasons unkown (related to memory alignment also?). But this current patch is important as it cures a real bug on unix/linux/mac.

PR #491, opened 2018-10-24
Parallel random

GillesDuvert:
Use dSFMT, the double-precision SIMD-accelerated Mersenne Twister code.
Accelerates noticeably the double precision RANDOMU and RANDOMN functions.
Speed increase is less extreme with floats as these are converted from doubles.


GillesDuvert:
This is too early. The random generation should be parallelized for large numbers and the state should be available as is the case for the random() legacy version. See what python::randomgen does.

PR #493, opened 2018-10-25
Revert "various improvments in the benchmark"

GillesDuvert:
Reverts gnudatalanguage/gdl#490
as "read_ascii_basic()" is not present and thus benchmark_info_os fails on linux


alaingdl:
hi @GillesDuvert 
I don't understand ! the function src/pro/utilities/read_ascii_basic.pro does exist and I carefully tested this PR on 3 machines (2 Linux and 1 OSX)

Yes, you need to have /src/pro/utilities in the path ... 
and yes, usaully, I used "quick_start_GDL.sh" to statup the env.

For sure, we do have to add sub. dirs of src/pro in the distros

PR #495, opened 2018-10-25
WIP: Appveyor classic

maynardGK:
appveyor_classic
   The functionalities of the appveyor remain with some stylistic improvements:
 1. different platforms are recognized and environmental variables created for them.
 2. continuation lines and output redirection are used for the wxwidget,xdr, and gsl builds.
plplot is still not expected to include wxwidgets: it does a version check that fails
 because the downloaded version is more advanced than the checking module
 (see scripts/patch/findwx.cmake.patch)


maynardGK:
Still, the AppVeyor.yml script does not survive the appveyor_plplot.msys build,
unlike build 624

PR #497, opened 2018-10-29
Use the dSFMT library that provides a simd-oriented fast mersenne transform.

GillesDuvert:
... And can even be further parallelized.


GillesDuvert:
OK , usual, fails on windows due to test_where, not this merge per se. So, related to #478?
Another puzzle is, why plplot (appveyor) does not find wxwidgets (sucessfully compiled)?
At the moment, there is no need to wait to merge this PR.

alaingdl:
same problem for me : all is ok on the Travis side, pb on Appveyor ...

let's go :)

alaingdl:
go

to be test, including for values and benchs

slayoo:
@alaingdl @acoulais  @GillesDuvert Please, please, please do disable the tests that are to be broken with a given merge before merging - otherwise (when master has failing CI) all other PRs are effectively marked "red"...

slayoo:
BTW, Debian does have the dSFMT package: https://packages.debian.org/source/sid/dsfmt
If there is a really good reason to have the convenience-copy code within GDL tree, we should at least provide a CMake option to use external one - otherwise this will most likely delay Debian (and hence Ubuntu) package release

GillesDuvert:
Hi,
I've taken care to update the tests specific for random values, so they pass everywhere in travis and appveyor, I've checked (even if it is... painful).
@slayoo you probably refer to the faulty test_where. Could you remove it on appveyor?
It should indeed be avoided on appveyor, so we get green lights, but this will not be completely satisfactory:
- avoiding it means we'll forget this problem (unless an issue has been opened) --in this particular case (windows build) there would be almost no chance I could investigate.
- which drives a second comment, that, despite its interest, mixing appveyor and travis builds in a PR is very frustrating as is it evident that the appveyor build scripts, for gdl or the libraries needed, are very unstable and always give some unrelated problem. I repeat my demand to have a separate branch for Windows where Windows experts could leisurely furbish GDL-for-Windows.

> BTW, Debian does have the dSFMT package: https://packages.debian.org/source/sid/dsfmt
> If there is a really good reason to have the convenience-copy code within GDL tree, we should at least provide a CMake option to use external one - otherwise this will most likely delay Debian (and hence Ubuntu) package release

@slayoo numpy's randomgen have their own version of dSFMT, as we have, and nobody's complaining! The reason why we have our own copy is that, not only it can be present on every platform disregarding the presence of sid/dsfmt, but it has been slightly modified to suit our needs and will probably be modified again (keeping copyrights of course), as not all functions needed are in sid/dsfmt. With dSFMT as with many other small useful faster-than-light code, the best performances are when the code is modified as to fit exactly to suit GDL's APIs etc. 
One could certainly change GDL to use stock dsfmt but I would advise against it.

slayoo:
test_where removed on appveyor: https://github.com/gnudatalanguage/gdl/commit/0eb2cd8d7cbcac4cc6cca40612284ed8209d1213

slayoo:
@gnudatalanguage/gdlpackagers any comments concerning the convenience-copy of dSFMT? Thanks

slayoo:
Concerning windows branch, I fully understand the frustration... although I guess the vision of having windows-osx-linux checks for new features (e.g., recent geoTIFF branch, etc) is correct. And I doubt relocating windows development elsewhere will make anyone's (non-win devs, win devs, users) life easier...

olebole:
As long as there is a way to use the package provided by the system (optimally by configuration option), I am fine with it.

slayoo:
> test_where removed on appveyor: [0eb2cd8](https://github.com/gnudatalanguage/gdl/commit/0eb2cd8d7cbcac4cc6cca40612284ed8209d1213)

Still failing (https://ci.appveyor.com/project/slayoo/gdl/branch/master/job/1jmgavjlxk1b2f99?fullLog=true)
```
The following tests FAILED:
	103 - test_tv.pro (SEGFAULT)
	105 - test_window_background.pro (SEGFAULT)
```

PR #499, opened 2018-10-30
several minors changes

alaingdl:
* in src/pro/utilities, adding a file to simply know the name of current routine (inside it)

* various changes to check and set X or WIN (or MAC) and exit otherwise. Should allow some tests to work on MSwin _soon_ 

* better type checks in ROBERTS, SOBEL & PREWITT (I saw some others codes which need similar improvments. no time for that now.) We still not have tests for thoses three functions (my tests are not useful becuase of large size of ref output, based on Saturn.jpg)


alaingdl:
how can we skip the travis tests when the change impact only the Appveyor side ?!

slayoo:
you can insert "[skip ci]" in a commit message to disable Travis checks

PR #500, opened 2018-10-31
Hash/List functions, finished, by MaynardGK - minimal PR.

GillesDuvert:
A tentative to replace #431 with all code-related improvements from @maynardGK , beneficial for all, without the commits related to Appveyor building.


maynardGK:
I think your gdl build in (commit 0608771) missed wxwidgets  because the find_package does not have 3.0.4 as a possible version number.  If you pull in 3.0.2 instead, the same wxWidgets build can be done and it will be recognized by "cmake gdl -G Mingw", but then it will pick up the static wx components.  If this is successful (use instead wxWidgets_ROOT_DIR), the wxWidgets may be missing libs: jpeg, lzma, and tiff as it does in the shareable image case.

GillesDuvert:
@maynardGK  thanks, my recent patches on appveyor are just a bunch of hit-and-miss tests. I just wanted to test all test procedures, see exactly what fails (some that were avoided in fact work). Just to get an idea. Finally I'll restore the .yml files to a more "normal" state.
Why do we recompile plplot, wxwidgets and gsl ? Cannot we download binaries like for zlib etc?

maynardGK:
> Why do we recompile plplot, wxwidgets and gsl ? Cannot we download binaries like for zlib etc?

For plplot, it really has no "default" build - it can be made for any number of softwares.
wxWidgets and GSL are both available on the msys2 repositories.  The appveyor.yml + scripts/ files
found in the latest #478 will do these downloads.
 wxWidgets will not always bind to the rest of GDL cleanly, and for reasons unknown this build worked
for Jeongbin using mingw32-make.   For me it was broken a long time also, using msys, until I removed the "-mwindows" in the link with GDL.  That change is made (using sed on wx-config) when wx is downloaded.
The format of wxWidgets is different for unix style (which uses <top>/lib, /include, and /bin) and for
windows (which uses the directory libgcc_dll created in the make).  In early days of #431 , I  tried to make a "converted" wxWidgets from the binaries in the downloaded  distribution but that never worked. When cmake is invoked with "Mingw Makefiles" it will take the Win32 style; "Msys Makefiles" uses the unix style.  In FindwxWidgets.cmake, there is a lot of coding for either style, separately.

GillesDuvert:
I've messed (again) with this PR branch. Closing and opening a clean one.

PR #505, opened 2018-11-03
Check failure where windows

GillesDuvert:


GillesDuvert:
closing -- will replace by cleaner patch

Issue #498, opened 2018-10-30
EXECUTE and undefined procedure:


alaingdl:
current Git version _diverges_ when just calling EXECUTE on an undefined procedure
(on my laptop, 100% CPU then all the memory is going to be used !)

Few months ago, it was OK :

```
GDL> res=execute('plotsym123')
% Procedure not found: PLOTSYM123
```

From my laptop, it seems that the regression appeared between April 25 and 26
(ofmt.hpp, objects.cpp, dpro.hpp, datatypes_minmax.cpp, widget.cpp)

alaingdl:
I saw nothing in the Git diff history at that date (it is not the first time we had a problem) but the situation is clear on my machine, maybe in relation with the SVN code ?????

sorry not being more useful

Alain

PS: it seems that the errors might come from change in _dpro.hpp_

```
164c164
< 	i != key.end(); i++, ix++) if( searchKey(*i)) {
---
> 	i != key.end(); ++i, ++ix) if( searchKey(*i)) {
413c413
< 	 c != common.end(); c++)
---
> 	 c != common.end(); ++c)
```
and 3 more : lines 428, 439, 460

Issue #492, opened 2018-10-24
double call to ROUTINE_INFO() crash ...


alaingdl:
```
GDL> a=routine_info(/sou)
GDL> a=routine_info(/sou)
```
gives a segfault ...

(tested on various versions on Linux)

maynardGK:
It looks like routine_info needs a non-zero number of procedures when called like this.
  If there are none, it will in any case call 
```
stru = new DStructGDL( stru_desc, dimension(  0));
```
and return the result (probably hosed at this point).

PR #508, opened 2018-11-05
Solves #492 and #498

GillesDuvert:


GillesDuvert:
removed. will open 2 PRs, one for #498 the other for #492

PR #510, opened 2018-11-06
Solves 498

GillesDuvert:
very ugly as it introduces a parameter in Run(). Better than not having autoprint?


GillesDuvert:
Still, as mentioned, .g files should be updated.

opoplawski:
Hmm, I guess it was - why no CI build tests for it?

GillesDuvert:
It's because you recompile from the .g files using "stock" ANTLR.
As I mentioned, .g files should be updated. (easy)
Build checks are not using the .g files.
Such situation is bound to happen again if Travis does not check it.

PR #512, opened 2018-11-08
Amply-featured appveyor builds using MSYS2

maynardGK:
This PR brings over the scripts and appveyor.yml that will build, successfully, 32-bit and 64-bit
windows GDL. As these scripts appear here, the 32-bit executable will have both GraphicsMagick and wxWidgets integrated.
To manipulate which methods are employed, between graphicsmagick and wxWidgets, the powershell env variables GMNAME and WXNAME are used.
  For GraphicsMagick an 8-bit image is used from the MSYS2 distribution, although it can be manipulated to attempt to build and incorporate a 32-bit library.  This does work in my home but fails on Appveyor due, probably, because of the rolling nature of the MSYS2 project.  (My home MSYS2/mingw system does not get rolling updates since it was used to test plplot a few years ago.  It uses GCC 5.3 whereas the current mingw-w64 version is at gcc 8.2).

  plplot-5.13.0 is built linked with wxWidgets-3.0.2, which is also used when GDL is built.    Patches to the  drivers/wingcc and wingdi are made that allow proper focus control. Also in the wingdi case, a feature is struck out that gdl won't use.  PLplot builds very quickly in this mode, and because it is specialized, its install target gets a separate directory in /c/projects/gdl/mingw/.  Similarly, the wxWidgets is kept in a separate directory and the FindwxWidgets.cmake is patched to allow plplot to divert there.

 bsd-xdr is downloaded from google and built very quickly.  At 64 bits there are some type conversion warnings which are suppressed (with -Wno-); this looks o.k. because the code involved are concerned with the lower-order bytes.

 Attempts were made to build wxWidgets-3.0.4 using cmake -G "Mingw Makefiles" and use it for plplot and for GDL.  Plplot on appveyor failed to integrate the wxwidgets device until -DCMAKE_CXXFLAGS='-std=c++11' was taken off of the cmake command (see scripts/appveyor_plplot.msys); this will probably work for the MINGW case, also. 
 The default wxwWidgets, 3.0.2-17, from the MSYS2 respository includes a patch that allows programs to avoid the use of "D_UNICODE" flags.  This patch is designed for 3.0.2 and is re-applied, which works out fine.  If another version of wxWidgets is used though, the patch will not be so effective and either a new patch need be made or "D_UNICODE" needs to be set in CFLAGS.

Eigen3 will generate a lot of warnings if GCC is version 7.2, 7.3, or 8.2.  To avoid this I set gcc to be 6.3 by setting the env variables $env:CC, $env:CXX.


maynardGK:
Note in continuation: at 64-bits, the Appveyor MSYS2 build doesn't work, so a mingw32-make build is done (mingw_bldgdl.bat). Without wxWidgets or graphcismagick.

GillesDuvert:
:1st_place_medal:

maynardGK:
I cancel the travis build in this commit - the appveyor build will make the same binaries but have more tests.

maynardGK:
test_make_dll hangs 64-bits appveyor tests.

maynardGK:
Substantial changes made at this point in the preparation for "make check" in appveyor_gdl.msys.

maynardGK:
The commit (069b5df) pushes the 64-bit build to the msys by setting $env:platform="" after establishing the GCC and MSYSTEM variables.  It fails:

```
[ 16%] Built target antlr
[ 17%] Building CXX object src/CMakeFiles/gdl.dir/datatypes.cpp.obj
C:/mingw-w64/x86_64-6.3.0-posix-seh-rt_v5-rev1/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/as.exe: CMakeFiles/gdl.dir/datatypes.cpp.obj: too many sections (48310)
C:\msys64\tmp\ccVMABsM.s: Assembler messages:
C:\msys64\tmp\ccVMABsM.s: Fatal error: can't write 27 bytes to section .text of CMakeFiles/gdl.dir/datatypes.cpp.obj because: 'File too big'
C:/mingw-w64/x86_64-6.3.0-posix-seh-rt_v5-rev1/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/as.exe: CMakeFiles/gdl.dir/datatypes.cpp.obj: too many sections (48310)
C:\msys64\tmp\ccVMABsM.s: Fatal error: can't close CMakeFiles/gdl.dir/datatypes.cpp.obj: File too big
make[3]: *** [src/CMakeFiles/gdl.dir/build.make:190: src/CMakeFiles/gdl.dir/datatypes.cpp.obj] Error 1
make[2]: *** [CMakeFiles/Makefile2:130: src/CMakeFiles/gdl.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:315: testsuite/CMakeFiles/check.dir/rule] Error 2
make: *** [Makefile:240: check] Error 2
```

maynardGK:
@GillesDuvert @slayoo @alaingdl 
This completes the baseline win32 build that I feel we should move to wrt appveyor.yml; please review and merge.

 Both 32- and 64- bit versions are built and they don't pass all of the tests, but the build doesn't fail on these tests.  Tests that would hang appveyor  are excluded, which includes most tests involving plots. 

The scripting uses two tracks the "msys2" track and the "mingw" track.  The mingw- track build is triggered by $env:platform="mingw64630x8664"  and the sys2 track is all other platforms.  By
re-setting the variable $env:platform after it has been used to set up other variables, the mingw track can be used for other platforms.

  Only platforms using the GCC 6.3 compiler are designated, becauses the use of a more recent gcc release results in a headache of warnings from the eigen3 installations.

   32- and 64- bit builds have distinct behaviors; indeed, 64-bit will not compile GDL under msys2.  
 Generally much cleaner results are obtained when compiled or run on a native system.  To operate an appveyor-similar build on my local system, I created NTFS symlinks on C: such that C:\projects\gdl points to my local GDL git fork, and C:\mingw-w64\ contains a set of GCC compiler distributions (NTFS symlinks can also be used so that this directory does not actually reside on C:).  Then, the appveyor scripting can be followed (approximately) with cut-and-paste to a powershell console window.  Also, a few of the artifacts left include mingw/ and win32libs/ which can be unpacked directly in the gdlgit tree for this.

maynardGK:
This PR result has been incorporated into #520

Issue #511, opened 2018-11-07
libproj path changed in debian 9


alaingdl:
The path to libproj.so has changed in Debian and is now under /usr/lib/x86_64-linux-gnu/
(before /usr/lib/). 

It is not detected by current CMake files.
Could some one helps to improve the detection ? Thanks !

A.

alaingdl:
Due to MultiArch tricks

locally solved using
https://wiki.debian.org/Multiarch/Implementation#CMake

```
include(GNUInstallDirs)
```
in CMakeLists.txt 

to be in a PR ASAP but I have another related bug to solve before

(no effect on a previous bug related to no-detection of NetCDF-4)

alaingdl:
done in #515

PR #515, opened 2018-11-10
small issue in Basic_benchmark, better finding for recent OS of libs.…

alaingdl:
should solve MultiArch #511 

fixing a detail in "basic_benchmark.pro"


alaingdl:
unclear for me why it is failing within Clang !

slayoo:
the OSX issue was fixed here (a homebrew hack that was no longer needed): https://github.com/gnudatalanguage/gdl/commit/5992075c4a051f13dc215e40ca47974805aade8d

PR #518, opened 2018-11-14
Fix annoying bugs in dSFMT-aware random() functions.

GillesDuvert:
bugs fixed:
1) (arbitrary) maximum number of threads was not taken into account properly (crash!)
Here the seed parallel structure is created with the maximum number of threads available. Should cover all cases, even machines with 24 or 48 threads. Only drawback: initial seed generation time increases with number of available threads.
2) in the absence of seed, the sequence was not using a new systime(), so different implicit seeds were the same. Bug cured, for gsl and dSFMT variants.


alaingdl:
successfully tested on my 48-cores machine :)

Issue #198, opened 2018-04-18
Insure  ANTLR - created files (GDL*.[c|h]pp files) are compatible with their ANTLR sourcefiles (*.g)


GillesDuvert:
Sometimes one make patches to GDL*.[c|h]pp files *ONLY*
These patches should not be accepted if they are not present also in the parent *.g files, as the GDL* files are bound to be recreated by `for i in *.g; do antlr $i; done;`
It is necessary to add (in Travis?) a test to check that such a situation does not happen.

slayoo:
A relevant comment from @opoplawski suggesting to remove the generated files from the repo:
https://github.com/gnudatalanguage/gdl/pull/274#issuecomment-394369044

slayoo:
+ we need also to ensure that the convenience-copy antlr headers are replaced upon regeneration! Should we then remove these (the whole antlr directory) from the repo @opoplawski?

slayoo:
This is addressed here: https://github.com/gnudatalanguage/gdl/pull/529
The question of removal of antlr-generated files and the headers is in fact part of the still open: https://github.com/gnudatalanguage/gdl/issues/45

PR #529, opened 2018-12-01
WIP: Initial attempt to implement throwImmediately in ANTLR grammar

opoplawski:
This is an initial attempt to implement the changes in #510 in the ANTLR sources.  However, I'm not able to pass arguments in the statement rules.  Reading https://stackoverflow.com/questions/21747097/antlr-rules-with-arguments leaves me to believe that it should be possible, though highly discouraged, but it isn't working for me with antlr 2.7.7 at least.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
     - os: linux
       env: COMPILER=gcc   CMAKE_BUILD_TYPE=Debug    DEPS=FULL  CODECOV=yes
     - os: osx
-      env: COMPILER=clang CMAKE_BUILD_TYPE=Debug    DEPS=FULL
+      env: COMPILER=clang CMAKE_BUILD_TYPE=Debug    DEPS=FULL  ANTLR=yes
 
 #    - os: linux
 #      env: COMPILER=icc   CMAKE_BUILD_TYPE=Release  DEPS=MINI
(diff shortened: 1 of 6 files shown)


opoplawski:
Okay, looks like perhaps had to re-order.  Harder to compare the differences in the generated files now though.  Starting to run builds...

opoplawski:
Builds seem to be fine here.  Someone should double check these changes, but I think they are correct at this point.

slayoo:
#198 - would be great to kill two birds with one stone

slayoo:
The travis script now includes a block with instructions for checking if the generated files match in input .g files

GillesDuvert:
I suggest to resubmit a patch with minimal changes in the .g file (here some blocks of text has been unnecessarily moved). The only reason is to see clearly the small incremental change in an otherwise cryptic language, given that a better solution to bug #498 could be sooner or later be proposed by our ANTLR expert, Marc S.

GillesDuvert:
Besides, thanks @slayoo for the travis patch and closing of #198 .
Still, there will always be the doubt about how GDL behaves with or without using different flavours of stock ANTLR (not speaking about old versions of ANTLR being dropped at some point). Compared to the need to reconstruct with C++ compiler only on mr everybody's machine.

slayoo:
@GillesDuvert First step - revert: https://github.com/gnudatalanguage/gdl/pull/543

PR #539, opened 2018-12-04
Revamp sorting codes

GillesDuvert:
Rewrote SORT() using modern methods for a faster sort. Result is a noticeable-to-huge speed gain, depending on array sizes and types. This version uses 1) modified code from https://github.com/Pierre-Terdiman/RadixRedux (adapted to GDL and other types than floats and longs) plus 2) new versions of other sorting methods, as the various codes found on the web do not sort indexes but values and do not sort NaNs at all.
I've added a GDL_SORT() "private" function that permits to use one of: /RADIX, /QUICK, /MERGE and /INSERT sorting method (as well as /CHECK the result). This function is used in the procedure testsuite/benchmark/compare_sort_algos to plot the performances of all the different sorts (insertion is not used as it is in N^2 and would take forever) and adjust the combination of all these methods that makes for the performance in the final sort().
With a modicum of work, this could be used to fine-tune the SORT() method for a particular machine or purpose, maintaining top performances in specific cases (clusters etc).


GillesDuvert:
oops, temporal coherence is lost for N>1E5. Do not use yet.

GillesDuvert:
any hint why this does not pass with windows? :frowning_face:

GillesDuvert:
closing --- wil submit a cleaner PR.

PR #546, opened 2018-12-10
Fast median filter

GillesDuvert:
Finished 1D and 2D median filter optimization, with incredible speed gains for even and odd-sized data.
(see bench_median!)
Note that the sliding window can have 2 different values, as in B=MEDIAN(A,[33,68]) which is now permitted by GDL (a Warning is issued that this is a GDL-only feature).
Old code removed.


GillesDuvert:
oops, a branch mismatch. to be redone.

PR #547, opened 2018-12-10
Patch linfit

GillesDuvert:


GillesDuvert:
apparently there is a spurious "stop" command left in the code.
Tests needed.

PR #548, opened 2018-12-10
Faster median filter

GillesDuvert:
finished 1D and 2D median filter optimization, with incredible speed gains for even and odd-sized data.
(just check with bench_median.pro in testsuite/benchmark)
Note that the sliding window can have 2 different values, as in B=MEDIAN(A,[33,68]) which is now permitted by GDL (a Warning is issued that this is a GDL-only feature).
Old code removed.


GillesDuvert:
oops! wrong cut and paste!

GillesDuvert:
closing temporarily --- rebasing code was definitely a good idea.

Issue #555, opened 2018-12-20
bug when first call of CONTOUR,/overplot


alaingdl:
In a new session, the following sequence  don't plot at all the contour.
```
plot, findgen(10)
x=indgen(10)
contour, dist(10), x,x,/overplot, nlev=12, col='ff00'x
```
(a derivate version based on [Coyote](http://www.idlcoyote.com/tips/mapcontourdata.html) plots only one line !)
It should be noticed that it's happen only when contour is not initialized before

the following sequence is OK:

```
contour, dist(10) 
plot, findgen(10)
x=indgen(10)
contour, dist(10), x,x,/overplot, nlev=12, col='ff00'x
```

alaingdl:
My mistake !! 
same behavior in IDL & GDL on basic case, but it is related to mandatory "geographic mapping coordinates are in effect."

alaingdl:
nevertheless my bug is a true bug and I have a solution ! (few lines in plotting_contour.cpp)
to do a PR ASAP

alaingdl:
My solution is OK when using MAP_SET

I am trying to write a test for that, but I would like to avoid to have MAP_SET around. What should I do to circumvent it. At the beginning I think that I need only : 
```!x.type=3 & !map.projection=9```
but it is clearly not enough

GillesDuvert:
Hi,
This is arguably NOT A BUG of IDL that GDL would have somewhat mimicked.
Documentation states explictely that /OVERPLOT makes  CONTOUR “overplot” a previous CONTOUR, not anything else, like a PLOT. That is, the current graphics screen is not erased, no axes are drawn and the previously established scaling remains in effect. One must explicitly specify either the values of the contour levels or the number of levels, unless geographic mapping coordinates are in effect.
As there was no previous "CONTOUR" command in the first example, it is not a surprise that the boundaries, scale etc are not ready and thus no output happens.

alaingdl:
I hope the example in PJ will help to solve the last case

GillesDuvert:
I do not see any example in PJ ?

PR #556, opened 2018-12-23
solves #555

GillesDuvert:


alaingdl:
I will merge this PR but, in fact, it is not the final answer to the problem 

We do have CONTOUR, /overplot, _tab_, x, y

* when (x, y) are a 1d vector, it is now OK (with this PR)
* when (x,y) are a 2d array same size than _tab_, the overplotted contour is not OK

I will put in #555 a snippet for easy test that (in the revised gshhG frame ...)

A.

PR #592, opened 2019-05-03
 Align structure data to intrinsic size (as for IDL)

maynardGK:
Without performance advantage, structure data was aligned on 16-byte boundary when eigen was in use..  This aligns the data based on their intrinsic sizes

Also,  in  DUStructDesc::AddTag(), convert structure field names to upper case to avoid an easy-to-make coding error.


maynardGK:
This now passes test_n_tags.pro:

GDL> help,tmp47
TMP47           STRUCT    = -> DATA Array[1]
** Structure DATA, 12 tags,memsize =48, data length=47/47:
   YDH             LONG                 0
   NUM             LONG                 0
   T97             DOUBLE           0.0000000
   F               FLOAT           0.00000
   DT              FLOAT           0.00000
   DF              FLOAT           0.00000
   AUTOX           FLOAT           0.00000
   AUTOZ           FLOAT           0.00000
   CROSSR          FLOAT           0.00000
   CROSSI          FLOAT           0.00000
   ANT             BYTE         0
   A               INT              0
GDL> .run test_n_tags
% Compiled module: TEST_NTAGS_BY_TYPE.
% Compiled module: TEST_NTAGS_MIXING_TYPE.
% Compiled module: TEST_N_TAGS.
GDL> test_n_tags
% Compiled module: GIVE_LIST_NUMERIC.
% TEST_NTAGS_BY_TYPE:   NO errors encountered during TEST_NTAGS_BY_TYPE tests
% Compiled module: ERRORS_CUMUL.
% TEST_NTAGS_BY_TYPE:   NO errors encountered during TEST_NTAGS_BY_TYPE tests
% TEST_NTAGS_BY_TYPE:   NO errors encountered during TEST_NTAGS_BY_TYPE tests
% TEST_NTAGS_MIXING_TYPE,:   NO errors encountered during TEST_NTAGS_MIXING_TYPE, tests
% TEST_NTAGS: ===================================================
% TEST_NTAGS: =                                                 =
% TEST_NTAGS: =  NO errors encountered during TEST_NTAGS tests  =
% TEST_NTAGS: =                                                 =
% TEST_NTAGS: ===================================================
GDL>

GillesDuvert:
Hi @maynardGK , works for me.

GillesDuvert (approved):

GillesDuvert:
I still have doubts, nevertheless. I'm afraid this gonna crash on some architectures with eigen:: active (thus, optimisation at -O2 level or more) and default alignment by compiler less than 16 bytes.

maynardGK:
Note that alignment on < 16 would only be for byte data; fixed data would align at 16-bits, long at 32, etc.
Can we generate VMs to test on ARM architectures?   Maybe Ole is running them.

GillesDuvert:
16 Bytes, not bits. 
Basically we need to avoid the alignment pitfall see http://eigen.tuxfamily.org/dox-devel/TopicPitfalls.html
Meaning that we can have a perfect working code, without even a warning, but practically no speed gain. And probably erratically, at that.
I'm having secound thoughts.
As the risks are too great to produce slow code and grumbling users,  I suggest to postpone this change until somebody sets up a testsuite that would validate it on all architectures/compilers.
Besides, I'm not even sure we really use Eigen:: correctly everywhere.

maynardGK:
I did a few speed tests long ago when I first implemented it in my code.  Even if one sets alignment=2
(everything aligned at 2 byte boundaries) I found no effect.  The reason, I believe, is that any
vector operations done structures require extracting the data out of the structure first.

maynardGK:
Here's a test in my windows 7 setup, indicating a substantial speedup using the proposed alignment.
"$ gdl64" uses an earlier version of GDL built on appveyor,
"$ gdlgit" uses the current git version, with this PR modification, & built on my machine.
````
greg@Homerw7  ~
$ gdl64
% WARNING: your version of the GraphicsMagick library will truncate images to 8 bits per pixel

  GDL - GNU Data Language, Version 0.9.9 git

- Please report bugs, feature or help requests and patches at:
  https://github.com/gnudatalanguage/gdl

% Compiled module: PATH_SEP.
D:/home/greg/.gdl_startup: Exception: basic_string::substr: __pos (which is 4294967295) > this->si
() (which is 0)
D:/home/greg/.gdl_startup: Exception: basic_string::substr: __pos (which is 4294967295) > this->si
() (which is 0)
D:/home/greg/.gdl_startup: Exception: basic_string::substr: __pos (which is 4294967295) > this->si
() (which is 0)
GDL> tic & toc & rt={a:0,b:1,r:1.,d:1.D}
% Time elapsed :      0.015600204 seconds.
GDL> t=replicate(rt,100,10000)
GDL> tic & for k=0,9 do begin & tr=t.r & t.r = tr & endfor & toc
% Time elapsed :        1.8408041 seconds.
GDL> tic & for k=0,19 do begin & tr=t.r & t.r = tr & endfor & toc
% Time elapsed :        3.7148070 seconds.
GDL> tic & for k=0,99 do begin & tr=t.r & t.r = tr & endfor & toc
% Time elapsed :        18.583633 seconds.
GDL>GDL> help,rt
RT              STRUCT    = -> <Anonymous> Array[1]
** Structure <Anonymous>, 4 tags,memsize =64, data length=16/16:
   A               INT              0
   B               INT              1
   R               FLOAT           1.00000
   D               DOUBLE           1.0000000
````
````
$ gdlgit

  GDL - GNU Data Language, Version 0.9.9 git

- Please report bugs, feature or help requests and patches at:
  https://github.com/gnudatalanguage/gdl

% Compiled module: PATH_SEP.
D:/home/greg/.gdl_startup: Exception: basic_string::substr: __pos (which is 4294967295) > this->size
() (which is 0)
D:/home/greg/.gdl_startup: Exception: basic_string::substr: __pos (which is 4294967295) > this->size
() (which is 0)
D:/home/greg/.gdl_startup: Exception: basic_string::substr: __pos (which is 4294967295) > this->size
() (which is 0)
GDL> rt={a:0,b:1,r:1.,d:1.D}
GDL> help,rt
RT              STRUCT    = -> <Anonymous> Array[1]
** Structure <Anonymous>, 4 tags,memsize =16, data length=16/16:
   A               INT              0
   B               INT              1
   R               FLOAT           1.00000
   D               DOUBLE           1.0000000
GDL> t=replicate(rt,100,10000)
GDL> tic & for k=0,9 do begin & tr=t.r & t.r = tr & endfor & toc
% Time elapsed :       0.42120004 seconds.
GDL> tic & for k=0,19 do begin & tr=t.r & t.r = tr & endfor & toc
% Time elapsed :       0.62400103 seconds.
GDL> tic & for k=0,99 do begin & tr=t.r & t.r = tr & endfor & toc
% Time elapsed :        3.1668050 seconds.
GDL>
````
@alaingdl Although an exaggerated synthetic/contrived test, the performance difference is substantial (x6 or x7).  This is probably just due to the shortening of memory access needed for the task.

maynardGK:
Aug 8: @GillesDuvert @slayoo @alaingdl  I copied the original commit to a new branch (named "dstructdesc") which includes the updated master without the merger noise.

slayoo:
Closing this one then and opening #614

slayoo:
and big thanks @maynardGK !

PR #623, opened 2019-08-28
Adding support for object-methods to routine_filepath

hillbergt:
Adding support for object-methods to routine_filepath, following the behaviour of IDL.
Also added test-cases for the various ways to call routine_filepath.

### DIFF
--- a/src/basic_fun.cpp
+++ b/src/basic_fun.cpp
@@ -7166,13 +7166,12 @@ BaseGDL* routine_filepath( EnvT* e)
     if( nParam > 0)  {
         BaseGDL* p0 = e->GetParDefined( 0);
         if( p0->Type() != GDL_STRING)
-          e->Throw("String expression required in this context: " + e->GetParString(0));
+        e->Throw("String expression required in this context: " + e->GetParString(0));
         p0S = static_cast<DStringGDL*>( p0);
-      } else {          // routine_filepath()
-        p0S = new DStringGDL(
-            dynamic_cast<DSubUD*>((e->Caller())->GetPro())->Name());
+    } else {          // routine_filepath()
+        p0S = new DStringGDL( dynamic_cast<DSubUD*>((e->Caller())->GetPro())->ObjectName() );
         p0S_guard.Init(p0S);
-        }
+    }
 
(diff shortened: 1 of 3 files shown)


maynardGK:
The windows' version will work OK but the IDL test has file separator issues;
for routine_filepath it is windows but from routine_dir() it comes back unix.  

````
;
function pathfromunix,strin
; convert path from unix-style (switch '/' to '\')
bs = byte(strin)
jj = where(bs eq 47,nj)
if nj gt 0 then bs[jj] = 92
return,string(bs)
end
;
````
Then, conversion only needed on the one variable ref:
````
    ref = ROUTINE_DIR() + 'test_routine_filepath.pro'
    if !version.os_family eq "Windows" then ref=pathfromunix(ref)
````

Result
````
GDL> test_routine_filepath
System filepath for this script: F:\github\gdlgit\testsuite\test_routine_filepath.pro
% TH_PRO1234:   NO errors encountered during TH_PRO1234 tests
% TH_FUNC1234:   NO errors encountered during TH_FUNC1234 tests
% TH_STRUCT::PROC1:   NO errors encountered during TH_STRUCT::PROC1 tests
% TH_STRUCT::FUNC1:   NO errors encountered during TH_STRUCT::FUNC1 tests
% TEST_ROUTINE_FILEPATH: ==============================================================
% TEST_ROUTINE_FILEPATH: =                                                            =
% TEST_ROUTINE_FILEPATH: =  NO errors encountered during TEST_ROUTINE_FILEPATH tests  =
% TEST_ROUTINE_FILEPATH: =                                                            =
% TEST_ROUTINE_FILEPATH: ==============================================================
GDL>
````

hillbergt:
Hi Maynard,

I would not recommend converting the ref-variable here.
If ROUTINE_DIR returns a unix-path on windows-systems, then it is ROUTINE_DIR that is not working properly, since it should return a "valid" path to the current script-file.
I had a quick look and to me it looks like the Dirname function (file.cpp:1203) is the problem.

I suggest opening a separate issue for this, which is probably better investigated by someone with a working windows setup (I only have GDL under linux at the moment).

/Tomas

PR #624, opened 2019-08-29
Some magick problems

GillesDuvert:
our magick_cl tool is only compatible with imagemagick version 6 or less and definitely not with IM >=7.

Also, catch all messages and errors issued by graphicsmagick.
IM supports now HDR data, so it would be better to switch to IM (IMHO). Not a big deal.


GillesDuvert:
retry for test on travis (?)

GillesDuvert:
hm. so: work in progress

GillesDuvert:
closing as not worth the effort right now.

PR #626, opened 2019-09-01
Adding builtin function for SCOPE_VARNAME, and a test-script for it.

hillbergt:
N.B. There seem to be some discrepancies when routines are compiled, so that
the order of the COMMON blocks might be different when compared to IDL.
For that reason, using SCOPE_VARNAME with the /COMMON keyword might not have predictable outcome until the order of the COMMON-blocks have been sorted out.

### DIFF
--- a/src/basic_fun.cpp
+++ b/src/basic_fun.cpp
@@ -8540,7 +8540,111 @@ BaseGDL* parse_url( EnvT* env)
     e->Throw("LVariable not found: " + varName);
     return NULL; // compiler shut-up
   }
-  
+
+    BaseGDL* scope_varname_fun( EnvT* e ) {
+
+        SizeT nParam = e->NParam();
+
+        EnvStackT& callStack = e->Interpreter()->CallStack();
+        
+        DLong currentLvl = callStack.size();
+        DLong level = currentLvl;               // default to current level
+
+        DStringGDL* retVal = nullptr;
+        SizeT count(0);
+        
+        static int commonIx = e->KeywordIx("COMMON");
+        static int countIx = e->KeywordIx("COUNT");
+        static int levelIx = e->KeywordIx("LEVEL");
+        
+        if( e->KeywordSet(commonIx) ) {
+            
+            if( e->KeywordSet(levelIx) ) e->Throw("Conflicting keywords.");
+                
+            DString commonName = "";
+            e->AssureStringScalarKW( commonIx, commonName );
+            DSubUD* pro = static_cast<DSubUD*>( e->Caller()->GetPro() );
+            SizeT nComm = pro->CommonsSize();
+            DCommon* common = pro->Common( StrUpCase(commonName) );
+            count = nParam;
+            retVal = new DStringGDL( dimension( nParam ), BaseGDL::NOZERO );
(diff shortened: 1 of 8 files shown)


GillesDuvert:
unfortunately this PR breaks GDL, the 'dummy' byte added in dcommon.hpp fools boost_array.pro of IDLAstro by returning a "byte" type whatever the type of the common variable.

hillbergt:
Hi Gilles,
Could you please give me a code-snippet to reproduce the problem?
I've been molesting boost_array every conceivable way now, and can't get any weird behaviour. Is this a windows-specific problem, or also on other systems?

GillesDuvert:
@hillbergt , First, thanks for your involvement in GDL.
Too bad there is a glitch on this PR (the others are OK!, Thanks again!) 
Could you please:
- download a FITS example file, I used https://www.cv.nrao.edu/fits/data/samples/bintable/j000015.fits
- run the bug_pr_626 script (below) (you need some FITS idl utilities of https://idlastro.gsfc.nasa.gov/ftp )
```
GDL> bug_pr_626
% Compiled module: BUG_PR_626.
% Compiled module: FXBOPEN.
% Compiled module: FXBFINDLUN.
% Compiled module: BOOST_ARRAY.
% BOOST_ARRAY: Data arrays should be either both string or both non-string
% Error occurred at: BOOST_ARRAY         75 /home/opt/idl/liblocal/IDLAstro/pro/boost_array.pro
%                    FXBFINDLUN          95 /home/opt/idl/liblocal/IDLAstro/pro/fxbfindlun.pro
%                    FXBOPEN            248 /home/opt/idl/liblocal/IDLAstro/pro/fxbopen.pro
%                    BUG_PR_626           6 /tmp/test/bug_pr_626.pro
%                    $MAIN$          
% Execution halted at: BUG_PR_626           5 /tmp/test/bug_pr_626.pro

```
The procedure:
```
pro bug_pr_626
  file='j000015.fits'
  exten=1
  errmsg=''                     ; necessary!
  fxbopen, unit,file,exten,header,errmsg=errmsg
end
```
the correct output is
```
GDL> bug_pr_626
% Compiled module: BUG_PR_626.
% Compiled module: FXBOPEN.
% Compiled module: FXBFINDLUN.
% Compiled module: FXHREAD.
% Compiled module: FXPAR.
% Compiled module: GETTOK.
% Compiled module: VALID_NUM.
% Compiled module: FXBPARSE.
% Compiled module: FXBTFORM.
% Compiled module: FXBFIND.
% Compiled module: STORE_ARRAY.
GDL>
```

hillbergt:
Hi Gilles,

The problem is that if the common variables are undefined, they are not "comparable" against other variables (which is the whole point of the SCOPE_VARNAME function), so it has to have a type attributed to it. One way (probably the neatest way) is to add an "undefined" type, to use in these cases, but that would be a bit more work.
I now changed the placeholder-type from byte to !NULL, that seems avoid the problem you found and still makes the SCOPE_VARNAME work as intended. I hope this passes your tests.

regards

GillesDuvert:
@hillbergt unfortunately a bug was discovered (see #650) .
The bug appears with the "map_set" procedure called (which has now been written for GDL).

PR #627, opened 2019-09-01
New implementation of Dirname.

hillbergt:
A re-implementation of Dirname, which should return valid windows-paths on Windows, and unix-paths under linux/cygwin.
N.B. I have not tested this properly on Windows yet, hopefully I can do that in the next few days.

### DIFF
--- a/src/file.cpp
+++ b/src/file.cpp
@@ -1200,67 +1200,42 @@ static void FileSearch( FileListT& fileList, const DString& pathSpec,
 #endif
 
 
-static std::string Dirname(const string& tmp,
-    bool mark_dir = false)
-{
+static string Dirname( const string& in, bool mark_dir=false ) {
 
-    char buf[ PATH_MAX+1];
+    char win_sep = '\\';
+    char unix_sep = '/';
+    char path_sep = unix_sep;
 
-// G. Jung Simplify the alternatives to psalt, PS using same processing.
-// for mark_dir, always just place a "/"
+#if defined (_WIN32) && !defined(__CYGWIN__)
+    path_sep = win_sep;
+    DString dname(in);
 
-    char PS='\\';
-    char psalt[] = "/";
+    std::replace( dname.begin(), dname.end(), unix_sep, win_sep );  // replace separators in input
+    DString::size_type pos = dname.find_last_of( path_sep );        // strip filename if present
 
-#ifndef _WIN32
-    strncpy(buf, tmp.c_str(), PATH_MAX+1);
-    string dname = dirname(buf);
+    if( pos != DString::npos ) {
+        dname.erase(pos);
+    }
 #else
-   char drive[_MAX_DRIVE];
-   char dir[_MAX_DIR];
-   char fname[_MAX_FNAME];
-   char ext[_MAX_EXT];
-   DString::size_type pos = 0, offset = 0;
-   DString tmp2(tmp);
-// if 0/ if 1: by preference.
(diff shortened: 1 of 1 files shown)


maynardGK:
It would be trivial enough to convert all file names with "/" to "\\" but they weren't, for good reason. 

I operate GDL through a posix/cygwin-style shell (the MSYS2 shell).  In this way one hardly needs to notice which system is in use, especially as they both operate on the same disk files.  I use GDL on both Linux and Windows; the old-style backslashes are always a PITA, and usually the only visible difference in operation - they share the same "$GDL_STARTUP" script, in fact.  

 Windows' file specification allows the use of forward slashes since at least Windows XP.  Its use in the (DOS) CMD shell is not, however, supported.  Using that shell ("ls" is an old UnixUtils command):
For a user that wishes backslashes in their filenames we should deliver a global option to run in that mode or to remain in the posix-compatible mode.  If we were to sense what sort of shell GDL were invoked from, we could trigger the appropriate option.

````
Microsoft Windows [Version 6.1.7601]
Copyright (c) 2009 Microsoft Corporation.  All rights reserved.

D:\cmd>ls c:/
$Recycle.Bin               System Volume Information
$WINDOWS.~BT               Users
$Windows.~WS               Windows
AdwCleaner                 boot
Bovada                     cmd
Documents and Settings     extensions.sqlite
ESD                        found.000
IO.SYS                     found.001
MSDOS.SYS                  msys64
NVIDIA                     nsinst.log
Program Files              packages.txt
Program Files (x86)        profile
ProgramData                programs
Recovery                   projects
Reflect_Install.log        usr

D:\cmd>dir c:/
Invalid switch - "".

D:\cmd>dir c:/packages.txt
Parameter format not correct - "packages.txt".
````
The Windows 10 ISO allowed me to install, without a valid code. ( Although I don't know how long it will last).  I recommend using a separate disk drive to hold it if another system is hosted on the machine.

GillesDuvert:
@maynardGK you commented this PR but there is no specific comment on this patch and its consequences on Windows . The test procedures for filenames, dirnames etc work on  the Appveyor build. So a merge is mandatory, yes?

maynardGK:
I presume the patch works but it changes behavior in Windows that I don't agree with.  In my usage,
(which occurs through a posix shell on the console) it is more convenient that -fewer- backslashes appear in reported filenames, not more.   
  There are no applications that I know of that need backslash protocols- powershell, for instance, is quite comfortable with forward slashes.
````
PS C:\Users\greg> ls c:/boot


    Directory: C:\boot


Mode                LastWriteTime     Length Name
----                -------------     ------ ----
d----         8/21/2019   7:18 PM            macrium

````

hillbergt:
The motivation for rewriting DirName was triggered by #623 (as noted by maynardGK), namely that ROUTINE_DIR returned a path with /-separators under Windows, which broke 'test_routine_filepath.pro'. (actually, I think the problem was that the path contained "both" "/" and "\\")
So you *do* know of at least 1 instance where backslashes were required. ;-)

I stand by my opinion that this is the right fix, otherwise the output of ROUTINE_DIR & FILE_DIRNAME could potentially break other scripts in the same way as it did with test_routine_filepath in #623.
Also: IDL always returns backslash-paths under Windows, does it not?

maynardGK:
I acknowledge the usage of "\\" or "/" has been inconsistent and I've coded (enough, I think) a
solution that allows posix or windows file specifications in #635.

GillesDuvert:
closing & reopening to be sure

PR #629, opened 2019-09-03
Faster ?  FINITE() , CONVERT and TOTAL() functions.

GillesDuvert:
Cannot be slower, anyway. Passes 221/225 tests here...


GillesDuvert:
OK, conflicts with recent merges are too large. Will resubmit a simpler PR.

PR #631, opened 2019-09-08
Added implementation for TERMINAL_SIZE. No support for setting width/…

hillbergt:
…height yet though.

### DIFF
--- a/src/basic_fun.cpp
+++ b/src/basic_fun.cpp
@@ -7536,7 +7536,18 @@ BaseGDL* routine_filepath( EnvT* e)
     *p0 = NULL; // make parameter undefined
     return ret;
   }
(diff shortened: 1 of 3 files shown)


GillesDuvert:
close and reopen, to be sure.

GillesDuvert:
reopen, test compiles same ?

maynardGK:
My guess is that TermHeight() is acquired accidentally when GCC is used with DEPS=FULL;
There's no assignment to lib:: in terminfo.cpp but how does a TermWidth() call (from gdl and from print) pass through?

alaingdl:
related to what I did last week. Problem of missing TermHeight() should be solved

I will accept it now and check further

PR #637, opened 2019-09-23
WIP lib tests for windows support (change the entries in scripts/appveyor_1.msys)

maynardGK:
This is a branch off of posixpaths which has green in both CI builds.  These will use gcc 6.3 for both i686 and for x86_64 builds.  gcc 7.3 is also available for x86_64 but not for i686 builds.
1. readline. try 8.0: Version 7 did not work under gcc 6.3


maynardGK:
The artifacts (gdl_build.zip) do not contain the readline shareable images ! This seems to be a random failure, as artifacts in the next run (which fails for 32-bit compilation) do contain the readline8/libhistory8 librarys.

PR #642, opened 2019-10-02
Clean up some compiler-warnings.

maynardGK:


slayoo:
fails with clang on:
```
gdl/src/math_fun_ac.cpp:771:25: error: call to 'abs' is ambiguous
```

maynardGK:
Is a Python.h inclusion really needed on every compilation?  It is called out in "includefirst.hpp" and so is included for every module, which seems excessive. @hillbergt @olebole 
''''
[ 21%] Building CXX object src/CMakeFiles/gdl.dir/datatypes.cpp.o
In file included from /Users/travis/build/gnudatalanguage/gdl/src/datatypes.cpp:18:
In file included from /Users/travis/build/gnudatalanguage/gdl/src/includefirst.hpp:47:
In file included from /usr/local/Cellar/python@2/2.7.16_1/Frameworks/Python.framework/Versions/2.7/include/python2.7/Python.h:88:
''''

PR #643, opened 2019-10-03
CMakeList: added CentOS, Fedora/CentOS package suggestions

pkubanek:
* replaced Fedora with Fedora/CentOS, to make sure CentOS uses same package names
  * added more Fedora/CentOS package hints


slayoo:
Thanks!

PR #645, opened 2019-10-03
 use macro to constrain processing the Python.h inclusion.

maynardGK:
INCLUDE_PYTHON macro as a flag to consider/not Python.h in includefirst.hpp
````
modified:   src/basegdl.cpp
modified:   src/basic_fun.cpp
modified:   src/datatypes.cpp
modified:   src/gdl.cpp
modified:   src/gdlpython.cpp
modified:   src/includefirst.hpp
modified:   src/libinit.cpp
modified:   src/nullgdl.cpp
modified:   src/objects.cpp
modified:   src/pythongdl.cpp
````
Anything that touches readline seems to require the fulll python treatment.

Added, after experimentation here:
````
basic_pro.cpp
dinterpreter.cpp
gdlhelp.cpp
read.cpp
terminfo.cpp
````


GillesDuvert:
@maynardGK this will clash with #629 and probably #468. Let's wait for #468 to happen.
Besides, most of the diffs are due to a difference of formatting, making PR unreadable.

maynardGK:
Sorry I hadn't noticed the diffs being so unwieldly. Is that just from whitespace or is it a CR/LF thing?  
Mostly, of course, this PR requires INCLUDE_PYTHON to be set before Python.h is actually invoked from includefirst.hpp; there's no hurry to incorporate it, I mostly wanted to see how small of an impact it would have.

PR #647, opened 2019-10-04
Add sincos fallback for OS X 10.8 and earlier

ryandesign:
This PR allows gnudatalanguage to build on OS X 10.8 and earlier. (`__sincos` was first available in OS X 10.9.)

It might be even better to add detection of `__sincos` to the configure script, rather than hardcoding information about different operating systems in this file.


slayoo:
Thanks!

PR #654, opened 2019-10-12
Implement PR 635 on recently revamped master.

maynardGK:
Under windows,  designation of filenames uses the backslash notation ("\") as directory separation character.  This can be overridden for use in POSIX shells so that files are reported using the forward slash, exclusively.     To run in posix mode, invoke using "--posix" on the invocation line or, alternatively, by setting the GDL_USE_POSIX environment variable.
    The posix/non-posix condition can be determined with the (new) !GDL.GDL_POSIX read-only system variable. In non-windows builds, !GDL.GDL_POSIX is set to 1 (true).  The procedure PATH_SEP()
is written to use this in a backwards-compatible manner.


maynardGK:
Note: Running gdl built with this code on my computer gives NO errors for test_routine_filepath, both with/without GDL_USE_POSIX set.

maynardGK:
Result of "make check" after build on suse:
````
99% tests passed, 3 tests failed out of 229

Total Test time (real) =  78.00 sec

The following tests FAILED:
	 98 - test_fft_leak.pro (Failed)
	141 - test_make_dll.pro (Failed)
	185 - test_resolve_routine.pro (Failed)
Errors while running CTest
gmake[4]: *** [Makefile:96: test] Error 8
````

GillesDuvert:
@maynardGK  sorry sorry I forgot to follow this PR, and now the "files_changed" list shows changes in files that are not related to your posix patches (?) . Is that because your gdl_posix branch must be realigned to gdl/master ? All this (git or github?) is very confusing to review a PR.

maynardGK:
There were a few files in testsuite/benchmark/ which I changed to what looked like the correct
version, and also testsuite/test_plot_linestyle.pro.  Here is the list I see:
````
2  scripts/appveyor_gdl.msys
 6  src/dialog.cpp
 24  src/dpro.hpp
 38  src/file.cpp
 4  src/file.hpp
 11  src/gdl.cpp
 7  src/initsysvar.cpp
 36  src/pro/path_sep.pro
 3  src/sorting.cpp
 8  testsuite/benchmark/bench_matrix_multiply.pro
 8  testsuite/benchmark/benchmark_plot_cartouche.pro
 10  testsuite/test_file_search.pro
 97  testsuite/test_plot_linestyle.pro
````
I can't conjure up the command that will update my local gdl_posix to include the latest edits for smooth, etc.  I was able to get the new material merged on my old computer, so that is pushed next.

GillesDuvert:
OK, thaks!

slayoo:
@acoulais @maynardGK @GillesDuvert is it then good time for 1.0.0-rc2?

alaingdl:
Can we wait 2 or 3 days (I am supposed to work tomorrow)

It would have been great to solve super simple #678  before

I have few details to push, not urgent (eg manage the path to map files in demo ...)

PR #655, opened 2019-10-20
WIP: attempt Appveyor with updated libraries

maynardGK:
here on Appveyor the gcc version is 6.3; first reproduce previous result of #637 .


maynardGK:
readline 8 was loaded in x86_64 artifact this time, and i686 release worked out as usual but debug mode
stuck on test_constants.pro - which is just a kick-out for not being compiled with UDUNITS2.
Otherwise running (i686) from a native system:
  test_angles - failure
test_binfmt - success
test_base64 - success
test_byte_conversion - failure, 4 errors
... etc mostly successes with the i686 artifact, even on tests that failed on Appveyor.

maynardGK:
General conclusion of the experiment: changes in gcc between versions 6 and 8, possibly ABI-related,
prevent a safe co-mingling of foundational library components.  For 32-bit builds especially, the version
of download should stay about the same until new compiler versions are available.

PR #650, opened 2019-10-06
Revert "Adding builtin function for SCOPE_VARNAME, and a test-script for it."

GillesDuvert:
Reverts gnudatalanguage/gdl#626

Reason: error when restoring variables in a save file.
Valgrind says:
```
Internal error: !NULL destructor called.
Save your work and restart GDL (GDL is still functional, but !NULL will not work anymore).
Please report at https://github.com/gnudatalanguage/gdl/issues
==16665== Invalid free() / delete / delete[] / realloc()
==16665==    at 0x4835DD0: operator delete(void*) (vg_replace_malloc.c:586)
==16665==    by 0x11D761D: BaseGDL::~BaseGDL() (basegdl.cpp:28)
==16665==    by 0x1257515: DSubUD::ReplaceExistingCommonVar(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, BaseGDL*) (dpro.hpp:459)
==16665==    by 0x1A290B3: lib::restoreNormalVariable(EnvT*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, BaseGDL*) (saverestore.cpp:1294)
==16665==    by 0x1A2B847: lib::gdl_restore(EnvT*) (saverestore.cpp:1817)
```

This is due to the NULLGDL in Dcommon::AddVar().
Probably a patch in lib::restore() will suffice, but this is not on my shortlist, so I revert.

PR #656, opened 2019-10-21
AGAIN: Adding builtin function for SCOPE_VARNAME, and a test-script for it.

GillesDuvert:
Resubmission of "Adding builtin function for SCOPE_VARNAME, and a test-script for it.", originally a PR by T.Hillberg (#626) to avoid the problems identified in #650.

This version complies with the requirements without having to give values to otherwise undefined common parameters. Plus other improvements, as the finding of variables in commons was found to be buggy.


GillesDuvert:
As usual now, compil is broken with OSX , vritual machine problem?

PR #662, opened 2019-11-05
 create IS_BIGENDIAN macro at cmake stage for usage in sorting.cpp

maynardGK:
(presuming it will work) cmake's TEST_BIG_ENDIAN macro is included and a test is made.
In case it is big (motorola 68000 series, VAX, XBOX) then radix sort should work out.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,7 @@ include(CheckCSourceRuns)
 include(CheckCXXSourceCompiles)
 include(FindPkgConfig)
 include(FindPackageHandleStandardArgs)
+include(TestBigEndian)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR}/CMakeModules)
 
@@ -149,6 +150,9 @@ set(GDL_DATA_DIR "/share/gnudatalanguage" CACHE PATH "GDL: data directory relati
 if(${CMAKE_SIZEOF_VOID_P} EQUAL 8)
     set(HAVE_64BIT_OS 1)
 endif(${CMAKE_SIZEOF_VOID_P} EQUAL 8)
+
+# check for endianness, and set macro in case it is big.
+TEST_BIG_ENDIAN(IS_BIGENDIAN)
 
 #### check headers and libraries
 
--- a/config.h.cmake
+++ b/config.h.cmake
@@ -57,6 +57,7 @@
 #ifndef HAVE_EXT_STDIO_FILEBUF_H
 #cmakedefine HAVE_EXT_STDIO_FILEBUF_H 1
 #endif
+#cmakedefine IS_BIGENDIAN
 #cmakedefine OLD_DARWIN
 #cmakedefine PYTHON_MODULE 1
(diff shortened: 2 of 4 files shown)


GillesDuvert:
src/dSFMT/dSFMT.h  has a serie of #defines to test endianess ( which is important here since I presume mersenne random numbers wont be random at all if endianess is wrong ):
```
#if !defined(DSFMT_BIG_ENDIAN)
#if defined(__BYTE_ORDER) && defined(__BIG_ENDIAN)
#if __BYTE_ORDER == __BIG_ENDIAN
#define DSFMT_BIG_ENDIAN 1
#endif
#elif defined(_BYTE_ORDER) && defined(_BIG_ENDIAN)
#if _BYTE_ORDER == _BIG_ENDIAN
#define DSFMT_BIG_ENDIAN 1
#endif
#elif defined(__BYTE_ORDER__) && defined(__BIG_ENDIAN__)
#if __BYTE_ORDER__ == __BIG_ENDIAN__
#define DSFMT_BIG_ENDIAN 1
#endif
#elif defined(BYTE_ORDER) && defined(BIG_ENDIAN)
#if BYTE_ORDER == BIG_ENDIAN
#define DSFMT_BIG_ENDIAN 1
#endif
#elif defined(__BIG_ENDIAN) || defined(_BIG_ENDIAN) ||                         \
    defined(__BIG_ENDIAN__) || defined(BIG_ENDIAN)
#define DSFMT_BIG_ENDIAN 1
#endif
#endif

```
If these definitions are OK,[ this Apple reference ]( https://developer.apple.com/library/archive/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html) specifically suggests using these macros is more portable than relying on the execution of a (CMake) code snippet.
having an equivalent 'GDL_BIG_ENDIAN' computed that way would then be more efficient IMHO, no?

maynardGK:
I just took what was available ...  - as long as it stayed undefined it would work for now.
I listed the macros in my GCC :
echo "" | gcc -E -dM -x c - | sort
````
__MSYS__ 1
__NO_INLINE__ 1
__ORDER_BIG_ENDIAN__ 4321
__ORDER_LITTLE_ENDIAN__ 1234
__ORDER_PDP_ENDIAN__ 3412
#define __BYTE_ORDER__ __ORDER_LITTLE_ENDIAN__
````

I've used these macros in my update to #654

**Comment by slayoo** [comment]
the failing test_deriv was commented out in the travis runs - closing and reopening to re-check

**Comment by slayoo** [comment]
sorry, the test_deriv.pro is still disabled on master. merging this, re-enabling test teriv, and let's see...

**Comment by alaingdl** [comment]
OK, I succeed today to test the same CMake test
```
include (TestBigEndian)
TEST_BIG_ENDIAN(IS_BIG_ENDIAN)
```
then, yes, go ahead !

maynardGK:
IS_BIGENDIAN from should work, but another solution to the endianness is effected in the sorting.cpp of #654 that may be preferable,  In that, the test is done on the pre-existing macros provided by compilers.  Also in that PR is the solution to the use of posix shells on windows machines.

PR #663, opened 2019-11-07
cleaning TEST_HASH & TEST_PLOT_RANGES

alaingdl:
1/ removing few extra lines in TEST_HASH which result in problem for TEST_RESOLVE_ROUTINES
(+ few internal cleaning in TEST_HASH)

2/ When run in NULL device, TEST_PLOT_RANGES had trouble. We now check the status of the device


alaingdl:
the problem on OSX is not related to GDL

slayoo:
reported on homebrew issue tracker: https://github.com/Homebrew/brew/issues/6716

alaingdl:
thanks !

I am ready for a 1.0.0-rc1 :)

The only remaining problem is : test_fft_leak.pro (useful ??)

I will report ASAP a problem with std::badAlloc but I still have to clarify the issue

slayoo:
the homebrew problem is now fixed, but we still have the SORT problem! (and test_sort does fail on OSX + test_deriv is disabled)

PR #667, opened 2019-11-15
Update README.md

alexmyczko:
$ ldd /usr/bin/gdl |grep -i proj
        libproj.so.15 => /usr/lib/x86_64-linux-gnu/libproj.so.15 (0x00007f39fc411000)
$ ldd /usr/bin/gdl |grep -i glpk
        libglpk.so.40 => /usr/lib/x86_64-linux-gnu/libglpk.so.40 (0x00007f998cf48000)

$ gdl -V
% WARNING: your version of the GraphicsMagick library will truncate images to 16 bits per pixel
GDL - GNU Data Language, Version 0.9.9


slayoo:
thank you!

PR #683, opened 2019-12-22
Still work on projections

GillesDuvert:


GillesDuvert:
@slayoo  now that the projections work, a 1.0rc2 release would be great!

PR #684, opened 2019-12-23
Solves segfaults

GillesDuvert:


slayoo:
merging as the only failure on travis is for Intel Compiler which simply stopped being supported (see https://github.com/nemequ/icc-travis/issues/15)

Issue #693, opened 2020-01-06
PTR_VALID() has issues.


GillesDuvert:
as found in #425 PTR_VALID() produces incorrect results.
A rapid inquiry shows that PTR_VALID is not at all in line with current (say, 8.7) help and has thus been completely rewritten. PR is on its way.

GillesDuvert:
see #694

PR #694, opened 2020-01-06
Solves #693

GillesDuvert:


maynardGK:
Verify in IDL that anything can be passed to PTR_VALID.  http://www.harrisgeospatial.com/docs/PTR_VALID.html
indicates pointers or in cast of /CAST, integers.  Structures and strings are not mentioned; wording in the doc indicates in fact that passing anything other than pointer type may as well be an error - unless /CAST is specified.

With this difference, and other than returning ordinary integers rather than unsigned integers in the case of /GET_HEAP, the previous version is adequate - not "Completely" wrong.

GillesDuvert:
The current implementation of  ptr_valid should reproduce the original behaviour correctly. Of course reporting an issue is always possible. "completely wrong" was an overstatement due to a passing access of frustration and is now edited out.

PR #703, opened 2020-01-20
Edit a few Tests

maynardGK:
Some light clean-up for test_list, test_hash reflecting recent issues.


alaingdl:
also a problem with python & numpy

PR #732, opened 2020-03-06
Hsi followup after hash 719

GillesDuvert:
A lot of improvements mainly in save/restore, now the hsi/hessi objects are correctly saved ad restored, including the compatibility with IDL savefiles.
Also, correct treatment of heap indexes, variables and help,/heap and help,/obj
plus a few extraneous things.


GillesDuvert:
Hm... the discrepancy between IDL_CONTAINER and GDL_CONTAINER is the cause of the test_obj_valid failure.
Will try to do better.

GillesDuvert:
wille be splitted in smaller parts. stay tuned.

PR #738, opened 2020-03-31
better way to detect UdUnits2; few cleaning in CMakeLists.txt

alaingdl:
* I had some trouble for detecting udunits2 on some OS, because the files .h may be now under /usr/include/udunits2 which is not managed by current FINDudunits.cmake (to be removed later)

* this change is included in CMakeLists.txt (and I change variables related to UdUnits2 into "2")

* I also made some cosmetic changes in CMakeLists.txt 

* I activated by default GLPK, GRIB, UDUNITS2 in CMakeLists.txt 

* GDL_STATUS has now a more mnemonic keyword, /missing

All these changes have been tested on Debian 10, Ubuntu 14.04 & 16.04, Fedora 28 & RH 7.
No test up to now on OSX :(( sorry


alaingdl:
Please note the option to de-activate UdUnits2 is now -DUDUNITS2=off

I activiated by default (and tested) GLPK, UDUNITS2 & GRIB, may be some related tests are missing :)

PR #751, opened 2020-05-11
trying to manage TERMINAL_SIZE(cols [,lines])

alaingdl:
* moving TERMINAL_SIZE() and GET_KBRD() from basic_fun.cpp in terminfo.cpp
* trying to add management of Lines and Columns in TERMINAL_SIZE(cols [,lines]). 
May not work outside Linux+Readline (TBC)

TBD: managing all the consequences of that in HELP procedure (I began for HELP, /recall)

I expect some side effects on OSX with EditLine & maybe on Window

### DIFF
--- a/src/basic_fun.cpp
+++ b/src/basic_fun.cpp
@@ -35,10 +35,8 @@
 
 #include "includefirst.hpp"
 
-// get_kbrd patch
-// http://sourceforge.net/forum/forum.php?thread_id=3292183&forum_id=338691
 #ifndef _WIN32
-#include <termios.h> 
+//#include <termios.h> 
 #include <unistd.h> 
 #endif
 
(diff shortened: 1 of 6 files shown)


alaingdl:
@maynardGK 
sorry one time again I broke something on the Win side (bad message by appveyor which is not successful :(
Could you help me please ? Just because I moved few code from _basic_fun.cpp_ in _terminfo.cpp_

maynardGK:
The Get_KBRD() function doesn't compile when the readline stuff is included, I removed the readline aspects of get_kbrd() and retained all else except for gdlhelp stuff. #753 (with gdlhelp stuff added) should then be a good baseline to get the features into get_kbrd() that are missing.

PR #753, opened 2020-05-11
Repair file.cpp for gcc 9/win32; gdlpython compile error.

maynardGK:
gdlpython changes addresses issue #742 but there remains a warning for the call to import_array()
file.cpp change assures that an S_ISLNK is defined if it hasn't been already, as is case for GCC 9

PR #755, opened 2020-05-12
merge changes by Maynard and me

alaingdl:
unclear why this is a new PR :(

### DIFF
--- a/src/basic_fun.cpp
+++ b/src/basic_fun.cpp
@@ -1,4 +1,4 @@
-/***************************************************************************
+ /***************************************************************************
                           basic_fun.cpp  -  basic GDL library function
                              -------------------
     begin                : July 22 2002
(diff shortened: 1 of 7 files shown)


alaingdl:
the problem within MSwin remained the same but the way we try to correct was a mess (you and me loose time). Let's try to merge then to solve it later. Sorry @maynardGK

maynardGK:
When you merged your stuff onto #753 it un-did what I did for terminfo.  So #756 takes that result and branches, adding a _WIN32 version of get_kbrd() and removing all "#ifndef _WIN32" references in the code that brings in readline.h.

alaingdl:
Sorry for that, I follow some instructions provided by the system. At the end I was lost :(

alaingdl:
hon ca I do ?

PR #759, opened 2020-05-15
Repair for version >= 3 of numpy.

maynardGK:


maynardGK:
@slayoo I withdraw this simple change, which failed spectacularly.

Issue #757, opened 2020-05-15
With clang, the call to import_array() does not produce a warning, but an error:


slayoo:
With clang, the call to import_array() does not produces a warning, but an error:

```
[ 22%] Building CXX object src/CMakeFiles/gdl.dir/convol.cpp.o
cd /usr/ports/science/gnudatalanguage/work/.build/src && /usr/local/libexec/ccache/c++  -DHAVE_CONFIG_H -DWXUSINGDLL -D_FILE_OFFSET_BITS=64 -D_THREAD_SAFE -D__WXGTK__ -I/usr/ports/science/gnudatalanguage/work/gdl-1.0.0-rc.2/src -I/usr/local/include -I/usr/local/include/ImageMagick-6 -I/usr/local/lib/wx/include/gtk3-unicode-3.0 -I/usr/local/include/wx-3.0 -I/usr/local/mpi/openmpi/include -I/usr/local/include/python3.7m -I/usr/local/lib/python3.7/site-packages/numpy/core/include -I/usr/local/include/eigen3 -I/usr/ports/science/gnudatalanguage/work/gdl-1.0.0-rc.2 -I/usr/ports/science/gnudatalanguage/work/gdl-1.0.0-rc.2/src/antlr -I/usr/ports/science/gnudatalanguage/work/.build -O2 -pipe -fstack-protector-strong -fno-strict-aliasing  -std=gnu++11 -DACCEPT_USE_OF_DEPRECATED_PROJ_API_H=1 -I/usr/local/include/ImageMagick-6 -DMAGICKCORE_HDRI_ENABLE=0 -DMAGICKCORE_QUANTUM_DEPTH=16 -O2 -pipe -fstack-protector-strong -fno-strict-aliasing  -std=gnu++11 -DACCEPT_USE_OF_DEPRECATED_PROJ_API_H=1 -fPIE   -std=gnu++11 -o CMakeFiles/gdl.dir/convol.cpp.o -c /usr/ports/science/gnudatalanguage/work/gdl-1.0.0-rc.2/src/convol.cpp
--- src/CMakeFiles/gdl.dir/datatypes.cpp.o ---
In file included from /usr/ports/science/gnudatalanguage/work/gdl-1.0.0-rc.2/src/datatypes.cpp:27:
[...]

PR #767, opened 2020-05-22
Updates for 1.0.0

GillesDuvert:
This PR collects the changes i had in my branch, mostly related to objects and save/restore, all suggested by the behaviour of GDL in handling the HSI suite of instrumental pipelines, taken as an example of a really professional work.


GillesDuvert:
OK lets go!

alaingdl:
I do not understand why this job only is not running to the end :(
(I restarted twice) Message unclear for me.

https://travis-ci.org/github/gnudatalanguage/gdl/jobs/690130447

GillesDuvert:
strange as antl-compile of the .g files pass here on my machine (admitedly with warnings). I believe this is a spurious event.

alaingdl:
all the other ones are OK ... seems that the test trigered, but I don't understand why (and all other do the same)

This is a huge PR, on what topic(s) we should be especially cautious ?

GillesDuvert:
69e40cc (that removed 20 lines of ifdef) may have side effects on BSD ? see #757
Please try on some non-trivial instrument pipelines, if you have.

maynardGK:
Notice the extreme reduction in failed tests (to 4!) for 64-bit appveyor build.

slayoo:
> strange as antl-compile of the .g files pass here on my machine (admitedly with warnings). I believe this is a spurious event.

```
$ if [ -v ANTLR ]; then git diff -G '(^[ANTLR 2.7.7 (])' | tee diff.log; fi

diff --git a/src/GDLInterpreter.hpp b/src/GDLInterpreter.hpp

index 1ccb4ca..ceca37c 100644

--- a/src/GDLInterpreter.hpp

+++ b/src/GDLInterpreter.hpp

@@ -3,7 +3,7 @@

 

 #include <antlr/config.hpp>

 #include "GDLInterpreterTokenTypes.hpp"

-/* $ANTLR 2.7.7 (2006-11-01): "gdlc.i.g" -> "GDLInterpreter.hpp"$ */

+/* $ANTLR 2.7.7 (20160104): "gdlc.i.g" -> "GDLInterpreter.hpp"$ */

 #include <antlr/TreeParser.hpp>

 

 

@@ -23,7 +23,6 @@

 #include "accessdesc.hpp"

 #include "initsysvar.hpp"

 #include "gdljournal.hpp"

-#include "nullgdl.hpp"

 

 //class ProgNode;

 //typedef ProgNode* ProgNodeP;

@@ -248,7 +247,7 @@ public:

     {

         BaseGDL* del = (*it).second.get();

         objHeap.erase( id); 

-        if (!NullGDL::IsNULLorNullGDL(del)) delete del;

+        if (!NullGDL::IsNULLorNullGDL(del)) delete del; //avoid destroying !NULL

     }

     static void FreeObjHeap( DObj id)

     {

@@ -718,6 +717,7 @@ std::cout << add << " + <ObjHeapVar" << id << ">" << std::endl;

         }

 // The counters are reset for easier human readability.

        heapIx = 1;

+       objHeapIx = 1;

     }

 

     // name of data

The command "if [ -v ANTLR ]; then git diff -G '(^[ANTLR 2.7.7 (])' | tee diff.log; fi" exited with 0.

$ if [ -v ANTLR -a `cat diff.log | wc -l` -gt 0 ]; then travis_terminate 1; fi;
```

This is not a compilation thing - it's an error reported by diff-based line in .travis.yml meant to point out that the generated files do not match the ones in the repo - likely .g files need an update in the PR. This check is run only when env var ANTLR is set, and this is done only for one Travis job (https://github.com/gnudatalanguage/gdl/blob/master/.travis.yml#L9) - that's why only one test failed.

GillesDuvert:
@slayoo very clever, will see to it.

PR #774, opened 2020-05-29
WIP: re-insert some tests that should work ok with Appveyor builds.

maynardGK:
This should be a minimal change to Appveyor.yml, further attempts to change it (#766 ) have done very poorly!

Issue #771, opened 2020-05-26
wrong calculations when array size is bigger than 100.000 elements on Debian


brandy125:
I noticed very strange behaviour if the array sizes get bigger than 100000 elements:

GDL - GNU Data Language, Version 1.0.0-rc.2 git

GDL> c=dindgen(100000) & print,max(c)
       63749.000

With an array of maximum 100.000 elements I get correct results:
GDL> c=dindgen(100000-1) & print,max(c)
       99998.000

This is on a debian release.
On my suse release I get correct results:
GDL> c=dindgen(100001) & print,max(c)
       100000.00

Can anybody help me?

alaingdl:
Oups ... thanks for this important bug report ...

I can *partially* reproduce it on a debian machine, with different output
```
GDL> c=dindgen(100000) & print,max(c)
       83319.000
```
but not on my old U16.04

could you say which computer is using OpenMP, or not ? And which version of Gcc ??

Which tests fail ? (mack check)

Thanks

alaingdl:
the problem is in MAX() because when I plot DINGEN() or any other INDGEN() I have a full slope to the end.

Why _test\_minmax.pro_ is crashing on all the computers I am testing with current GDL ?! But OK with IDL ?!

alaingdl:
I did not see a good test on MAX() in the testsuite :((

slayoo:
```
GDL> c=dindgen(100001) & print,max(c)
       74999.000
```
(0.9.9 from GDL Debian package)

alaingdl:
Very few machines in my pool have the problem.

On a Debian 10 (buster) one, bad till 0.9.9 vanilla then recompiling 0.9.8 vanilla (26 March 2018)
and a TGZ (gdl-0.9.8svn180426) I have on 26 April 2018. 0.9.8 vanilla is OK, 26 April is KO.

I suspect we have to look in the history of minmax_include.cpp https://github.com/gnudatalanguage/gdl/commits/master/src/minmax_include.cpp introduced on April 5, 2018 ...

Up to now no Ubuntu nor OSX affected :(

alaingdl:
OK, pb is clarify : pb with OpenMP on such debian : cmake .. -DOPENMP=off 
goes to good computations.

gcc version 8.3.0 (Debian 8.3.0-6) 
libgomp1:amd64

compiling with clang 7 give the same kind of problem

alaingdl:
[...]

PR #776, opened 2020-06-03
going ahead for serious issue #771; solving #774

alaingdl:
* Maybe it is not the end of the story for #771 but this small change is enough for all my tests. Just we need to rework _test\_minmax.pro_

* Should solve #774 and a related test (coming from _test\_minmax.pro_) is now included in _test\_total.pro_


slayoo:
would this warrant a new release?
rc3? (as per recent comments from @GillesDuvert)

alaingdl:
*rc3* is ok for me now.

I have few changes, and need to push a revisited _minimum\_script4gdl.sh_ and we need to have a serious look in MINMAX code, but nothing to block such rc3.

PR #789, opened 2020-06-18
refactors around test automation

slayoo:
Introducing GDLDEV mode for CMake.

In short, ```cmake -DGDLDEV=ON``` makes ```make check``` execute all files that match testsuite/test_*.pro pattern. The default behaviour is to take the list of tests from testsuite/LIST file.


slayoo:
@acoulais @GillesDuvert @maynardGK @olebole 
Does it sound like a solution?

With this patch, by default only the tests that are listed in ```testsuite/LIST``` file are run by "make check", and the LIST file includes those tests that currently pass on Travis. This is the default that packagers would likely use, and any failures will mean that something fails that works on Travis on OSX and Linux - worth reporting, and worth fixing before releasing the package.

Furthermore, this patch introduces GDLDEV option to cmake, which replaces the list of tests with a list of all files matching ```testsuite/test_*.pro``` pattern. Consequently, doing ```cmake -DGDLDEV``` enables all tests that were disabled for Travis, but also will remind us of all the tests that are there but were not listed in Makefile.am (now LIST).

slayoo:
Well, let's do it, release rc3, and revert in case of negative feedback...

PR #796, opened 2020-06-21
fixing 77 exit code handling in launchtest.c

slayoo:


slayoo:
Looks good:

image

alaingdl:
that is a very good news ! thanks !

maynardGK:
Launchtest recognizes a skip, but GDL is still exiting.
For those tests which need a special software component, couldn't we simply produce a return from the top of the test?  How does it help us to cause exit,status=77 ?
````
GDL> .run test_simplex
% Compiled module: TEST_SIMPLEX_IDL.
% Compiled module: TEST_SIMPLEX_GLPK.
% Compiled module: TEST_SIMPLEX.
GDL> test_simplex
% TEST_SIMPLEX: GDL was compiled without GLPK support
````

PR #806, opened 2020-06-29
minor patches

maynardGK:
: CMakelists.txt -> "\n" in final message.  
GDLTokenType.hpp: #undef NOP (allows HDF=ON to compile) line 9
 (I can't figure how to change .g file so that this is included)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1185,10 +1185,10 @@ message("")
 
 # do we have a Conda env. around ? see issues #780
 if ((DEFINED ENV{CONDA_PREFIX}) OR (DEFINED ENV{CONDA_EXE}))
-  message("=========================================================================="\n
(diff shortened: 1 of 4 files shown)


GillesDuvert:
@maynardGK it is not possible to modify GDLTokenTypes.hpp without reporting the change in gdlc.g . And here we have a problem as the compilation of gdlc.g by ANTLR 7 will not write this `#undef NOP` line. So it will go back anytime.
For some reason, hdf5 headers clash with GDLTokenTypes.hpp . I suggest putting `#UNDEF NOP` in hdf_fun.cpp, and this, only if inverting lines 29 and 30 of hdf_fun.cpp does not solve the problem.

GillesDuvert:
@maynardGK you confirm it passes OK on your side with #undef NOP in hdf_xxx (checks alone would not show it ?)

maynardGK:
Yes I compiled it on OpenSuse with HDF=ON

PR #828, opened 2020-08-25
WIP: Update apptests to segregate and track failing i686 cases

maynardGK:
Appveyor will fail more readily when compiling the i686 case than for x86.  Now the LIST modification is made 
separately dependant upon the platform, a distinction found in scripts/ps1/appveyor_skip_tests.ps1


maynardGK:
continued in #828

maynardGK:
OBE

Issue #611, opened 2019-07-09
gdl-0.9.9 against latest graphicsmagick-1.3.32


dirteat:
Hello there,

this compiles fine, but crashes as soon as gdl starts:

>gdl
gdl: magick/semaphore.c:601: LockSemaphoreInfo: Assertion `semaphore_info != (SemaphoreInfo *) NULL' failed.
Aborted (core dumped)

Any ideas on what's going on?
Thanks,
Cheers,
chris.

alaingdl:
Could you please give more information on your configuration ?
I am ok on Debian with GM 1.3.30,  u1404 & GM 1.3.18 
and OSX 10.14.4 & GM 1.3.31 

(osx, compiler flavor & version)

thanks

A.

dirteat:
Yes, sorry.
That's on Mageia 7, gcc version 8.3.1, GM version 1.3.32 (gdl worked fine with GM < 1.3.32), so I suspect that's the suspect here, especially in view of the error message. DuckingGo on the web suggests it is an issue with GM initialization, but I cannot tell if the bug is in GM or within GDL initialization of GM!?

alaingdl:
thanks
I received in private yesterday the exactly same bug report than you but for OSX 10.14.5 with unknown GM version and was not able to reproduce up to now

A.

michel-plugge:
I have the same issue on fedora and CentOS 6/7 with precompiled version of gdl. Using GraphicsMagick.x86_64 0:1.3.32-1.el7 / GraphicsMagick-c++-1.3.32-1.fc29.x86_64  fails, but after downgrading to  GraphicsMagick-1.3.31-2.el7.x86_64.rpm / GraphicsMagick-c++-1.3.30-3.fc29.x86_64 gdl runs fine.

Michael

darthoctopus:
I can confirm this on Archlinux — I get the error with `graphicsmagick-1.3.32-1`, which goes away when downgrading to `graphicsmagick-1.3.31-3`

MQMLab:
Similar, on OS 10.14.6. All seemed fine after sudo port install gnudatalanguage
After typing gdl, though:  
Assertion failed: (semaphore_info != (SemaphoreInfo *) NULL), function LockSemaphoreInfo, file magick/semaphore.c, line 601.
Abort trap: 6

griegner:
Same issue on OS 10.14.6:
[...]

PR #863, opened 2021-01-03
Make GDL compilable on appveyor

pjb7687:
1. The current appveyor build doesn't work because some old libraries are missing on MSYS2 repo.
 - This patch firstly downloads the latest versions info available from MSYS2 repo and use them to download libraries. Also this uses wxwidgets and plplot from MSYS2 repo - doesn't build them anymore.
2. This patch defines missing `UNICODE` macro in `includefirst.hpp`. With this change, GDL compiles well without patching wxWidgets.
3. Includes a patch for GraphicsMagick assertion error (#611). This happens because GraphicsMagick's class type `Image` was initialized before `START_MAGICK` was called.

https://github.com/gnudatalanguage/gdl/blob/d1e75dffaaa4844706a1c59a45dfed2859f44d3b/src/magick_cl.cpp#L61


maynardGK:
Readline 5 should not be needed - I've used readline 8 on the home system.  readline 7 presented an issue nobody figured out what it was.

Patches made to the plplot drivers were needed for a correct behavior of keyboard focus, especially when used with cursor.    Avoiding the UNICODE patch is fine, though.

To streamline the queue, until you've built a working artifact, the number of modes should be limited to 1.

pjb7687:
> Readline 5 should not be needed - I've used readline 8 on the home system. readline 7 presented an issue nobody figured out what it was.
> 

There is a problem with `termcap` library on Windows. I am trying out various versions of readlines in combination with a patched termcap.

> Patches made to the plplot drivers were needed for a correct behavior of keyboard focus, especially when used with cursor. Avoiding the UNICODE patch is fine, though.
> 

`UNICODE` patch is for wxwidgets. For plplot, I am looking for another solution w/o patching plplot. Currently the main problem that I have encountered is that the plplot comes with mingw64 is not using dyndrivers - currently it complains with the following message: `PLplot installation lacks the requested driver: wingdi`

> To streamline the queue, until you've built a working artifact, the number of modes should be limited to 1.
How should I do that?

maynardGK:
With readline, just going to V6 required regex, also.  In msys then to get this stuff I included the following:
````
curl -O -s $pream/mingw/${arch}/mingw-w64-${arch}-libiconv-1.16-1-any.pkg.tar.xz
curl -O -s $pream/mingw/${arch}/mingw-w64-${arch}-termcap-1.3.1-5-any.pkg.tar.xz
curl -O -s $pream/mingw/${arch}/mingw-w64-${arch}-libsystre-1.0.1-4-any.pkg.tar.xz
curl -O -s $pream/mingw/${arch}/mingw-w64-${arch}-libtre-git-r128.6fb7206-2-any.pkg.tar.xz
````
libiconv may already be there & unrelated.
Limit the cbuild modes:

````
configuration:
  - Release
  # - Debug
environment:
  matrix:
  - platform: mingw64630x8664
  - platform: mingw64630i686
`````

The gcc6 compiler should be adequate

windgdi is a new device in plplot with better lettering.  It is very much like wingcc, and our code can be used with either, depending on compilation,  when plplot is built with the patches applied,  To revert back to the old (still needs patching for cursor focus) wingcc, in the GDL cmake command set   -DUSE_WINGDI_NOT_WINGCC=OFF.

**Comment by pjb7687** [comment]
Hi,

1. I reverted back the script to use the latest readline which is v8. We can just include any dependency in the propsed script, without designating their version numbers. The main idea is to avoid using specific version numbers in the build script, because by the time these will be deprecated someday and eventually  make GDL fail to compile. Well that's actually what is happening now.. Currently the latest appveyor build of the master branch does not include an executable in the artifact. The proposed script at least successfully builds `gdl.exe` executable in the end...

2. Basically I agree all modules should use the same GCC. The latest libs from mingw64 project is using automated build script (e.g. https://github.com/msys2/MINGW-packages/runs/1652959631) which uses their latest GCC, which is currently GCC 10. The proposed script will try to compile everything with the latest GCC (i.e. GCC 10), just as like their build script.

3. Well I would just like to say, I just don't like the idea for us to 'patch' plplot, since this will eventually break the build when a new plplot is released, and will make us hard to manage the build... (otherwise we can just stick to an old 'patched' plplot... but I don't like this idea either). Other than, I have absolutely no objections using the new `wingdi` driver. The problem I have is that when I compiled GDL with the mingw64 version of Plplot, GDL (or plplot?) just can't find that driver. I will have a closer look perhaps in this weekend.. My holidays are over now and I only have limited timeframe at the moment.

4. Another major issue is that currently everything printed in stderr just vanishes, only I can see the output directed to stdout. I will also check this later...

**Comment by slayoo** [comment]
@pjb7687  @maynardGK big thanks for working on it!
Let me also ask if perhaps the https://github.com/gnudatalanguage/gdl/issues/615 could be addressed?

**Comment by pjb7687** [comment]
> @pjb7687 @maynardGK big thanks for working on it!
> Let me also ask if perhaps the #615 could be addressed?

Hi,

I appended ` || exit 1` for cmake and make command so that the script will exit with 1 when compile fails. I can't find the code to run tests, may I ask where could I find this?

**Comment by pjb7687** [comment]
Stdrr issue is fixed now. My bad, it wasn't related to readline but I just missed the 'magic sauce' for wx.

**Comment by slayoo** [comment]
Thanks!

These seem to be candidate locations for the test run:

https://github.com/gnudatalanguage/gdl/blob/c76a7ee168aed6b45649ecdff10e9f9f306f27e6/scripts/appveyor_gdl.msys#L90

https://github.com/gnudatalanguage/gdl/blob/c76a7ee168aed6b45649ecdff10e9f9f306f27e6/scripts/mingw-bldgdl.bat#L90

**Comment by pjb7687** [comment]
Done! Now it builds fine. Test also works well, with 8 failures though. Thats all for today - Epiphany (public holiday here). I will tackle plplot part later - perhaps in this weekend.

**Comment by slayoo** [comment]
Kudos!

**Comment by pjb7687** [comment]
Okay. The problem was simply mingw-w64 version of plplot didn't include wingcc nor wingdi drivers. I suggested a patch in their repo: https://github.com/msys2/MINGW-packages/pull/7625.

Now I confirm that nearly everything works. If no objections, I would like to merge this pull request.

**Comment by maynardGK** [comment]
The plot drivers need to be in the /bin directory where the program is run from.  
wingdi.dll will not execute properly because they scrambled the order of variables in the declaration,
and didn't unscramble them as I requested early on.  
 So at least wingcc needs to be copied for the artifact, and the build needs to use wingcc.  wingdi is then irrelevant.  
Then, BWToTop->GetHwnd() looks like it should work but I don't believe it will (but I can't test the artifact).
If plplot is built in appveyor, it will only take a minute, and we won't bring all the other drivers that the full plplot is built with.    A good chance, the c-code patch works on V5.15 (and up) identically to v5.13.

**Comment by pjb7687** [comment]
> The plot drivers need to be in the /bin directory where the program is run from.

The mingw-w64 build plplot doesn't use dyndrivers. They are statically linked to the dll.

> wingdi.dll will not execute properly because they scrambled the order of variables in the declaration,
> and didn't unscramble them as I requested early on.
> So at least wingcc needs to be copied for the artifact, and the build needs to use wingcc. wingdi is then irrelevant.
> Then, BWToTop->GetHwnd() looks like it should work but I don't believe it will (but I can't test the artifact).

No worries, I have already worked on wingdi driver and it works perfectly.
I will make another pull request for it - just not here. This is for making appveyor build to work.

> If plplot is built in appveyor, it will only take a minute, and we won't bring all the other drivers that the full plplot is built with. A good chance, the c-code patch works on V5.15 (and up) identically to v5.13.

I can see only several mb of differences of plplot dll between statically linked version and dyndriver version. Is it worth for us to build it every time?

**Comment by pjb7687** [comment]
Okay - so the only remaining thing seems that whether we build plplot or not. But since this seem to be easily adjusted whenever if needed, I merge this.

File: .ci/build_gdl_windows.msys [hunk 1]
```diff
@@ -0,0 +1,206 @@
...
+    
+    make -j2 || exit 1
+    
+    log "Installing GDL..."
+    make install || exit 1
+}
+
+function test_gdl {
+    cd ${ROOT_DIR}/build
+    make check # JP: why check not test? it's confusing..
```
slayoo, line 195: IIRC, ths is because we need to depend on other targets, i.e. launchtest and test_ce:
https://github.com/gnudatalanguage/gdl/blob/cd85af4ad4455bc351861e2f29f6e9b65d6fb7e7/testsuite/CMakeLists.txt#L95

pjb7687: Thanks!

maynardGK:
I suspect that the wingdi driver is by default not built for plplot (in cmake, no DPLD_wingdi=ON): 
````
$ bash -c "export PATH=.:$PATH && /d/f/github/gdlpjb/bld1452/bin/gdl"
  GDL - GNU Data Language, Version 1.0.0-rc.3 git

GDL> pwd
% Compiled module: PWD.
D:/f/github/gdlpjb
GDL> plot,findgen(100)
% PLplot installation lacks the requested driver: wingdi
% Execution halted at: $MAIN$
````
The wingdi driver was going to use other windows libraries that need comctl32, so in that 
branch you'll need the line in CMakelists.txt:
````
    endif(PCRE_FOUND)
    LINK_LIBRARIES(shlwapi comctl32)
````
Meanwhile it is looking good, if you built using wingcc then it should test ok.

pjb7687:
> I suspect that the wingdi driver is by default not built for plplot (in cmake, no DPLD_wingdi=ON):
> 
> ```
> $ bash -c "export PATH=.:$PATH && /d/f/github/gdlpjb/bld1452/bin/gdl"
>   GDL - GNU Data Language, Version 1.0.0-rc.3 git
> - For basic information type HELP,/INFO
> - Please report bugs, feature or help requests and patches at: https://github.com/gnudatalanguage/gdl
> 
> GDL> pwd
> % Compiled module: PWD.
> D:/f/github/gdlpjb
> GDL> plot,findgen(100)
> % PLplot installation lacks the requested driver: wingdi
> % Execution halted at: $MAIN$
> ```
> 
> The wingdi driver was going to use other windows libraries that need comctl32, so in that
> branch you'll need the line in CMakelists.txt:
> 
> ```
>     endif(PCRE_FOUND)
>     LINK_LIBRARIES(shlwapi comctl32)
> ```
> 
> Meanwhile it is looking good, if you built using wingcc then it should test ok.

I made another pull request rearding wingdi driver: https://github.com/gnudatalanguage/gdl/pull/864

Please test it instead. Thanks!

PR #864, opened 2021-01-10
better wingdi driver patch

pjb7687:
As mentioned in https://github.com/gnudatalanguage/gdl/pull/863#issuecomment-757424097


pjb7687:
For testing, you can use the plplot artifacts from here:
https://github.com/msys2/MINGW-packages/runs/1677180400

Just overwrite dlls before running GDL.

pjb7687:
I merge this since wingdi enabled plplot is available from msys2 repo.

Issue #602, opened 2019-06-25
make an installer for Windows version


GillesDuvert:
the title says all..

pjb7687:
I have no idea this would be still useful:
https://sourceforge.net/projects/gnudatalanguage-win32/files/NSIS%20script/

GillesDuvert:
+ include gdlde and Jupyter kernel?

pjb7687:
Added initial script: https://github.com/gnudatalanguage/gdl/pull/865

slayoo:
WOW!

GillesDuvert:
needs probably a bit of polishing as the gdl.exe must find external libraries files (plplot, plplot drivers, proj...).
Plus documentation (or, rather, advertising).
I leave open.

PR #865, opened 2021-01-10
added script to generate windows installer

pjb7687:


pjb7687:
Solving #602.

GillesDuvert:
Looks nice, installer works!
Two caveats, first, launching the installer is awkward because the .exe is not trusted by Microsoft (normal)
Second, !PATH is empty (try `plot,dist(3)` )

pjb7687:
It seems like when the installer launches GDL, it cannot find envvars properly. Could you try relaunching GDL and run the command again? And for plotting, this one is needed: https://github.com/gnudatalanguage/gdl/pull/864#issuecomment-757503007

pjb7687:
Seems like mine works well..
image

GillesDuvert:
Thanks, for me now a plot appears...
But then disappears instantly. Will the future installers contain the patched plplot?

pjb7687:
I think you are not using the patched `GDL.exe` from https://github.com/gnudatalanguage/gdl/pull/864. You can download the artifact there and replace `GDL.exe` for testing.
And yes, the patched plplot will be automatically included when MSYS2 repo is updated. Maybe after several days.

PR #866, opened 2021-01-10
Update osx image travis ci

pjb7687:
MacOS build fails since the plplot homebrew package no longer exists for osx older than Mojave (10.14). For backward compatibility, osx image is updated to ~the oldest image based on Mojave~ the image that supports at least XCode 11.3 due to https://github.com/Homebrew/homebrew-core/issues/52461.


pjb7687:
The `mini` builds works well, but the `full` build stalls since homebrew's `llvm` cannot find `libomp` (although it is already installed!). I do not have a mac so I cannot debug more, maybe anyone having a mac can help here?

GillesDuvert:
is the problem the same as in [this link ?](https://stackoverflow.com/questions/46414660/macos-cmake-and-openmp)

pjb7687:
> is the problem the same as in [this link ?](https://stackoverflow.com/questions/46414660/macos-cmake-and-openmp)

I am not sure, but I think we can try this out..

pjb7687:
...seems like it doesn't work.

alaingdl:
May I help you ?

Last weeks I spend some time on two OSX 10.14 machines to install & test current git GDL

What can I say ? 
* _minimal__script.sh_ is OK (a way to have simple options for plplot)
* basic compilation is OK (but I need a clean readline !)
* Using the default compiler, I received a large number of messages due to OpenMP (which was detected and accepted) and I can compile to the end only if I deactivate OpenMP
* Brew updated LLVM into version 11. Cmake complains and stop, saying clang cannot compile. I try few suggestions around, without success up to now :( 

(and yes, at the end, I can help you to have the display ON  remotely, need to make few changes in /etc/ssh/ssh_config & sshd_config )

A.

pjb7687:
Hi Alain, Yes this problem happens after LLVM is updated by homebrew (see `MINI` build worked well..). If you have a old Xcode, it seems like it needs to be updated to 11.3 otherwise it cannot compile: https://github.com/Homebrew/homebrew-core/issues/52461#issuecomment-632372338

I solved this by using Xcode 11.3 image, but now it cannot find openmp.

GillesDuvert:
See this page https://mac.r-project.org/openmp/

pjb7687:
The build works fine now. I haven't tested the binary, but I think it's always better to have something at least compilable than nothing. So I merge this... If the compiled binary doesn't work, then please open another issue.

Issue #841, opened 2020-10-17
WSET cannot switch to draw widget (windows problem only)


PaddyLeahy:
Under windows, the following code fails

tlb   = WIDGET_BASE(/column)
draw1 = WIDGET_draw(tlb,xsize=512,ysize=512,retain=2) 
WIDGET_CONTROL, tlb, /realize
WIDGET_CONTROL, draw1, get_value = plot1
WSET, plot1

Error message is 
% WSET: Window is closed and unavailable.

This code snippet works fine under linux, with Wxwidgets.

GillesDuvert:
Do you have a better success with "gdl --use-wx"?
Drawing in wxWindows depends on the version of plplot used, and on the plplot wxwidgets driver. Already tricky under linux, fingers crossed on Windows...

JWF9912:
> Do you have a better success with "gdl --use-wx"?
> Drawing in wxWindows depends on the version of plplot used, and on the plplot wxwidgets driver. Already tricky under linux, fingers crossed on Windows...

JWF9912:
This is still a problem as of the 14Feb21 build.  When the draw widget is realized it is assigned a window index but the WSET procedure will not accept the window index value.  Also, the system variable !D.WINDOW is not set to the window index of the draw widget.

If GDL --use-wx is used the draw widget window index is set correctly but GDL returns a warning message

![image](https://user-images.githubusercontent.com/78502834/107989063-6763a980-6f9f-11eb-9ea5-8cce73d01b31.png)

GillesDuvert:
this is completely solved in version 2 of gdl widgets (PR #882) .

PR #882, opened 2021-02-04
WIP Widgets version2, compilation tests

GillesDuvert:
only test this time


alaingdl:
Wouhaou, great

But sorry, could you repeat the simplest way (serie of CLI instructions) to download that, for example starting from zero
or from an already cloned master ? (same situation in fact)

thanks 

A.

Ps: I may also be able to test on one OSX where I already compiled GDL within Widgets

GillesDuvert:
@alaingdl , hello, I cannot figure what you want? ( and I'm surprised by all those new builds ).
Let me adress the small problems found during this first mega-compilation.

GillesDuvert:
retry

GillesDuvert:
excellent, with this scheme aborting the tests will still produce the artifacts.

GillesDuvert:
image

GillesDuvert:
Works OK on Windows.

pjb7687:
Works well!

pjb7687 (approved):

GillesDuvert:
image

src/gdlwidget.hpp
```diff
@@ -39,9 +39,6 @@
 #include <wx/popupwin.h>
 #include <wx/notebook.h>
 #include <wx/dcbuffer.h>
-#if __WXMSW__
-#include <wx/msw/wx.rc>  //should provide manifest, but is linker OK?
```
pjb7687: Does it require manifest? AFAIK you can use `windres`...

GillesDuvert: @pjb7687 the widgets have no Look & Feel on Windows, because somewhere in the build process, one shoudl provide a gdl.exe.manifest for the windows executable.
But I cannot find an example of how to do that, wxWidget documentation seems to believe everyone adds manifests just like they are breathing and do not need any explanation or example.
At least, #include <wx/msw/wx.rc> only was a failure.
Probably you can help?

pjb7687: Added 2 commits below... It should work now.

pjb7687:
I've recently fixed the macOS build, but this patch seems incompatible on macOS. Here you can see some test failures...
https://travis-ci.com/github/gnudatalanguage/gdl/jobs/482830812

src/gdlwidget.hpp
GillesDuvert: thanks a lot, the look and feel has improved!

maynardGK:
I have been attempting to incorporate the new GDL source as well as update my MSYS package:  I could run the CI-built distros in the previous BASH, but because of windows' upgrades or something I lost the ability to properly connect with the libraries I held to run, i.e.  the Q32 version of GraphicsMagick.

So that's where I am at the moment, and I notice again about the BSD/XDR package,
the basic distro provides an xdr,h which is different from the xdr.h given by the hdf4
package:
````
$ pacman -Qo /mingw64/include/xdr.h
/mingw64/include/xdr.h is owned by mingw-w64-x86_64-hdf4 4.2.15-2
$ cat /mingw64/include/xdr.h | head
/*      $NetBSD: xdr.h,v 1.19 2000/07/17 05:00:45 matt Exp $    */

/*
 * Copyright (c) 2009, Sun Microsystems, Inc.
 * All rights reserved.
 *
 * Redistribution and use in source and binary forms, with or without
 * modification, are permitted provided that the following conditions are met:
 * - Redistributions of source code must retain the above copyright notice,
 *   this list of conditions and the following disclaimer.

greg@i7Wx MSYS /d/sources/bsd-xdr-1.0.0
$ cat rpc/xdr.h | head

/*
 * Copyright (c) 2009, Sun Microsystems, Inc.
 * All rights reserved
 ````
````
41,45c41,42
< #ifndef __XDR_HEADER__
< #define __XDR_HEADER__
<
< #include "H4api_adpt.h"
<
---
> #ifndef _RPC_XDR_H
> #define _RPC_XDR_H
46a44,48
> #include <rpc/types.h>
>>
> #ifdef __cplusplus
> extern "C" {
> #endif
57,60c59,60
<  *    bool_t
<  *    xdrproc(xdrs, argresp)
<  *        XDR *xdrs;
<  *        <type> *argresp;
---
>  *        bool_t
>  *        xdrproc(XDR *xdrs, <type> *argresp)
````

GillesDuvert:
@maynardGK actually, installing the last artifact of this build on a windows 10, i cannot open images at all,  **so libmagick is not working**. I can open tiff 8 bit images however, this does not use graphicsmagick/imagemagick. But no 24-bit tiff images (the tests in testsuite/tiff), but this is true also for the linux version, the error is:
`% Parser syntax error: unexpected token: READ_TIFF`
meaning a NULL has been returned by whatever function is called when reading 24 bit Tiff images.

pjb7687:
> @maynardGK actually, installing the last artifact of this build on a windows 10, i cannot open images at all, **so libmagick is not working**. I can open tiff 8 bit images however, this does not use graphicsmagick/imagemagick. But no 24-bit tiff images (the tests in testsuite/tiff), but this is true also for the linux version, the error is:
> `% Parser syntax error: unexpected token: READ_TIFF`
> meaning a NULL has been returned by whatever function is called when reading 24 bit Tiff images.

This should be fixed now: https://github.com/gnudatalanguage/gdl/issues/883
If you still can't open any images please open the issue again.

pjb7687:
> I have been attempting to incorporate the new GDL source as well as update my MSYS package: I could run the CI-built distros in the previous BASH, but because of windows' upgrades or something I lost the ability to properly connect with the libraries I held to run, i.e. the Q32 version of GraphicsMagick.
> 
> So that's where I am at the moment, and I notice again about the BSD/XDR package,
> the basic distro provides an xdr,h which is different from the xdr.h given by the hdf4
> package:
> 
> ```
> $ pacman -Qo /mingw64/include/xdr.h
> /mingw64/include/xdr.h is owned by mingw-w64-x86_64-hdf4 4.2.15-2
> $ cat /mingw64/include/xdr.h | head
> /*      $NetBSD: xdr.h,v 1.19 2000/07/17 05:00:45 matt Exp $    */
> 
> /*
>  * Copyright (c) 2009, Sun Microsystems, Inc.
>  * All rights reserved.
>  *
>  * Redistribution and use in source and binary forms, with or without
>  * modification, are permitted provided that the following conditions are met:
>  * - Redistributions of source code must retain the above copyright notice,
>  *   this list of conditions and the following disclaimer.
> 
> greg@i7Wx MSYS /d/sources/bsd-xdr-1.0.0
> $ cat rpc/xdr.h | head
> 
> /*
>  * Copyright (c) 2009, Sun Microsystems, Inc.
>  * All rights reserved
> ```
> 
> ```
> 41,45c41,42
> < #ifndef __XDR_HEADER__
> < #define __XDR_HEADER__
> <
> < #include "H4api_adpt.h"
> <
> ---
> > #ifndef _RPC_XDR_H
> > #define _RPC_XDR_H
> 46a44,48
> > #include <rpc/types.h>
> >>
> > #ifdef __cplusplus
> > extern "C" {
> > #endif
> 57,60c59,60
> <  *    bool_t
> <  *    xdrproc(xdrs, argresp)
> <  *        XDR *xdrs;
> <  *        <type> *argresp;
> ---
> >  *        bool_t
> >  *        xdrproc(XDR *xdrs, <type> *argresp)
> ```

I don't think the difference in the XDR is a big problem. HDF4 just has its own implementation of XDR, which is dependent on other internal HDF4 headers, and this can be considered as something independent from the normal XDR library that we use (i.e. BSD-XDR). Also, the header location is different (one is in `include`, the other one is in `include/rpc`) and they do not overwrite each other.

Regarding GraphicsMagick, then I think it would be nice to build a Q24 or Q32 version in the CI script. But I have a question: Isn't Q16~Q24~ enough for most cases?

pjb7687:
> I've recently fixed the macOS build, but this patch seems incompatible on macOS. Here you can see some test failures...
> https://travis-ci.com/github/gnudatalanguage/gdl/jobs/482830812

Interesting, I've tried compiling it on macOS Catalina, it worked very well. 100% test passed.
I uploaded a compiled binary so that you can test it: https://filebin.net/aakbf0y84sq1ybr6/gdl-full-debug-catalina.zip?t=voqy7v6y

FYI the binary was compiled on Github Actions: 
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md

maynardGK:
> Regarding GraphicsMagick, then I think it would be nice to build a Q24 or Q32 version in the CI script. But I have a question: Isn't Q24 enough for most cases?

The goal I had was for a Q16 version, since, I beliieve, that is the level most distros have.  The MSYS2/mingw64 version was Q8 last time I checked ... if that were changed to Q16 the world would be a better place.

pjb7687:
> > Regarding GraphicsMagick, then I think it would be nice to build a Q24 or Q32 version in the CI script. But I have a question: Isn't Q24 enough for most cases?
> 
> The goal I had was for a Q16 version, since, I beliieve, that is the level most distros have. The MSYS2/mingw64 version was Q8 last time I checked ... if that were changed to Q16 the world would be a better place.

Yes I meant Q16. Then I will make another PR to build Q16 GraphicsMagick in the CI script.

maynardGK:
> 
> I don't think the difference in the XDR is a big problem. HDF4 just has its own implementation of XDR, which is dependent on other internal HDF4 headers, and this can be considered as something independent from the normal XDR library that we use (i.e. BSD-XDR). Also, the header location is different (one is in `include`, the other one is in `include/rpc`) and they do not overwrite each other.
> 

OK but  we need to note that lib/libxdr.dll.a is named the same for both, so it may be ok if used by bsd-xdr but probably a few more entry points when used for libxdr,dll (.vs. mgwxdr-0.dll).

pjb7687:
On Windows, the name of the plotting device is still 'X' and it seems like this needs to be changed to 'WIN'.
I have a question - after we merge this, libwxwidgets becomes one of the mandatory packages since this will replace the current devicex or devicewin, is this correct?

GillesDuvert:
> On Windows, the name of the plotting device is still 'X' and it seems like this needs to be changed to 'WIN'.
> I have a question - after we merge this, libwxwidgets becomes one of the mandatory packages since this will replace the current devicex or devicewin, is this correct?

Yes thanks for the opportunity to expose my plans:

1. I suggest to have wxWidgets a 'main' dependency. Removing the need of X11 for unix/linux/MacOS, as well as xwin and wingcc/wingdi . This by default. Behavior of plots, either in a WIDGET_DRAW or in a WINDOW, will be identical.
2. provided wxWidgets library exist, the (only) device thus existing will be 'X'. (can be if you prefer 'MAC' on macs and 'WIN' on windows, buts passing as 'X' is the safest solution IMHO since the behaviour of GDL is unique?). 
3. if the wxWidgets library does not exist, then it is a compilation failure unless cmake enforces X11 (-DX11=ON) (or -DWINGCC or equivalent for windows) . We get then the 'old' behaviour: no widgets, plot device is 'X' or 'WIN'.
4. If wxWidgets is present, and cmake enforces X11 (-DX11=ON)[or -DWINGCC=ON) , then by default behaviour is as in 2. unless the option `--no-use-wx` is used in calling gdl. With `--no-use-wx` the widgets will use wx, the plot window will use wxin or wingdi, the device will be 'X' or 'WIN' respectively, and the wxwidgets plot device will be available as 'MAC' (as of today).

pjb7687:
> > On Windows, the name of the plotting device is still 'X' and it seems like this needs to be changed to 'WIN'.
> > I have a question - after we merge this, libwxwidgets becomes one of the mandatory packages since this will replace the current devicex or devicewin, is this correct?
> 
> Yes thanks for the opportunity to expose my plans:
> 
> 1. I suggest to have wxWidgets a 'main' dependency. Removing the need of X11 for unix/linux/MacOS, as well as xwin and wingcc/wingdi . This by default. Behavior of plots, either in a WIDGET_DRAW or in a WINDOW, will be identical.
> 2. provided wxWidgets library exist, the (only) device thus existing will be 'X'. (can be if you prefer 'MAC' on macs and 'WIN' on windows, buts passing as 'X' is the safest solution IMHO since the behaviour of GDL is unique?).
> 3. if the wxWidgets library does not exist, then it is a compilation failure unless cmake enforces X11 (-DX11=ON) (or -DWINGCC or equivalent for windows) . We get then the 'old' behaviour: no widgets, plot device is 'X' or 'WIN'.
> 4. If wxWidgets is present, and cmake enforces X11 (-DX11=ON)[or -DWINGCC=ON) , then by default behaviour is as in 2. unless the option `--no-use-wx` is used in calling gdl. With `--no-use-wx` the widgets will use wx, the plot window will use wxin or wingdi, the device will be 'X' or 'WIN' respectively, and the wxwidgets plot device will be available as 'MAC' (as of today).

Thanks! I would say, additionally we might just want to deprecate wingcc/wingdi devices on Windows. Actually, these devices are too old, for example, they do not inherit GraphicsMultiDevice class and produce weird errors e.g. #841.

Regarding the name of the device, using 'WIN' as driver name on Windows is just IDL behavior (https://www.l3harrisgeospatial.com/docs/supported_devices.html). For this, we can just rename the new driver name on Windows using ifdef.

On Linux/macOS it is up to others' opinions, but in principle, I think it is better to deprecate X device too and focus on improving the wxWidgets-based codebase. Because in this case we only have one multi-platform single device for plotting/widgets, and this will be much easier to manage.

GillesDuvert:
close,reopen

GillesDuvert:
work in progress: works definitely on linux & windows (slow on windows) . Severe problem on Mac OSX (GDL stuck in eventloop). Probably signals a (not unexpected, this!)  bad eventloop construction/use in GDL, solving it may certainly improve GDL efficiency.

GillesDuvert:
normally this version has the same behaviour on OSX, Windows and linux, does not need X11 or wingcc, is fast etc.
Provided the installed version of plplot has the wxwidgets driver, everything should be ok.
If problems, will insert the wxwidgets plplot driver in gdl's code and we get rid of all thise problems once and 4 all.

GillesDuvert:
OK, passes everywhere except for some tests that are plainly wrong and the 'mini' dep that evidently fails because the compilation conditions have changed (MINI must now be compiled with -DX11=ON or -DINTERACTIVE_GRAPHICS=OFF).
And, python tests fail for some silly reason i do not care at all.
I suggest to remove all the builds except the full macsox and the full linux + codecov, this will avoid many headaches. All other builds have no interest at all, the one and only GDL is with all options and one unified graphics through wxwidgets.

alaingdl:
HI. I am not sure I fully understand you. 
What will happen on computing nodes on HPC where we have no wx and no X11 ?

GillesDuvert:
> HI. I am not sure I fully understand you.
> What will happen on computing nodes on HPC where we have no wx and no X11 ?

provided -DINTERACTIVE_GRAPHICS=OFF is present in the cmake command, everything will be fine.

maynardGK (commented):
I can only build, not run, with the mingw/msys2 gcc 10 on win10: runtime doesn't find
the $PATH where shared libraries are kept.
I can download and run the github builds (shareds gotten from directory holding gdl.exe).
windows graphics with --use-wx still gives the error pop-up box with message
````
 ../wxWidgets-3.5.01/src/msw/dc.cpp(550) assert "hrgn region" failed
 in SetClippingHrgn(); invalid clipping region
````

GillesDuvert:
(post commit) works:
image

PR #886, opened 2021-02-08
WIP: Link GDL to native python interpreter on Windows

pjb7687:
On Windows, GDL cannot find Numpy on the native Python installation (which is, the one installed using the official installer).
This patch makes GDL linked against the native Python on the system, which should resolve the problem.


pjb7687:
Close, fixed as a part of https://github.com/gnudatalanguage/gdl/pull/971

PR #891, opened 2021-02-15
Build GraphicsMagick-Q16 in CI script

pjb7687:


GillesDuvert:
OK for me.
One must handle the return of reading failure on 24-bit tiffs:
```
 GDL> a=read_tiff("24bit_color.tif",geo=b)
% Parser syntax error: unexpected token: READ_TIFF
```
which is apparently not trapped in GDL.

pjb7687:
> OK for me.
> One must handle the return of reading failure on 24-bit tiffs:
> 
> ```
>  GDL> a=read_tiff("24bit_color.tif",geo=b)
> % Parser syntax error: unexpected token: READ_TIFF
> ```
> 
> which is apparently not trapped in GDL.

I found that it works on Windows. However TV doesn't work, seems like the current TV function cannot handle color images.
Also, it seems like TIFF file handling is done by libtiff, not GraphicsMagick.

PR #902, opened 2021-03-22
WIP: Use Github Actions to build Linux/macOS

pjb7687:


GillesDuvert:
@pjb7687 there needs to add "-DINTERACTIVE_GRAPHICS=OFF" when building MINI products.
(personnally I doubt building MINI has any usefulness now that all 3 target platforms have the same capabilities)

slayoo:
> (personnally I doubt building MINI has any usefulness now that all 3 target platforms have the same capabilities)

The idea of the MINI builds was to cater to those (incl. myself) that mostly use GDL in non-interactive non-Python non-X non-widgets setups and intend to build minimal GDL on platforms where they have little control over the environment (e.g., supercomputing nodes, etc)

Also, it proved to be useful to detect issues where some dependency was assumed within GDL without proper include-sequence or include-guard, etc - when all works fine with typical build, but e.g. things brake in strange manner when someone disables one dependency and a problem arises in different place (e.g. when disabling Eigen or X)

I would say it's useful to have such a minimal build.
@acoulais ?

GillesDuvert:
Dear @slayoo I agree that building GDL without interactivity has some interest, but I would reserve it to a special build request, not the continuous integration deployed on GH/Travis etc.
Further, MINI as of now has no utility. What you refer to is just GDL with everything, minus X11/widgets and obviously **PLUS MPI**. Which is not the case at the moment. (and would need MPI-related tests, absent at the moment).
I note that MINI wants plplot, and indeed supercomputing nodes may want to use graphics commands, even 'internally'. It is easy to design (I've started to do it) a 'Notebook' version of GDL that has no X11 or widgets, but produces nice svg plots in the Jupyter notebook.
Actually there are way too many #ifdef in the GDL code, as the basic product needs everything present (except, indeed, X11 and wxWidgets), and why not, as all the needed libraries are there everywhere? What is the point of GDL not reading a png file, be unable to load an http file, decode an xml file, understand a projection,  etc? Allowing people, or, worse, packagers, to build GDL without this or that dependency is just not possible!
Finally, MINI builds are a mess regarding the maintainance of the tests, and now that we are able to build the same GDL on every platform, with all dependencies, I would put more effort of the continuous integration facility in testing the 32 bits and little-endian issues than the confidential MINI uses.
Sorry to be so frank but I see there a too few people actually providing code and paches and tests to GDL, we cannot disperse ourselves on so many topics.

pjb7687:
Hi, I actually sort of agree with both of you, @slayoo and @GillesDuvert 

As @slayoo pointed out, it makes sense to build GDL without some libraries, at least X11. AFAIK the main problem in a cluster environment is usually X11, I am pretty sure it isn't installed (and won't be available) on many systems due to various reasons. Plotting still can be done without X11, e.g. using Postscript drivers.

But as @GillesDuvert mentioned, when I have a look at the list of dependencies of full GDL, I personally think there should be no problem including other libraries... e.g. I can't imagine any Linux distro for supercomputing nodes without Python. Regarding Eigen and X: (1) I think Eigen should be a hard dependency since it is a header-only library without any dependency, there should be no problem compiling GDL with it on any modern environment, (2) I suggest disabling X by default and unify everything to the wxWidgets driver, since if X is available then wxWidgets should also be available. Now @GillesDuvert's wxWidgets patch seems to work well, and it would be nice to remove the fragmented drivers, i.e. wingdi, wingcc, and X drivers, from the source tree to keep it simple.

Anyway for cluster environments, I think it makes sense to maintain a build without wxWidgets (also without some extra libs?) and change its name from 'MINI' to something more straightforward: 'gnudatalanguage-nogui' or 'gnudatalanguage-cli'.

GillesDuvert:
Hi, the PR #1338 introduces a new patch for plplot for Windows (recompile with DYNDDRIVERS ON). This is more than important.
I'm not sure this patch is played since the GH Action restores a cached version? And does the installer and artifact propagate all these changes and necessary dlls?

pjb7687:
Hi @GillesDuvert, simply you can change the cache key then the cache will be rebuilt. Just to make sure you changed the cache key. And yes all the required dlls will be automatically included.

GillesDuvert:
Hi @pjb7687 thanks, but what is exactly the cache key?

pjb7687:
This one (Windows): https://github.com/gnudatalanguage/gdl/blob/master/.github/workflows/main.yml#L186

You can change it to any new string.

pjb7687:
Just fyi this is for the others (Linux, macOS): https://github.com/gnudatalanguage/gdl/blob/master/.github/workflows/main.yml#L79

GillesDuvert:
ok, this is an incremental cache key. Thanks.

pjb7687:
I usually increase it by 1 but in principle you can use any random string - just make sure it is not the same as the previous one.

Issue #912, opened 2021-04-09
missing procedure truncate_lun


brandy125:
Anybody can implement truncate_lun?

GillesDuvert:
Hi @brandy125 I looked into that, and saw that some file (lun) procedures were in gdl, not in c++. Pretty bad for performance. Looking into that now.

GillesDuvert:
introduced in GDL. please test, as we have no test file for this.

brandy125:
Hi @GillesDuvert, unfortunately truncated_lun does not work as expected:

GDL> openw,1,'testfile'
GDL> writeu,1,fltarr(1000)
GDL> $ls -al testfile
-rw-r--r-- 1 chris users 4000 Apr 14 14:36 testfile
GDL> truncate_lun,1
GDL> $ls -al testfile
-rw-r--r-- 1 chris users 0 Apr 14 14:36 testfile

In this case it should preserve the file size.

Could you have a look to it?
Thanks

GillesDuvert:
indeed.

GillesDuvert:
corrected!

brandy125:
Hi @GillesDuvert, now works fine. Thanks!

PR #916, opened 2021-04-12
Revamped io

GillesDuvert:
complete support of copy_lun, truncate_lun, skip_lun in C++ and for compressed files too. Speed increase for file handling also.


GillesDuvert:
closes #912

PR #937, opened 2021-05-06
fixing ishft

eloirozier:
The faster version of ishft recently implemented had problems dealing with arrays (the test test_ishft.pro failed). One line fixed this. However the code could still benefit from a little cleaning.

### DIFF
--- a/PYTHON.txt
+++ b/PYTHON.txt
@@ -1,6 +1,8 @@
 GDL can embed python as well as being
 embedded in python.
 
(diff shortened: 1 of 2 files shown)


GillesDuvert:
Thanks @eloirozier, patch accepted!
I forgot to correct the GDL master as I have [a faster version of ishft in stock](https://github.com/GillesDuvert/gdl/tree/speedup) but since many people rely on master now, let's do it.

PR #946, opened 2021-05-25
benchmark for windows

eloirozier:
Very small corrections making benchmarks possible to be run and saved on MSWindows

### DIFF
--- a/testsuite/benchmark/benchmark_generate_cpuinfo.pro
+++ b/testsuite/benchmark/benchmark_generate_cpuinfo.pro
@@ -31,7 +31,7 @@ if (os_name EQ 'linux') then begin
    ;;
 endif
 ;
-if (os_name EQ 'darwin') then begin
+if (os_name EQ 'windows') then begin
    ;;
    SPAWN, 'sysctl -n hw.cpufrequency', resu_Hz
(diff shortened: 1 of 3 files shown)


GillesDuvert:
@eloirozier I do not understand: mswindows does have this commands, so the procedure will fail on windows. ???

eloirozier:
Sorry i should have detailed more

Well the command
"date"
exists on mswindows but it outputs more than one string which makes everything crash
So i looked for an other way to get the date, and it turned out that the simpliest was to use "powershell -command..."
And also on mswin the ':' char cannot be put into a file's name so the code had to be corrected anyway.
I hope this explanation is clear enough!

GillesDuvert:
I was referring to 
`SPAWN, 'sysctl -n hw.cpufrequency', resu_Hz`
which is now valid under windows following your changes in `benchmark_generate_cpuinfo.pro ` and `benchmark_info_cpu.pro`

eloirozier:
Now I see where I made a mistake. I didnt know that 'darwin' corresponded to MacOSx so I foolishky replaced it with 'windows', thinking of it as a weird mistake...
I should have completed the IF statement with something like
OR osname EQ 'windows'

PR #948, opened 2021-05-26
Benchmark on windows 2 - the return

eloirozier:
I hope this time it's right...

### DIFF
--- a/src/interpol.cpp
+++ b/src/interpol.cpp
@@ -391,6 +391,8 @@ namespace lib {
     DType t=type; //V's type by default.
     //size of the out value
     SizeT nout=0;
+    //type of output
+    bool isDouble = false;
     
     if (nParam==2) {       // we make X, only V has to be chacked for Nan
       BaseGDL* p1 = e->GetParDefined(1);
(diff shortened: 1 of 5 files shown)


GillesDuvert:
Hi Eloi, do you have a windows machine with gdl installed to test these scripts first? Because **this is not going to work**...
SPAWN sends a command to the OS. the Windows OS does not know about 'sysctl' and this will end in error.
Concerning the code in generate_filename.pro, as GDL has the command 'SYSTIME' that  does the job irrespective of the operating system, I suggest to use it instead of a dangerous SPAWN command. Use also PATH_SEP() function to deal with the peculiarities of Windows vs. Unix+Mac instead of painstakingly teste if the OS is Windows or Linux etc.

GillesDuvert (changes requested):
GDL does not have (yet) timestamp()
Either use cgtimestamp() or create a timestamp() for GDL (easy: take example from cgtimestamp() )

eloirozier:
Thanks for telling me
I assumed timestamp() existed on GDL because i tested my code while IDL librairies where included in my !path variable... I'll use cgtimestamp() instead

alaingdl:
maybe we shall consider to add it, using c++ code ?
https://stackoverflow.com/questions/9527960/how-do-i-construct-an-iso-8601-datetime-in-c

PR #965, opened 2021-06-16
Benchmark on windows (using timestamp) -- tested on windows7, linux

eloirozier:


GillesDuvert:
Do I see a new Qhull function? this needs to be at least mentioned in the PR title.
Another, preferable, way is to make two PRs, one for QHULL debuts and another for benchmark procedures.

eloirozier:
Sorry, I should have precised that the QHull commits were a mistake which I've removed from the branch I'm asking to merge (see the "reversing unclean merging" commit).

You can see in the 'Files changed' section that only 3 benchmark files are modified

GillesDuvert:
My mistake indeed. Hope to see a QHULL PR soon.

PR #967, opened 2021-06-21
Enable Compilation on suse tumbleweed (plpot 5.15.0 and wxWidgets 3.1.5)

GillesDuvert:
As the title says, patch for a defect in plplot wxwidget driver when resizing the window (wxDC needs to be destroyed and recreated to have plplot realize a new pixmap with a different size has been created --- PLESC_RESIZE is not sufficient as before)
Comes with an update of build_gdl.sh that really makes GDL on Suze.
our FindRPC for CMake now add -ltirpc when needed


pjb7687:
FYI I think you can ignore macOS build failure here, this seems to be the build cache issue. Probably this won't happen after this is merged to the master branch.

GillesDuvert:
> FYI I think you can ignore macOS build failure here, this seems to be the build cache issue. Probably this won't happen after this is merged to the master branch.

This probably comes from the 'simplification' I made to Cmakelists.txt
I try a patch

GillesDuvert:
@pjb7687 ok now python is found on OSX, but there is a problem in GDL's Python code (not in sync with python3). My advice for 1.0: remove python dependency for MacOSX.
Next, windows: it does not have python3-numpy installed, only ptthon2.7-numpy, apparently. Yet another prep dependency?

pjb7687:
I think the problem would be probably because the cmake finds Python2 for GDL:
https://github.com/gnudatalanguage/gdl/runs/2917933914?check_suite_focus=true#step:8:167
```
Python                 ON (libs:/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/lib/libpython2.7.tbd; headers:/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/usr/include/python2.7)
```

I am currently working on it, will update when it works again

PR #968, opened 2021-06-21
Qhull

eloirozier:
Here is an initial implementation of Qhull, using qhull libqhullcpp and libqhull_r libraries.

I had to update CMake files, update the FindQHULL procedure (it works on my machine but it might need to be extended), remove the qhull functions declaration from triangulation.hpp and trianguation.cpp, and create new cpp source files for qhull.

From all the options available in IDL :

`Keywords: [, BOUNDS=variable ] [, CONNECTIVITY=variable ] [, /DELAUNAY ] [, SPHERE=variable ] [, VDIAGRAM=variable ] [, VNORMALS=variable ] [, VVERTICES=variable ]`

only BOUNDS, CONNECTIVITY, /DELAUNAY and VVERTICES are coded yet.
_EDIT: all keywords are coded but SPHERE_
Next qhull pull request should adress these options as well as proposing a working qgrid3 function


pjb7687:
Hi @eloirozier,

FYI, to make the CI build work, you can do below additionally:

1. Add the QHull packages to `build_gdl.sh`
 
https://github.com/gnudatalanguage/gdl/blob/master/.ci/build_gdl.sh#L34-L65

2. Update the keys of the package cache (e.g., `v2` to `v3`):

https://github.com/gnudatalanguage/gdl/blob/master/.github/workflows/main.yml#L67

and 

https://github.com/gnudatalanguage/gdl/blob/master/.github/workflows/main.yml#L189

And the builds should work fine.

eloirozier:
Thanks @pjb7687 
I've updated the packages lists in build_gdl.sh for windows and macos (it's more complicated on linux).
Could you please precise what do the keys of the package cache do, and why they need to be changed ? (my guess was that the script avoids going through installing all the dependencies if the current key is found in cache, i'm not sure at all though)

pjb7687:
> Could you please precise what do the keys of the package cache do, and why they need to be changed ? (my guess was that the script avoids going through installing all the dependencies if the current key is found in cache, i'm not sure at all though)

Your guess is right. You can just change it to an arbitrary new one.

CMakeModules/FindQHULL.cmake
```diff
@@ -1,42 +1,30 @@
...
+        )
 
-find_path(
-  QHULL_INCLUDE_DIR
-  qhull/qhull_a.h libqhull/qhull_a.h
-  PATHS /usr/local/include /usr/include
-)
+if(QHULL_INCLUDE_DIR)
+    find_library(QHULL_R_LIBRARY NAMES qhull_r HINTS ${CMAKE_PREFIX_PATH})
+    find_library(QHULL_CPP_LIBRARY NAMES qhullcpp HINTS ${CMAKE_PREFIX_PATH})
```
pjb7687, line 15: On my machine, there is no `libqhullcpp` library but only `libqhull`.

alaingdl: yes, @eloirozier mentioned that this file appeared in version 8.0+ which is not available on most Linux Distros
I got it doing : 
```
mkdir Qhull_Git
cd Qhull_Git/
git clone https://github.com/qhull/qhull
cd qhull/
make
[...]
ls lib/
libqhullcpp.a  libqhull_r.so  libqhull_r.so.8.0  libqhull_r.so.8.0.2  libqhullstatic.a  libqhullstatic_r.a
```

pjb7687: ~~how about:~~
~~find_library(QHULL_CPP_LIBRARY NAMES qhullcpp qhull HINTS ${CMAKE_PREFIX_PATH})~~

Just confirmed it doesn't work

pjb7687: I did some research.
1. On Ubuntu Focal, I have tried downloading and compiling the same version of QHull currently available on Apt (2015.2, downloaded from https://github.com/qhull/qhull/releases/tag/2015.2). By compiling it, I got `libqhullcpp.a`. When I compiled this PR with the resulting file, it worked fine.
2. On Fedora 34, the QHull package is still 7.2.1 (qhull-devel-7.2.1-7), this also contains `libqhullcpp.a`. I confirmed this PR compiles well.

Therefore I think the QHull library 7 also provides the `libqhullcpp.a`, but somehow it is not included in the Ubuntu Focal package. In other words, this seems to be a Focal-only problem. Since this is a static library, perhaps we can temporarily ship the binary for Ubuntu Focal.

Also some more things:

1. On SuSE Tumbleweed, you will need 2 packages: `qhull_r-devel` and `qhullcpp-devel-static`. After installing the packages, this PR compiles well.
2. On Windows, although the QHull package available via MSYS2 is the latest (2020.2), this results in a link error at the end, this is weird.
3. On macOS, CMake fails to find the libraries, although the one provided by Homebrew is the latest (2020.2). This is weird too.

PR #981, opened 2021-07-01
Fix #980

### LINKED ISSUE #980: bug in delvarrnew
I found the reason why `test_delvarrnew` stuck during windows tests. It turned out a bug in GDL. The test stuck because the 'assertion' error is displayed as a dialog on Windows, waiting for the user to dismiss the dialog.

On Linux, the error is reproducible too:

```
~/build/testsuite$ ../src/gdl -quiet delvarrnew.pro
Unable to init server: Could not connect: Connection refused
02:38:07: Error: Unable to initialize GTK+, is DISPLAY set properly?
WARNING: wxWidgets not initializing, widget-related commands will not be available.
IGOT            FLOAT     = Array[4]
IGET            FLOAT     = Array[4]
gdl: /home/pjb7687/gdl/src/datalistt.hpp:124: T& EnvTypePreAllocListT<T, defaultLength>::operator[](SizeT) [with T = EnvType; long long unsigned int defaultLength = 64; SizeT = long long unsigned int]: Assertion `i<sz' failed.
Magick: abort due to signal 6 (SIGABRT) "Abort"...
Aborted
```

The error occurs here:
https://github.com/gnudatalanguage/gdl/blob/master/src/envt.cpp#L1310

pjb7687:


pjb7687:
It seems to be fixed now. When I manually run `delvarrnew.pro`, I got a `GDLException` instead:

```
$ ../src/gdl.exe delvarrnew.pro
  GDL - GNU Data Language, Version 1.0.0-rc.3 git
- Using WxWidgets as graphics library (windows and widgets).
- No startup file read (GDL_STARTUP/IDL_STARTUP env. var. not set).

IGOT            FLOAT     = Array[4]
IGET            FLOAT     = Array[4]
% env and pro out of sync.
  At: C:\msys64\home\pjb7687\gdl\build\testsuite\rnewtest.pro
% env and pro out of sync.
```

I have no ideas what `env and pro out of sync` means, and whether it is okay or not, though.

GillesDuvert:
exactly my point. This test pops up from empty air, is rather convoluted and does not say what it is intended for.
besides, apparently it fails because of many otehr causes than int intended purpose.
I would remove it from the distro and put it in testsuite/wrong .
wrong not meaning its intrinsically wrong but to attract attention and, hopefully, a better version.

alaingdl:
If I remember well, this was provided by @maynardGK  years ago (in 2015 ?!). And the bug came from issue in Chianti code (??)
Maybe @maynardGK can help us to clarify & improve that test ? thanks !

pjb7687:
Thanks for the reviews. The main purpose of this PR is to debug for loop in `envt.cpp`:
It is obviously a bug removing items from the first to the last while popping the last item.

Regarding the test itself, I suggest opening another issue for it?

GillesDuvert:
merged.
@pjb7687 I've patched build_gdl.sh as the GDLDE version was not defaulted to v1.0.0
Also, Windows installer seems to insist to install in gnudatalanguage/bin (or even bin/bin) instead on gnudatalanguage ??

GillesDuvert:
BTW do not know how to approve this PR :smile:

pjb7687:
> Also, Windows installer seems to insist to install in gnudatalanguage/bin (or even bin/bin) instead on gnudatalanguage ??

Yeah unfortunately the GDL binary should be in `bin` because the DLLs try to find the resource files in a relative path (../share/ or ../lib/). BTW, `bin/bin` is weird, should not be like that.

PR #987, opened 2021-07-02
refactor: in routines 'h5[ad]_read_fun()', create a type-guard 

ogressel:
…for the HDF5 element datatype to be in-line with existing coding paradigm.

### DIFF
--- a/src/hdf5_fun.cpp
+++ b/src/hdf5_fun.cpp
@@ -568,6 +568,8 @@ hid_t
 
     // for array datatypes, determine the rank and dimension of the element
     int elem_rank=0;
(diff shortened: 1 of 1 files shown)


ogressel:
I just noticed that this breaks the `test_hdf5_string` build test. With `git bisect` I traced the issue to this commit. Looking into figuring out the cause now...

ogressel:
The `elem_dtype` that is created via `H5Tcopy()` evaluates to `5` in  `mapH5DatatypesToGDL()` despite testing `H5Tequal()` against the original `datatype`... very weird.

PR #990, opened 2021-07-02
X is NOW the default name of graphic device on every machine. 

GillesDuvert:
Variants such as "WIN" or "MAC" are still available using a switch in the command line.

### DIFF
--- a/src/devicewx.hpp
+++ b/src/devicewx.hpp
@@ -38,8 +38,8 @@
 class DeviceWX : public GraphicsMultiDevice {
   
 public:
-
-    DeviceWX(std::string name_="MAC") : GraphicsMultiDevice( 1, 3, 3, 0) { //force decomposed=true until we find a better way (::wxDispayDepth() crashes)
+//is called "WIN" as there is no other choice on WINDOWS, and "WIN" is still ok on linux and macOSX
+    DeviceWX(std::string name_="WIN") : GraphicsMultiDevice( 1, 3, 3, 0) { //force decomposed=true until we find a better way (::wxDispayDepth() crashes)
         name = name_; 
         DLongGDL origin(dimension(2));
         DLongGDL zoom(dimension(2));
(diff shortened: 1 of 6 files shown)


pjb7687:
Changing `MAC` to `X` makes sense. But we would still want to keep `WIN`, considering IDL behavior:
https://www.l3harrisgeospatial.com/docs/supported_devices.html#Microsof

GillesDuvert:
Theres is always the switch `--true-device-names` . But yes, you are right. "MAC" is in fact unix=X as !VERSION.OS_FAMILY is 'unix'.

GillesDuvert:
Actually, I see only one code in the IDL suite that will select 'WIN' on Windows:
```
  case STRUPCASE(!VERSION.OS_FAMILY) of
    'WINDOWS': newDevice = 'WIN'
    'UNIX': newDevice = 'X'
    'MACOS': newDevice = 'MAC'
    else: MESSAGE, 'Unknown OS_FAMILY.'
  endcase
```
And the 3rd party codes (e.g., NASA's idlastro) can eventually be patched (only coyote library seems to test such things).

So it would be even better to forget about switches, have 'X' on linux and MAC and 'WIN' on windows and stay like that.

I will just make this tiny change instead of this larger PR.

alaingdl:
I worked on that long time ago, and for sure made tests to ensure it was OK.

Are you sure IDL on OSX is not changing with time ? (e.g was MAC on OSX 10.10 and is now X on 11.1)

Or maybe we still have very old code since time for Mac OS before OSX ?!

I cannot made tests on a Mac today but maybe during the WE

A.

GillesDuvert:
I suspect !VERSION.OS_FAMILY has switched from 'MACOS' on old macs to 'UNIX' when Macs went Intel.
That was why I said  3rd party codes can eventually be patched.
But let's keep a switch for the unexpected cases: --MAC to get it as MAC.

PR #991, opened 2021-07-03
Rename to x:  Changed behaviour, see #990 

GillesDuvert:
'X' is the default name of graphic device on linux and MacOSX and 'WIN' on windows. the --MAC switch will enable the name "MAC" on Mac machines in the hope it can be useful for running some procedures.


GillesDuvert:
to recompile

PR #993, opened 2021-07-03
update Debian wx package hint (to libwxgtk3.0-gtk3-dev)

slayoo:


GillesDuvert:
Indeed, that's what is done in build_gdl.sh (prefer libwxgtk3.0-gtk3-dev, then libwxgtk3.0-dev, then libwxgtk2.8-dev)
You sure can merge.

PR #997, opened 2021-07-06
corrected severe bug when psym value was wrong (9 or gt. 10)

GillesDuvert:


alaingdl:
hi. Could you give an example of the problem ?
I have no difference on my tests here !

A.

GillesDuvert:
was: `plot,dist(4),psym=10 ` 
crashed because a recent speed optimization was done whatever the psym number, even > 8.

PR #1013, opened 2021-07-16
GRIB_API cleanup

slayoo:


GillesDuvert:
... and replace grib and grib_api with eccodes everywhere possible? for clarity?

slayoo:
Gilles, I doubt it would improve clarity. Inside ECCODES, nothing has changed (still grib_abi.h, still grib-prefixed routine names). IDL also prefixes the routines with GRIB_. Users want to read GRIB files, and they do not care if the handling is with ECCODES or anything else - like with graphics formats and ImageMagick.

slayoo:
https://www.l3harrisgeospatial.com/docs/GRIB_Routines.html

GillesDuvert:
Ok, thanks @slayoo for the explanation.

opoplawski:
So the problem with this is that eccodes does not support all of the architectures that Fedora does, and so we compile against grib_api on those.  I'm going to revert the essential part of this for now on Fedora.  The other option is that we simply drop support for those architectures for gdl.

opoplawski:
At the moment those arches are i686, armv7hl, s390x.

slayoo:
Thanks @opoplawski 
Perhaps let's just disable GRIB support in GDL on those platforms on Fedora?

opoplawski:
That's certainly a possibility as well.

GillesDuvert:
but is not https://github.com/gnudatalanguage/gdl/pull/1054 a solution?

PR #1022, opened 2021-07-21
PS outputs now have correct dimensions and position.

GillesDuvert:
However we do not success in preserving the aspect ratio of Hershey fonts and PSYMs as the plplot ps driver rotates, stretches and scale a fixed-aspect internal 'surface'.
The aspect is thus only preserved for sizes that scale in proportion 4/3.
To avoid unpleasant surprises, the font is always the PostScript font, not the Hershey one as the PS fonts are not stretched by the plplot driver.
It would be marginally possible to get nice plots by customizing the postScript header afterwards, you are welcome.

### DIFF
--- a/src/deviceps.hpp
+++ b/src/deviceps.hpp
@@ -60,7 +60,7 @@ class DevicePS: public GraphicsDevice
 
     if( nx <= 0) nx = 1;
     if( ny <= 0) ny = 1;
-    actStream = new GDLPSStream( nx, ny, (int)SysVar::GetPFont(), encapsulated, color, bitsPerPix);
+    actStream = new GDLPSStream( nx, ny, (int)SysVar::GetPFont(), encapsulated, color, bitsPerPix, orient_portrait);
 
     actStream->sfnam( fileName.c_str());
 
@@ -88,9 +88,8 @@ class DevicePS: public GraphicsDevice
      actStream->setopt( "portrait",NULL);
      actStream->sdidev( PL_NOTSET, PlplotInternalPageRatioXoverY, PL_NOTSET, PL_NOTSET ); //only OK if page ratio is 540x720 
      actStream->spage(DPI, DPI, XSIZE, YSIZE, YOFF, XOFF);
-
     } else {
-     actStream->spage(DPI, DPI, YSIZE, XSIZE, YOFF-YSIZE, XOFF); //invert axes, displace as does IDL..
+     actStream->spage(DPI, DPI, YSIZE, XSIZE, YOFF-XSIZE, XOFF); //invert axes, displace as does IDL..
      actStream->sdiori(2);
     }
 
--- a/src/deviceps.hpp
+++ b/src/deviceps.hpp
@@ -25,17 +25,13 @@
 
 #include "objects.hpp"
 
-#ifdef _MSC_VER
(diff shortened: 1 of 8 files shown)


gabrielastro:
Hello,
First of all, thank you (all of the developers) so much for all your work on GDL! I am impressed by the amount of efforts and the results, and enjoy very much using it.
I have the issue mentioned above of stretched Hershey fonts (which I want very much to use) for non-4:3 plot. Is there a workaround in the meantime or does someone know how to change the PostScript header? I am using 1.0.2 Git at epoch 1683496800 on 20.04.1-Ubuntu.
Thanks a lot!
Gabriel

GillesDuvert:
Hi, thanks for the kind comments.
Even if there has been improvements here since 1.0.2, the problem persists because it is in plplot itself.
plplot has a number of example programs that make nice plots with their default values, but are plain wrong when non-default sizes are used. In this particular case, calling any of their example programs with command line options like
  `plplot-5.15.0/build/examples/c++/x16 -geometry 1200x200 -drvopt "hrshsym=1","text=0" -dev ps`
show the problem you report.
And, plplot is not supported anymore, at least comatose since 2019. I suspect it would be a 2 lines patch somewhere in the plplot code...

gabrielastro:
Thanks a lot for your quick answer! Actually, a trick occurred to me: set the device size ratio x:y = 4:3 as `plplot` wants but then use a large ymargin only at the top or bottom, so that the actual plot area have the desired aspect ratio! E.g.:
```IDL
GDL> device,filename='alleluia.eps',/color,/inch, /dec, xsize=8., ysize=6., /enc
GDL> plot, [1,2,3], [0,2,1], chars=1.7, title='!6Test', thick=3,xthick=3,ythick=3,charthick=3, ymargin=[12,3]
GDL> device, /close
GDL> $epstopdf alleluia.eps
```
The result:

[alleluia.pdf](https://github.com/gnudatalanguage/gdl/files/11915227/alleluia.pdf)

The ratio of the ymargins sets the ratio of the actual plot. When converting to pdf, the whitespace is cropped automatically (and at worse one could call an external tool to do this cropping). This looks like a very acceptable solution.

By the way, when making characters thicker, as one should for better legibility, they get shaved off at the edge; apparently, `plplot` ignores the margins and sets them tight around the stroke centers. My workaround (not done here) is to use `XYOUTS` to place a character (e.g., respectively `_` or `|`) at the desired boundary of the figure, as "struts". Of course, I make them lime green to place them and then background-coloured for the final version. For the previous plot, for the bottom border (to prevent the x-axis numbers from being shaved off):
```IDL
xyouts, 2, -.5, '_', col='00ff00'x
```

And so on for the other sides. Note that these "protection characters" will sometimes be ignored if placed too far off the plot region; better to start close and move them out slowly. It requires a bit work but actually not too much.

I hope this can help someone! Could this thread be placed somewhere "easier" to find? I looked a lot on the Internet for this problem of stretched characters in GDL and it took a long time before I found your thread from above (which was only mentioning the issue). Without this solution, GDL would be much less attractive—no non-square plots with Hershey fonts would be impractical… But I am glad that there is a solution! Thanks again for your time and work, and looking forward to all the coming developments!

GillesDuvert:
Hi thanks for the workaround. It is now avaiable in[ the wiki](https://github.com/gnudatalanguage/gdl/wiki/Known-issues).

PR #1025, opened 2021-07-25
add some more features to build_gdl.sh

pjb7687:
- Changed `ROOT_DIR` to the source directory, not the dir one level above.
- Add dry run mode for 'prep': `build_gdl.sh prep_packages_dryrun`
- Separated configure, build, and install: `build_gdl.sh configure`, `build_gdl.sh build`, `build_gdl.sh install`
- Added `INSTALL_PREFIX` environment variable, which can be used to specify install location
- Removed `mini` build type, introduced three new build types: `standard`, `headless`, and `debug`.
  - `standard`: everything without MPI
  - `headless`: everything without wxWidgets
  - `debug`: includes both MPI and wxWidgets, for debugging purposes
- Removed Windows i686 builds.

### DIFF
--- a/.ci/build_gdl.sh
+++ b/.ci/build_gdl.sh
@@ -12,7 +12,7 @@
 LANG=C #script works only in english
 ME="build_gdl.sh"
 Configuration=${Configuration:-"Release"}
-DEPS=${DEPS:-"full"}
+DEPS=${DEPS:-"standard"}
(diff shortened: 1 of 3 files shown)


alaingdl:
Sounds great ! too late and too tired to test now, but for sure we will try tomorrow !

PR #1040, opened 2021-08-03
LA_SVD

algorab-dream:
First version of LA_SVD

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -33,6 +33,7 @@ jobs:
           echo "SETUP_TARGET_FOR_COVERAGE_LCOV(NAME codecov EXECUTABLE \${CMAKE_BUILD_TOOL} test)" >> testsuite/CMakeLists.txt
       - name: Build GDL
         run: |
+          scripts/build_gdl.sh configure
           scripts/build_gdl.sh build
       - name: Test GDL
         run: |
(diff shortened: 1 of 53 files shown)


slayoo:
Thanks!
Please do include a test routine so that the newly added code has some example usage in the repository and is covered with a test under continuous integration (to check if future changes to GDL do not introduce regressions).

alaingdl:
Thanks. @bouloukaouze will add some tests and also have to consider the (d)complex types ...

slayoo:
@bouloukaouze note that the last commit is in this PR - I guess it was meant to be on a separate branch. It would also be great to make a PR to the 1.0 branch

GillesDuvert:
hi @bouloukaouze,
concerning file_lines, your new code seems restricted to non-compressed UNIX files. To insure full compatibility (earlier behaviour) it is still needed to use gzread, etc if the file is compressed.
Besides, a 'line' is not always '\n'. It can be ('\r' or '\r\n' and should be tested on all platforms).
Making two cases, if (compressed) old code else new code (with '\r','\r\n') should do the job.

GillesDuvert:
conflicts since I've patched 'file' in #1090  to speed increase and preserve compressed file handling (and windows line termination as well).
This PR can be merged as soon as its last commit (https://github.com/gnudatalanguage/gdl/pull/1040/commits/cc374c667975b5edfdae59c8b4257c86a5564790) is reverted

GillesDuvert:
done after removing https://github.com/gnudatalanguage/gdl/commit/cc374c667975b5edfdae59c8b4257c86a5564790 's conflicts

PR #1047, opened 2021-08-08
Switch to un-deprecated Numpy API.

jkohnert:
This patch fixes the numpy depracated api warung when compiling gdl. It also fixes two deprecation warning on newer cmake versions by expicitly enabling the behaviour that is default on cmake 3.10 and cmake 3.12, respectively.

Tests for gdl still work; however I'm not too sure on how to run the python test not having installed the freshly built GDL module beforehand.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ build/
 install/
 gdlde/
 gdlde*.zip
+.vscode
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,10 @@
 
 cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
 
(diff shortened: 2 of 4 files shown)


slayoo:
@pjb7687 back in the days of Travis CI, we were launching with pytest the python test suite contributed by @olebole (https://github.com/gnudatalanguage/gdl/blob/master/testsuite/python/test-GDL.py):
https://github.com/gnudatalanguage/gdl/blob/094db2eb65f1152a9782b5979818db7906feeef8/.travis.yml#L275 

Am I wright that this got lost with the transition to Gihub Actions? 
It would really be great to have it back (and then the required script would answer @jkohnert's question).

PR #1054, opened 2021-08-14
WIDGETS improvements and bug removal

GillesDuvert:
Also, added heap_save and heap_nosave (as stubs doing pretty much nothing --- we do not support the NOSAVE property yet)


GillesDuvert:
Will update this as I found the widget event handling is not perfect (by far)

PR #1067, opened 2021-08-20
Qhull

eloirozier:
Here is a new, clean PR for Qhull.
It still needs to be re-tested on multiple platforms, especially the additions to the `builld_gdl.sh` script which might still need some work for Windows (there are problems with the shared librarires)

What the `build_gdl.sh` script does for qhull on Windows and Linux (since there are problems with the ubuntu package, the c++ library not being shipped as it should) :
 - download the qhull 2020.2 archive from github
 - extract and compile the source code with make (adding the SO=dll option for windows)
 - specify the directory to cmake at `configure ` using -DQHULLDIR=...

Questions I will try to answer on my own:

For windows: I still dont know what would be the best way to ensure that the correct libqhull_r.dll (located in `$ROOT_DIR/qhull-2020.2/lib`) is copied to the `install/bin `directory. Should I copy the built dlls to `/mingw64/bin/` after compilation at `configure` time ? Because this would mean replacing the existing dlls if qhull was already installed with pacman...

For Linux: I wonder if an additional step needed to link the built `libqhull_r.so` to the gdl executable ?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -177,7 +177,7 @@ jobs:
         with:
           path: |
             C:\msys64\cache
-          key: cache-gdl-deps-windows-msys2-${{ matrix.arch }}-v5
+          key: cache-gdl-deps-windows-msys2-${{ matrix.arch }}-v6
       - name: Check Version of Latest GDL Workbench
         if: matrix.arch == 'x86_64'
         id: gdl_workbench
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -195,6 +195,7 @@ jobs:
         run: |
           pacman -Q | tr ' ' '-' | sort > ~/snapshot1.txt
           scripts/build_gdl.sh prep
+          ls ../
           ls ../MINGW-packages/mingw-w64-graphicsmagick/
           mv ../MINGW-packages/mingw-w64-graphicsmagick/mingw-w64-*-graphicsmagick-*.tar.* /var/cache/pacman/pkg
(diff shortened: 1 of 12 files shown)


pjb7687:
> For windows: I still dont know what would be the best way to ensure that the correct libqhull_r.dll (located in `$ROOT_DIR/qhull-2020.2/lib`) is copied to the `install/bin `directory. Should I copy the built dlls to `/mingw64/bin/` after compilation at `configure` time ? Because this would mean replacing the existing dlls if qhull was already installed with pacman...
> 
> For Linux: I wonder if an additional step needed to link the built `libqhull_r.so` to the gdl executable ?

I think the easiest way is to just link the libs statically, then you don't have to care about the shared libs.

eloirozier:
In effect with a static link it now works on all platforms, thanks @pjb7687
However there is now a weird problem with the windows build check, which has to do with graphicsmagick
I modifed line 180 of main.yml: 

key: cache-gdl-deps-windows-msys2-${{ matrix.arch }}-v5
TO:
key: cache-gdl-deps-windows-msys2-${{ matrix.arch }}-v6

maybe i missed something ?

GillesDuvert:
Hi Mageia QHull libraries are not found by CMake.
they are in `/usr/lib64/libqhull_p.so` and  `/usr/lib64/libqhull_r.so`
Please add a note on the Install guides (Wiki) to guide people towards the 'good' Qhull library (obviously not the system's one).
Also, QHULL should be credited in the README.md

fawltylanguage:
Qhull is not found on OpenSuSE Tumbleweed, used to work before this commit. System qhull is 2020.2, it should work.

Issue #1035, opened 2021-08-01
New user: can't create a plot window


Genie2000:
Hi,

I am an IDL user wanting to try out GDL. 

I am using Linux Mint 19.1 Cinnamon. I am normally a Windows user, and my linux knowledge is extremely limited.

I installed GDL using the posted instructions. It seemed to proceed fine.

Below is a copy from the terminal window with some reactions.
***********************************************************************************************************************************************

          (base) **@***********:~/GDL/gdl$ bash quick_start_GDL.sh
          Running current GDL with ./src/pro and ./testsuite in GDL_PATH
          +/home/***/GDL/gdl/src/pro/:+/home/***/GDL/gdl/testsuite/
            GDL - GNU Data Language, Version 1.0.0-rc.3 git
          - For basic information type HELP,/INFO
          - Using WxWidgets as graphics library (windows and widgets).
          - No startup file read (GDL_STARTUP/IDL_STARTUP env. var. not set). 
          - Please report bugs, feature or help requests and patches at: https://github.com/gnudatalanguage/gdl
          
          GDL> window  
          ../src/common/object.cpp(251): assert "classTable->Get(m_className) == NULL" failed in Register(): Class "wxCommandEvent" already in RTTI table - have you used IMPLEMENT_DYNAMIC_CLASS() multiple times or linked some object file twice)?
          Gtk-Message: 13:55:56.825: GtkDialog mapped without a transient parent. This is discouraged.
          
          (gdl:9390): Gtk-WARNING **: 13:55:56.834: gtk_window_set_titlebar() called on a realized window
          Gtk-Message: 13:55:56.835: GtkDialog mapped without a transient parent. This is discouraged.
          quick_start_GDL.sh: line 91:  9390 Trace/breakpoint trap   (core dumped) ${Absolute_PATH}/src/gdl
          (base)  **@***********:~/GDL/gdl$ 

***********************************************************************************************************************************************
I get a window with the error message and a "stop" or "continue" option with the former exiting GDL and the former simply repeating the message.

Any help with fixing the issue would be appreciated. 

It might be worth noting that I had previously installed GDL 0.9.7. I had been able to produce a plot window before, but now this is how it looks:

            GDL - GNU Data Language, Version 0.9.7
          
          - For basic information type HELP,/INFO
          - Default library routine search path used (GDL_PATH/IDL_PATH env. var. not set): 
            +/usr/share/gnudatalanguage/lib
[...]

PR #1076, opened 2021-08-23
remove libwxgtk3.0-gtk3-dev from build_gdl.sh?

slayoo:


pjb7687:
1. Actually in Ubuntu focal, `libwxgtk3.0-dev` is deprecated and GDL only works with `libwxgtk3.0-gtk3-dev`. The problem mentioned in #1035 only happens in Ubuntu bionic. Instead of removing `libwxgtk3.0-gtk3-dev`, I think changing the order of these two libraries would work:

```
libwxgtk3.0-gtk3-dev,libwxgtk3.0-dev
```

to 

```
libwxgtk3.0-dev,libwxgtk3.0-gtk3-dev
```

so that `libwxgtk3.0-dev` is preferred first.

2. Perhaps we can also add a routine in `CMakeLists.txt` to check whether the WX version matches with the one used by Plplot.

GillesDuvert:
the build fails on ubuntu because the wxwidgets is not found at all... A cmake test? looking for the linked kibraries of the plplot driver? say:
```
$ ldd /usr/local/lib/plplot5.15.0/drivers/wxwidgets.so |grep gtk
        libwx_gtk2u_core-3.1.so.5 => /usr/local/lib/libwx_gtk2u_core-3.1.so.5 (0x00007f7c16812000)
        libgtk-x11-2.0.so.0 => /lib64/libgtk-x11-2.0.so.0 (0x00007f7c15e21000)
```
OK, this is only for linux (ldd)

PR #1094, opened 2021-09-01
add J Tappin to JOSS paper author list (arbitrary author order for now)

slayoo:
@jtappin, please confirm.
We are inviting all contributors and packagers to join the author list (but the process has just started)

### DIFF
--- a/paper/paper.md
+++ b/paper/paper.md
@@ -20,6 +20,9 @@ authors:
   - name: Sylwester Arabas
     orcid: 0000-0003-2361-0082
     affiliation: "5"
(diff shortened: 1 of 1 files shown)


jtappin:
Thanks, 
I'd suggest putting the folks who have contributed to the interpreter up front and those who've contributed a few library routines at the end.

I do have an ORCID ID but I'll need to retrieve it.

slayoo:
Thanks! Merging then, and please make a PR with ORCID if available.
We will certainly discuss the author order at some point, but let's first gather the authors :)

PR #1095, opened 2021-09-01
took more attention to text widgets input,output, and sizes.

### LINKED ISSUE #1092: Text widgets & Arrays (another one)
When setting a multi  line text widget with an array, the widget shows (correctly) one value per line when the text is an initial value in the `widget_text` call, but when it is set using `widget_control` the values are concatenated into a single string.

Quick demo code attached:
[test_text2.pro](https://github.com/gnudatalanguage/gdl/files/7077660/test_text2.txt)

Manjaro Linux, current GIT version of GDL (since the recent `widget_text` fixes, but the issue was there before but not isolated or analysed, so **not** a regression).

GillesDuvert:
Should also close #1092


GillesDuvert:
mmm... not completely satisfactory in one text_widgets case. This is definiteley WIP.

PR #1098, opened 2021-09-02
adding Takeshi to the author list (author order arbitrary as of now)

slayoo:
@tenomoto
おはよう Takeshi, we are inviting all GDL contributors and packagers to join the author list for the JOSS paper under preparation (but the process has just started). Please confirm, thanks,
Sylwester


slayoo:
ping @tenomoto :)

slayoo:
@alaingdl or @tenomoto  - plase confirm if we have consent from Takeshi (which is why I left this PR open)

PR #1170, opened 2021-10-26
Implement hyperslab selection when reading HDF5 datasets

ogressel:
Add 'h5s_select_hyperslab_fun()', 'h5s_create_simple_fun()', 'h5s_create_scalar_fun(), and enable data space functionality in 'h5d_read_fun()'

Extended the unit tests as well, and added some more stringent comparisons.

Still on the learning curve with GDL's environment framework and the guarding of variables, so please have an extra careful look if you catch anything odd.


slayoo:
let me close & reopen to trigger build with the updated CI scripts

ogressel:
I just noticed that data spaces created with my implementation of 'h5s_create_simple_fun()' have dimensions in the wrong order... will fix that in a moment. That should also make the odd 'reverse_memspace_dims' switch in the 'h5d_read_fun()' redundant. Man these memory order things between GDL and C++ are a real nuisance.

Issue #1140, opened 2021-09-20
!null and Exp() : Internal error: !NULL destructor called.


alaingdl:
Exp(!null) is broken, but Gsl_Exp(!null) not :smile: 
```
GDL> print, exp(!null)
Internal error: !NULL destructor called.
Save your work and restart GDL (GDL is still functional, but !NULL will not work anymore).
Please report at https://github.com/gnudatalanguage/gdl/issues
% EXP: Variable is undefined: !NULL
% Execution halted at: $MAIN$          
```

(sorry not to look at that, but I would like to finish the changes I am working on)

GillesDuvert:
please label "NULL" every issue related to the somewhat messy introduction of !NULL in GDL code...

GillesDuvert:
closed by #1174

Issue #1102, opened 2021-09-02
crash withe basic call to HELP : help, out='will_crash'


alaingdl:
```
(u1404)
GDL> help, out='tototototo'
*** Error in `/home/alaingdl/GDL/last-gdl/src/gdl': free(): invalid pointer: 0x00000000029ed8a0 ***

(debian 10 and U2004)
GDL> help, out='tototototo'
free(): invalid pointer
Magick: abort due to signal 6 (SIGABRT) "Abort"...
```

alaingdl:
Same issue may appear (with crash or no message), is in : ` a=histogram(findgen(10), loc='bad')`
which should issue a warning like : 
`% HISTOGRAM: Expression must be named variable in this context: <STRING   ('bad')>.`

GillesDuvert:
One-liner? This is important.

alaingdl:
I am working on that, with progress. I can just fix it locally, but I can also consider more generally
the situation where Keyword should not be a so-called "named variable"

The problem is that some good old code (in fact GetString in envt.cpp) is not that simple to cleanup

 const string EnvBaseT::GetString( BaseGDL*& p, bool calledFromHELP)

alaingdl:
And I would say #1089 is way more annoying because HISTOGRAM() is frequently used, and `test_histo.pro`
covers nothing ...

GillesDuvert:
histogram needs to be rewritten entirely at some point as it is 20 times slower than IDL. GSL is not known for its speed, but here the difference is striking.

GillesDuvert:
The crash was inavoidable as the programer did not check the variable was 'writeable'.
This because the functions to access variables are not well defined enough. One should insure a set of easy functions that make the difference between local and global variables as well as readonly and writeable variables (and prabably the boolean variables).
Nothing difficult, WIP.

GillesDuvert:
closed by #1174

Issue #428, opened 2018-07-20
operation '+' with string and !null  (+side effect on !path)


alaingdl:
a='aa'
b=!null
c=a+b

no message in GDL ...

IDL>  c=a+b
% Variable is undefined: B.
% Execution halted at: $MAIN$          

string should behave as other numerical types with !null

side effect :

!path=!path+!null      give : 
IDL> !path=!path+!null
% Variable is undefined: <UNDEFINED>.

GDL> !path

GDL> !path=!path+!null
GDL> !path
!NULL

GillesDuvert:
closed by #1174

Issue #221, opened 2018-04-19
!NULL-related segfault 


slayoo:
As reported by @maynardGK on SF.net (https://sourceforge.net/p/gnudatalanguage/bugs/698/):

```
GDL> tt=[ ]
GDL> p=ptr_new(tt)
GDL> tstcmp=lonarr(3,3,4)
GDL> i=where(tstcmp ne *p,nc)
gdl: /build/gnudatalanguage-nJX6mN/gnudatalanguage-0.9.8/src/prognodeexpr.cpp:340: void BinaryExprNC::AdjustTypesNCNull(Guard<BaseGDL>&, BaseGDL*&, Guard<BaseGDL>&, BaseGDL*&): Assertion `false' failed.
Aborted
```

```
Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007fffef0c8e7b in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007fffef0ca231 in __GI_abort () at abort.c:79
#2  0x00007fffef0c19da in __assert_fail_base (fmt=0x7fffef214d48 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x7ffff79aec6b "false", file=file@entry=0x7ffff7a59918 "/build/gnudatalanguage-nJX6mN/gnudatalanguage-0.9.8/src/prognodeexpr.cpp", line=line@entry=340, function=function@entry=0x7ffff7a5a200 <BinaryExprNC::AdjustTypesNCNull(Guard<BaseGDL>&, BaseGDL*&, Guard<BaseGDL>&, BaseGDL*&)::__PRETTY_FUNCTION__> "void BinaryExprNC::AdjustTypesNCNull(Guard<BaseGDL>&, BaseGDL*&, Guard<BaseGDL>&, BaseGDL*&)") at assert.c:92
[...]

PR #1174, opened 2021-11-02
!Null and other undefined variables (various patches)

GillesDuvert:
closes at least  #1140, #1102, #428 , #221, #220


GillesDuvert:
OOPS this PR introduces a regression.

GillesDuvert:
OOPS this because some GetKW calls still need to be redefined as GetWriteableKW. WIP.

PR #1179, opened 2021-11-04
correct typo in write_csv.pro (reported by brbarker)

slayoo:

PR #1180, opened 2021-11-04
add test_write_csv.pro

brbarker:
Added "test_write_csv.pro" to testsuite and this is a record for that

### DIFF
--- a/testsuite/LIST
+++ b/testsuite/LIST
@@ -190,3 +190,4 @@ test_wordexp_null_string.pro
 test_xdr.pro
 test_xmlsax.pro
 test_zeropoly.pro
+text_write_csv.pro
--- a/testsuite/LIST
+++ b/testsuite/LIST
@@ -190,4 +190,4 @@ test_wordexp_null_string.pro
 test_xdr.pro
 test_xmlsax.pro
 test_zeropoly.pro
-text_write_csv.pro
+test_write_csv.pro
--- a/testsuite/test_write_csv.pro
+++ b/testsuite/test_write_csv.pro
@@ -1,0 +1,8 @@
+pro test_write_csv
+
+a = [1,2,3,4,5]
+b = [6,7,8,9,10]
+
+write_csv,"issue_test.csv", a, b
+
+end
--- a/testsuite/test_write_csv.pro
+++ b/testsuite/test_write_csv.pro
@@ -1,8 +1,8 @@
 pro test_write_csv
 
-a = [1,2,3,4,5]
-b = [6,7,8,9,10]
+if execute(a = [1,2,3,4,5] $
+b = [6,7,8,9,10] $
+write_csv,"issue_test.csv", a, b) ne 1 then exit, status=1
 
-write_csv,"issue_test.csv", a, b
 
 end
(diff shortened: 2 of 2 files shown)


slayoo:
Thank you, yet this branch (brbarker:patch-3) does not contain the ``text_write_csv.pro`` file. You can add the file to this branch by going to https://github.com/brbarker/gdl/new/patch-3/testsuite and clicking on "Add file" in the right hand side of the screen as in this screenshot: 
image

slayoo:
the test was not run because there's a typo: ``text_write_csv``, should be ``test_write_csv``. I've just corrected it.

brbarker:
Thank you. Sorry about that. I'll be more careful.

slayoo:
Now, it is run, but it succeeds (even though the fix is not yet in place):
```
        Start 194: test_write_csv.pro
194/194 Test #194: test_write_csv.pro .................   Passed    0.48 sec
```
How to change the test file so that it triggers the error?

brbarker:
It shouldn't need any changes. The test_write_csv triggers it on my machine. I've attached a screenshot so you can see.
Error_Shot

slayoo:
Ah, indeed, sorry, we need to enclose it within something like ```  if execute(...) ne 1 then exit, status=1```

brbarker:
Okay, you're going to have to help me out here. I'm sorry. I was looking at other bug test files to see what you mean, but I couldn't decide. I've attached a picture with two options of what I think you want. I think it would help if I understood what this `if statement` was doing. What is it testing?
Issue_work

slayoo:
Definitely, the second option is what we need here.
The `IF` checks the return value of `execute` (https://www.l3harrisgeospatial.com/docs/execute.html) which will not be 1 if the  code fails.
Thanks!
Sylwester

brbarker:
Okay, awesome. I just edited the test file that I uploaded so that it is written like the second option.

slayoo:
Still, ``execute`` requires its argument to be a string, .e.g. ``execute("print, 123")``

brbarker:
Okay, I missed the string, but I just added it.

slayoo:
the quotes within string need to be escaped - just corrected it

slayoo:
now in the logs we can see:
```
        Start 194: test_write_csv.pro
194/194 Test #194: test_write_csv.pro .................***Failed    0.47 sec

```
success! thank you!

(the "Failed" should change to "Passed" after we merge the fix)

slayoo:
fix merged: https://github.com/gnudatalanguage/gdl/pull/1179
let's close & reopen to trigger a build with the updated GDL code

slayoo:
```
        Start 194: test_write_csv.pro
194/194 Test #194: test_write_csv.pro .................   Passed    0.46 sec
```
Hurray! Thank you @brbarker

brbarker:
Thank you for your help @slayoo! I see you've been doing work while I was in class today, so I appreciate it! Is there anything else that needs to be done on my end?

slayoo:
Both the fix and the test are now merged and will be part of the upcoming release.
Please do report any other issues you come across, thanks again

Issue #1145, opened 2021-09-23
PLOT: {xy}style=2 is ignored.


jtappin:
The plot procedure ignores the "extended range" flag `xstyle=2` etc. (All the other flags seem to work correctly).
For example:
````
!p.multi=[0,2,2]
plot, findgen(20)                           
plot, findgen(20), xsty=1, ysty=1
plot, findgen(20), xsty=2, ysty=2
plot, findgen(20), xsty=3, ysty=3
````
I'm not sure if it ever worked or not, but in the current git (21 Sep.) , the first & third and the second and fourth plots are identical.

GillesDuvert:
closed by #1190

Issue #1112, opened 2021-09-05
An idea for logarithmic axes.


jtappin:
Plplot is not very clever when labelling logarithmic axes, and puts one major tick of each power of 10, and 9 minor tick marks no matter how big the range. It also (I think) ignores any tick setting options. This is not very pretty when there are many orders of magnitude.

In the Fortran version of Graffer, I use a system of 4 bands where:

- If the range is less than the first limit, Plplot's logarithmic labelling is used.
- In the next band, major ticks are every 2 orders of magnitude with a single minor tick 
- In the third range, major tick are at 5 order intervals with minor tick every factor of 10.
- For very large ranges, the normal annotations for a linear plot are used, but a label format is used

The Fortran code used is:
````
    axsty%is_big_log = .false.

    if (pdefs%axtype(axis) == 1) then               ! Logarithmic plotting
       p10=int(abs(log10(pdefs%axrange(2,axis)/pdefs%axrange(1,axis))))
       if (p10 > axsty%log_bands(1)) then
          axsty%is_big_log = .true.
          options = trim(options)//'o'
          if (p10 < axsty%log_bands(2)) then
             spacing = 2._real64
             nminor = 2
          else if (p10 < axsty%log_bands(3)) then
             spacing = 5._real64
             nminor = 5
          else
             spacing = 0._real64
             nminor = 0
          end if
       else
          options = trim(options)//'l'
          spacing = 0._plflt
       end if
````
The `is_big_log` flag is used to decide whether to call a formatting routine.
The actual formatting is then done by:
````
    if (pdefs%axsty(iaxis)%is_big_log) then
       write(label, "('10#u',i0,'#d')", iostat=ios, iomsg=iom) nint(value)
       if (ios /= 0) label = ''
````
The default settings of `log_bands` that I have used is: `[6, 15, 30]` 

Maybe something similar could be built into GDL.

GillesDuvert:
Thanks James for the suggestion, this is very desirable and can indeed be implemented... At some point, it will be easier to use our own plotting routine than plplot..

GillesDuvert:
I believe this is it!

PR #1190, opened 2021-11-10
WIP Gdl plotting styles

GillesDuvert:
takes into account #1112 and #1145 and more.


GillesDuvert:
reopen, should not fail like this

PR #1195, opened 2021-11-12
Add wrapper function h5g_get_objinfo

ogressel:
See https://www.l3harrisgeospatial.com/docs/h5g_get_objinfo.html for reference.


slayoo:
Thank you!
Let's aim at adding coverage for it within the test suite as well

ogressel:
Thanks for pointing out the unit test. Part of the information (i.e. the file number) returned is not unique, but I guess one can always test for the deterministic entries in the structure.

slayoo:
Thanks! 
For a new test routine to be included in the CI runs, it needs to be placed in a file matching the function name (i.e. test_hdf5_obj_info.pro) and added to the LIST file - as of now, the added test function was not executed on the CI builds.

slayoo:
This is likely not very intuitive, but at least it is somehow explained in the testsuite README file:
https://github.com/gnudatalanguage/gdl/blob/master/testsuite/README
Comments welcome for how to make it more intuitive and robust

ogressel:
Thanks for the polite RTFM. Actually, the other sub-tests are simply called explicitly in the "TEST_HDF5" routine at the bottom. Just forgot to add the call to the new one there. Are there benefits for registering all sub-routines individually in the LIST file? For now, I'll just go with what's been done by the previous authors of the HDF5 module. But happy to change that, if needed.

alaingdl:
No, please do not register every sub-routines, they are supposed to be managed by the main one with the name equal to the name of the code (and the name in LIST)

In the future, we may try to merge, clarify, better organize on a functionality base (e.g. NetCDF functions are tested in several codes) instead of a bug report number. It was convenient long time ago.

Issue #1188, opened 2021-11-09
a='DEADBEEF'XL works in IDL but not in GDL (TEST_ROUTINE_NAMES fails on CI as a result)


slayoo:
```
% Long integer constant must be less than 2147483648.
  At: /home/runner/work/gdl/gdl/testsuite/test_routine_names.pro
% Procedure not found: TEST_ROUTINE_NAMES
```
e.g. here: https://github.com/gnudatalanguage/gdl/runs/4147065434?check_suite_focus=true

slayoo:
@acoulais @GillesDuvert 
I believe, this is a regression.
The minimal reproducer is:
```
GDL> a='DEADBEEF'XL                                          
% Long integer constant must be less than 2147483648.
```
Does it work with IDL?

Also, GDL fails to report which line the error comes from when executing the `test_routine_names` - does IDL tell line no. in such case?

alaingdl:
it is OK in IDL 8.4 on U14.04

```
IDL> a='DEADBEEF'XL    
IDL> a
  -559038737
```

HTH

slayoo:
thanks @alaingdl

GillesDuvert:
a regression indeed.
bug at line 222 of dnode.cpp
This routine forgets about negative numbers ?

GillesDuvert:
solved in #1204

PR #1199, opened 2021-11-16
solve issue #1188

GillesDuvert:
All this is pretty convoluted, as we use more than 1 way to decode decimal/hexadecimal values in all GDL... Ah well..

### DIFF
--- a/src/dnode.cpp
+++ b/src/dnode.cpp
@@ -188,49 +188,43 @@ void DNode::Text2UInt(int base, bool promote)
     }
 }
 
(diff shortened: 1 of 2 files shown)


slayoo:
thank you Gilles, let's remember to - after fixing - bring back to the LIST the test_routine_names which exemplifies the problem

GillesDuvert:
ahem, in fact the template checking the possible overflow of integers, etc in the legacy code (dnode.cpp) is too naive. So I reverted the PR, a better is coming.

PR #1217, opened 2021-12-02
various optimizations, huge changes, WIP

GillesDuvert:
in the course of testing loops speed I found:
1.  that loop speed is not a real problem
2. but that specific algorithms (here is the case for many STRxxx functions such as STRTRIM, but also REBIN, etc...) can be improved
3. but more important, that resorting to openmp in very many places in GDL was not done 100% correctly, inducing noticeable delay in the treatment of small simple operations. 
More specifically, it occured to me that parallelization, using OPENMP and the 'if' clause, was not the best approach in many cases, and was definitely not right in all places.

The if clause was wrong in two ways: 
  1. it would not test if tpool_nthread==1 , where it should just not parallelize at all
  2.  the test on tpool_max_elts is wrong (probably a bad cut and paste that has been propagated everywhere)

Besides, experiences show that it is much faster to avoid entering a #pragma omp parallel section if parallelization is not needed at all.
The main purpose of this PR is institue a new scheme:
 - parallelizing a loop must be done thus:
```
      bool parallelize= yes or no depending on various values, see below
      if (!parallelize) { //which is the most frequent
        for (...) // the loop
      } else { //jump in the same loop, but inside a parallel section:
#pragma omp parallel for num_threads(CpuTPOOL_NTHREADS) if (parallelize) (... eventually other necessary openmp clauses)
        for (...)
      }
```

The "if (parallelize)" above is unnecessary of course but may be handy in case the surrounding 'if (!parallelize) {} else {] ' disappears.

I do not understand why openmp+the if clause is blatantly slower than no openmp at all, but this is the repeated case on my machine, gcc+libgomp 8.4.0 using numerous tests (by adding or removing the #pragma clause etc...)

The minimal test for 'parallelize' is parallelize = (CpuTPOOL_NTHREADS > 1 && nE >= CpuTPOOL_MIN_ELTS && (CpuTPOOL_MAX_ELTS == 0 || CpuTPOOL_MAX_ELTS >= nE));

### DIFF
--- a/README.THREADPOOL
+++ b/README.THREADPOOL
@@ -39,9 +39,9 @@ Geeks corner:
 - some code like where() or interpolate() etc are not with this simple scheme as the section put in parallel is often very long. Usually, for any problem
   size, the speed is on par or better than IDL. Test reports welcome.
 
-- It is up to the user on mumti-core machines to adapt the values of !CPU (using the 'CPU' command) to get the most of the machine. Remember, one
-  can change the !CPU values at each line of code.
-
+- It is up to the user on multi-core machines to adapt the values of !CPU (using the 'CPU' command) to get the most of the machine. Remember, one
+  can change the !CPU values at each line of code, and ThreadPool friendly commands can even be called with one of the 3 keywords, TPOOL_NOTHREAD,
+  TPOOL_MAX_ELTS and TPOOL_MIN_ELTS
 - TRACE_OMP_CALLS (in src/typedefs.hpp) may be used to trace the calls to OpenMP sections.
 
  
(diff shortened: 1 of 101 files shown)


GillesDuvert:
OK, not unexpected failures. WIP.

alaingdl:
How to help you to *test* that ?

Does it means we will have to make some change in the bench ?

GillesDuvert:
@alaingdl well, unless I've accidentaly mistyped something, there should be no difference, except in speed for all functions using !CPU.TPOOL  things, that should be faster for all cases where 
`n_elements is < !CPU.TPOOL_MIN_ELTS`.
My results for 'time_test4' show a GDL on average faster than IDL.
I would suggest to run extensive GDL code in your possession (Planck?), seems to me the best way to cover all my changes. I did that on my side, no probs. (even compilation seems faster, ATV is certainly quicker).

GillesDuvert:
oops, pb in strjoin

GillesDuvert:
I want to merge this PR, as it passes all tests, mine also.
Speed test show a large improvement, GDL is now consitently faster than IDL on all functions described in [README.THREADPOOL](https://github.com/GillesDuvert/gdl/blob/speedups_including_rebin/README.THREADPOOL)

PR #1223, opened 2021-12-07
Toward hdf5 write capability

ogressel:
This is admittedly a bit rushed. Functionality can be tested via, e.g.,

data=[5.,4,3,2,1] & f_id=h5f_create('test.h5') & type_id=h5t_idl_create(data) & space_id=h5s_create_simple([5]) & d_id=h5d_create(f_id, 'a_dataset', type_id, space_id) & h5d_write, d_id, data & h5d_close, d_id & h5f_close, f_id

will add a unit test when I find the time...


GillesDuvert:
There is a problem with the macos builds, apparently a library called by plplot is not present:
```
dyld: Library not loaded: /usr/local/opt/glib/lib/libgobject-2.0.0.dylib
  Referenced from: /usr/local/opt/plplot/lib/libplplot.17.dylib
```
@pjb7687 any idea? probably brew forgot about it ?

ogressel:
Once the build issue on macos is resolved, this should be fine for merging. Will work on "H5A_WRITE" next, which should be very similar. Might push the commits to this branch, too, if we have to wait with the merge anyway...

Oh, I simply assumed that creating an output file in the working directory of the unit test is okay. Please let me know if a scratch space needs to be used, etc.

GillesDuvert:
a dozen of testsuite/ procedure do write things in the working directory, so it's okay.
At some point, all such procedures should be revised to use "[I|G]DL_TMPDIR" as the location for creating files etc.

ogressel:
As a quick recap:

1. writing numeric datasets (byte, int, long, float, double...) should now be supported for scalar and simple dataspaces
2. the same is true for fixed-length string data (with the mentioned IDL quirks [it's a feature, not a bug])
3. the biggest limitation is missing support for writing GDL_STRUCTs (how does one iterate members?)
4. other "todo" items are write capability for opaque datatypes, enums, etc. (is there demand for this?)
5. the functionality for memory and file data spaces provided via keyword params to H5D_WRITE needs to be tested

Other than that, I'd say this should be usable, so soliciting real-world feedback would be good at this point.

slayoo:
@GillesDuvert, any hints how to best iterate over GDL struct members from within C++ code?

slayoo:
@gnudatalanguage/gdlteam, this PR brings long-awaited HDF5 write support to GDL (kudos to @ogressel!), would be great to test it on some existing IDL code to check if the results match, anyone having some HDF5-writing routines at hand that could be used for testing, please let us know. Thanks

GillesDuvert:
windows test fail, this is not normal.

About structs, I do not see exactly what is meant by iterating, please just point to an example of command that should work with a struct?

ogressel:
> windows test fail, this is not normal.
> 
> About structs, I do not see exactly what is meant by iterating, please just point to an example of command that should work with a struct?

Is there an easy way to walk through the tags within a DStructGDL one by one? What I'd need to code is the inverse of 'hdf5_parse_compound', i.e. mapping the tags to "members" in H5T_COMPOUND recursively.

GillesDuvert:
see do_relaxed_struct_assign() in basic_pro.cpp . It explores a struct and compare with another.

Note (I dunno if it is pertinent here) that you can sometimes relax, write a .pro file that does the job between procedure or function parameters, and call that procedure from inside the C++ code.
See how "gdl_implied_print" (a procedure) is called in print_os() in print.cpp

The result of calling a .pro may be slower, but may also be simpler to write...

alaingdl:
same issue in CI for OSx than the previous one, related to plplot in OSX

PR #1224, opened 2021-12-07
this should avoid the "16 enumeration values not handled in switch" …

ogressel:
…clang warning.


alaingdl:
Unclear for me why OSX on CI failed ...

PR #1232, opened 2021-12-18
remove hardcoded extension and paths from FindGLPK.cmake

slayoo:


GillesDuvert:
Apart from now constant failures in MacOS and Windows, that are not related to GDL itself but some install problem---absence of libgobject.dll form plplot?? Again plplot!!! Always plplot!!!
**there is no GLPK on Windows and MacOS** in our builds.
But yes, the patch should pass as it is the right way to find libs in CMake.

slayoo:
Thanks and thanks @dmacks!

PR #1247, opened 2021-12-29
phase out gdl binary symlink creation by CMake

slayoo:


GillesDuvert:
Thanks Sylwester! :+1:

PR #1268, opened 2022-01-22
New attempt plplot driver local

GillesDuvert:
Major changes:
- removed wingcc
- enforce wxwidgets driver on windows
- add a small number of drivers, copied from plplot project, modified by us, to work in all cases with or without the corresponding plplot driver being installed or not by the system.
- these drivers can be adjusted for gdl needs (speed, but also good looking)
- the wxwidgets driver will be the old (deprecated) driver as it is so much faster and supports Hershey fonts.


GillesDuvert:
Needs more work. Removing for the time being.

PR #1278, opened 2022-02-23
Use local drivers if possible

GillesDuvert:
This implements a way to become more independent from the plplot drivers: we just use our owns. At the moment it is a copy of the plplot drivers, as it is just to avoid using the slow and buggy 'new' wxwidgets driver, but the idea is to eventually improve some drivers (the 'mem' driver or the X11 driver could become scalable font aware for example)

The PR removes the obsolete support of the wingdi, removes the 'otherdevices' directory. It installs a copy of the ps,x11,svg,mem and old_wxwidgets (named wwigdetsgdl) in gnudatalanguage/drivers, and, **if possible**, set up the env. var. such as to **force plplot to use these drivers**.

The only visible effects should be:

-  that !P.FONT can now switch from hershey (line) characters to fonts, in the X device;
- that surface,dist(100) will be plotted almost instantaneously, proving GDL effectively uses the fast and accurate old wxwidgets driver.
- and that the plot becomes blank after a change of window size (which is the IDL behaviour). Something we may improve in the future.


alaingdl:
Compiled on a Debian 10. I cannot have plots : 

* at starting : 

*** PLPLOT ERROR, ABORTING OPERATION ***
plInitDispatchTable: Could not open drivers directory, aborting operation

* after starting : 
```
GDL> set_plot, 'x'
GDL> plot, findgen(10)
% PLplot installation lacks the requested driver: wxwidgets
% Execution halted at: $MAIN$          
```
* I also tested SVG, same problem : 
```
GDL> set_plot, 'svg'
GDL> plot, findgen(10)
*** PLPLOT ERROR, ABORTING OPERATION ***
plInitDispatchTable: Could not open drivers directory, aborting operation
% PLplot installation lacks the requested driver: svg
% Execution halted at: $MAIN$          
```

HTH

A.

alaingdl:
OK. Worked after "make install". No clear change in perf :disappointed: 
Maybe I need to flag something at compilation ?

GillesDuvert:
Yes, it MUST be installed, same as when a .pro has been changed.
However, it is protected against not finding the driver in the 'gnudatalanguage/drivers' directory.
In this case it reverts to the default wxwidgets driver.
That apparently it does not find either, so this is a (minor) bug.

GillesDuvert:
Concerning the performance, I suspect your mileage varies, as the 'new' wxwidgets driver speed was OK with a previous version of the GTK, and that was linux only. Windows and Mac did not seem affected.
Anyway, forcing the use of 'ours' drivers is a step towards more stability (admitedly at the expense of maintaining them, in the absence of maintainance on the plplot project's side) 
You can check that it is 'our' driver is used: when it is the case, on startup GDL says:
`- Using local drivers in /usr/local/share/gnudatalanguage/drivers`
Typically, 'our' driver gives:
```
GDL> tic & surface,dist(100) & toc
% Time elapsed :       0.17748594 seconds.
```

Jeanne-jw:
Hello, it's great that we have succeeded to compile this version of GDL on my Mac M1 ( macOS Monterey). 
```
GDL> tic & surface,dist(100) & toc
```
It takes similar time with the version of GIT today.

PR #1279, opened 2022-02-25
Change in REBIN() and test_rebin

Jeanne-jw:
1. I made some changes in the order of calculation in datatypes.cpp, where there are functions for rebin(). 
2. The old test_rebin.pro isn't OK with IDL, because the results of GDL and IDL for one command are different. Now this bug is fixed.

### DIFF
--- a/src/datatypes.cpp
+++ b/src/datatypes.cpp
@@ -3902,8 +3902,6 @@ BaseGDL* Data_<SpDComplexDbl>::Rebin( const dimension& newDim, bool sample)
 
 // rebin over dimIx, new value: newDim
 // newDim != srcDim[ dimIx] -> compress or expand
-//#define tic      double tic_t = clock();
-//#define toc      std::cout << (clock() - tic_t)/CLOCKS_PER_SEC << " seconds" << std::endl;
 template< typename T>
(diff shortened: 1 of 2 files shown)


GillesDuvert:
Dear @Jeanne-jw , excellent! 
Could you however remove the statements `SizeT count_t = 0;` and the commented test lines (`//tic //toc //std::cout...` ) in order to keep the changes to the minimum? Thanks.

Jeanne-jw:
Ok ! No problem. Sorry to have left it there.

Jeanne-jw:
#### Notes:
1. Complex, double complex and string are not usable, in both IDL and GDL.
2. For float data, the speed has been increased to the same level of IDL, no matter for expand or compress/ sample or not sample. 
3. For int / long data, when **new dimension < original dimension**  or when **sample=1** the new version runs great. While when we try to expand a matrix in not sample way, it runs as slow as the old version. As shown in the ***right part of the first image*** (y : how long time it takes; x : different steps), the green line is much higher than the blue line. Need to continue in the future. 
speed_rebin_idl_vs_gdl

alaingdl:
It would be great to add a bench based on that around.

PR #1284, opened 2022-03-02
Remove shared static ints plotting

GillesDuvert:
They induce unwanted effects in all plotting options (xrange, etc.)
As a bonus, the number of ticks and subticks in the 'automatic' case should be more in line with IDL's choices.


GillesDuvert:
merged as necessary for 1.0.2

PR #1285, opened 2022-03-03
Add hdf5 compound write support

ogressel:


slayoo:
Thank you, @ogressel !!!

slayoo:
@pjb7687 Jeongbin, do you have any hints on what could be causing the failures on Windows?
The master branch seems to pass windows builds
Thanks

pjb7687:
@slayoo I've just done the first investigation. It breaks at `dinterpreter.cpp:1178`
```
(gdb) bt
#0  0x00007ff7c952cd38 in __cxa_throw ()
#1  0x00007ff7c8f5cbd9 in DInterpreter::ExecuteLine (this=0x2c699fef30, in=0x2c699fe320, lineOffset=0)
    at D:/a/gdl/gdl/src/dinterpreter.cpp:1178
#2  0x00007ff7c8f5dd53 in DInterpreter::ExecuteFile (this=0x2c699fef30, file="test-angles.pro")
    at D:/a/gdl/gdl/src/dinterpreter.cpp:1574
#3  0x00007ff7c8f5e2b7 in DInterpreter::InterpreterLoop (this=0x2c699fef30, startup="",
    batch_files=std::vector of length 1, capacity 1 = {...}, statement="")
    at D:/a/gdl/gdl/src/dinterpreter.cpp:1709
#4  0x00007ff7c94c2590 in main (argc=2, argv=0x147080819a0) at D:/a/gdl/gdl/src/gdl.cpp:525
```
Interesting, it doesn't seem to be related to this patch.

pjb7687:
And here is the exception:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::substr: __pos (which is 4294967295) > this->size() (which is 2)
```

PR #1291, opened 2022-03-25
Histogram() & test_histo

Jeanne-jw:


alaingdl:
Hi @Jeanne-jw ! You clearly made _several steps forward _ in the HISTOGRAM() story !

If I understood well, we just forgot to re-run the whole test suite ! And we did not check cases where HISTOGRAM() should failed, what is revealed by old forgot test ! (test_bug_2876372.pro)

I will look at that if I can correct that easily.

alaingdl:
OK. See from my side, bug in test_bug_2876372.pro is quite obvious, since binsize should be strictly greater than zero. (line ~1490 in gsl_fun.cpp)

But ! At the same time, another test is not OK anymore ! test_bug_2846561.pro

alaingdl:
Great ! all lights green !

Nevertheless we have to consider we don't have enough tests on tricky cases in HISTOGRAM,
the problems triggered here were outside test_histo.pro !

GillesDuvert:
I notice a use of `if (debug)` inside loops. Hopefully they will be filtered out by the compiler if `debug` is declared `false`. But for a release and to be on the safe side they should preferably be commented out or inside `#ifdef` constructs.

PR #1296, opened 2022-03-28
HISTOGRAM() for over limit situations

Jeanne-jw:
example of these situations :
The user inputs an array in type of Byte, and inputs max/min/binsize bigger than 255 or less than 0.

### DIFF
--- a/src/gsl_fun.cpp
+++ b/src/gsl_fun.cpp
@@ -129,19 +129,6 @@ namespace lib {
 
   };
 
-  bool IsInteger(DType pType)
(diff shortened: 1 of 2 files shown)


alaingdl:
Hi @Jeanne-jw . I am happy for sure that you add some tests and located one more tricky case in this complex piece of code HISTOGRAM() !
But I am not super happy with this PR because it seems that you revert some codes I removed during the Weekend. Bu I will carefully check again, and may-be we will re-do that tomorrow ?

PR #1297, opened 2022-03-30
TVRD() and TV speedups. Plus a few improvements on the widget's side

GillesDuvert:
add gdl icon for wxWidgets windows ( temporary icon of course, to be decided upon at some point)
align focus at creation of a plot window or a widget to IDL's behaviour (was not the case before).
reinstate normal menubar for linux and Windows, leaving taskbar+buttons for MacOSX (to be commented, I believ this has more advantages)
large speedup of "TV" for wxwidgets using specialized wxwidgets functions
large speedup for "TVRD" avoiding copie of large bitmaps when possible
improved bitmap<->gdl array copy/transfer loops by doing algebra for the compiler.


GillesDuvert:
Unfortunately Windows and Mac do not work correctly with these changes.

GillesDuvert:
Windows and MacOSX Catalina OK on my side (I always have -DPYTHON=OFF) --- Windows build has problem with PYTHON?

PR #1304, opened 2022-04-08
fast generic HISTOGRAM function

GillesDuvert:
rewrote histogram for speed and use templates for calculations in each types.
removed histogram from gsl_fun as it does not use gsl anymore.
added a few supplementary tests.


alaingdl:
just to mention that, yes, the speedup of performances for the keyword REVERSE_INDICES are very good and in line with the IDL ones : 
```
GDL> a=randomu(seed, 2e7)
GDL> tic & b=histogram(a, re=re) & toc
```
Before change, 10 sec., now 0.17 sec on same machine :smiley:

PR #1328, opened 2022-05-15
Fix intro typographical errors.

imoverclocked:
While the "early noughties" does have a little more pizzaz, let's stick to rough time periods in the intro.


slayoo:
Thanks. The first commit dates back to 2004-12-09 - this is not "early nineties", right?

imoverclocked:
til: noughties is actually a valid reference to the zeroth decade.

PR #1337, opened 2022-07-03
Remove deprecated std::unary_function and std::binary_function

jkohnert:
std::unary_function and std::binary_function are deprecated in the C++11 standard. This PR changes those to C++11's std::function instead.

Additionally, there are some updates to GitHub actions workflow:
1. Fix caching/restoring deps jobs, I could not get the MacOS build job to work otherwise
2. Fix compile definitions fpr one file in the debug windows build by adding "-Os" to its compile flags, since I got a build error on this.


GillesDuvert:
Many thanks, @jkohnert

PR #1338, opened 2022-07-06
3D Support and more, WIP

GillesDuvert:
**WARNING** This is a huge revamping of the graphics part of gdl.

To reproduce exactly IDL's 3D behaviour, I have hacked the plplot DRIVERS. Because the GDL 3D plots are just any plot, put in 3D perspective by the !P.T 4x4 universal transformation matrix, something plplot does NOT know about.
Hacking plplot drivers is possible because we now provide and "install" our own drivers, which are no more just a copy of the ones by the plplot project (not mentioning solving a lots of bugs in passing).
A benefit is also that we can now solve long-hanging problems like the PostScript QUALITY, you'll see that the PS outputs are now of excellent quality. Previously they were stuck in low-res at 72 DPI.

This revamping goes a long way in the direction of being independent of the plplot library maintainance and installation hazards, another step that will permit to solve other problems, plplot not always doing what we would like.
An other important change in GDL graphics is that it only relies now only on the graphic setup expressed by the !D,!P,!X,!Y and !Z values (so, they should now be OK, and procedures based on them should finally work), and *almost* all the plotting is now done avoiding the awkward plplot vpor() and wind() functions.

Next, changing the internal logic of GDL graphics permitted to find where several of our procedures, especially those related to projections (using the proj library) were inadequate. Although not entirely complete, now in all practical purposes all our projection-related procedures work OK, like plotting contours on an image projected with, e.g., a truncated projection like the Goode Homolosine.

Finally a set of bug corrections and improvements found during all the test phases are also registered in this PR.

I urge all good people to perform tests on this version and give a feedback. Simple, and many not so simple, tests were made and succeded, but I cannot cover everything.


GillesDuvert:
closing and restarting to check influence of https://github.com/gnudatalanguage/gdl/pull/1337

alaingdl:
Hi (sorry for the delay, quite busy with JWST data :smile: )

Tested on local machine running Ubuntu 22.04, compile with the script :smiley:

Good new : 
- I succeed to compile this PR 1338
- most of the graphical tests in Coyote are OK

Small issue : 
- the size of few symbols seems to be slightly bigger in GDL than in IDL (e.g.  
    basic_line_plot & plot_with_legend in Coyote)

Bad news : 
- problem with AXIS (test_axis &   additional_axes_plot (Coyote) are not OK)
- intermediate plots in IDL graphics_times1/2/3 are not shown. perfs. are not good for surface & shade (know problems)

GillesDuvert:
Thanks @alaingdl 

As always, there is something in the external libraries or installs that create discrepancies between my personal experience and those of other users.

Concerning the performance of surface and shade, even in its best moments, plplot alone is 2 times slower than ID, then the X11 or wxWidgets driver seem to be another 4 and 2 (respectively) slower. IDL takes 0.25 sec to plot a 256^2 surface. If you experience more than that, so, 1 to 2 secs then it is related to the [plplot issue](https://sourceforge.net/p/plplot/bugs/199/) 

AXIS: to be patched of course.

There are issues about the refreshing of plots with wxWidgets, too much slows down the plot, too little and nothing happens. This because the wxWidgets eventloop is not separate from the GDL eventloop, but cannot be separate as we need all events to be reviewed first by GDL. In the absence of a wxWidgets guru in the contributors this situation will probably not improve.

alaingdl:
Hi Gilles

Take care that in the new version we have a regression with the xstyle=8 (top line should not be displayed)

`plot, findgen(5), xstyle=8`

(maybe corrected in your last versions, but not on the version I tested)

I also have serious troubles with AXIS but I need time to clarify in my large code ...

A.

GillesDuvert:
Still impossible to get plplot in the windows version to find the drivers. This is insane.

GillesDuvert:
OK, actually the WINDOWS version of plplot seems to be able to use PLPLOT_DRV_DIR (contrary to my experience) to find our modified (for 3D and more) drivers. It may just be possible that the installer needs only to define PLPLOT_DRV_DIR to C:[..]\gnudatalanguage\share\gnudatalanguage\drivers . Keep fingers crossed.

GillesDuvert:
nope, is the (correct) env. var misinterpreted as plplot does not handle windows paths?

GillesDuvert:
Very Very Very frustrating.
with the current patch in build_gdl.sh in plplot, GDL will compile and run OK wth a plplot having dynamic drivers and OUR drivers, the same a for MacOSX and Linux. But it needs to be compiled **with build_gdl.sh and on the spot.**
Impossible to get a GDL transported on a new machine using the installer. Plplot does not find the drivers because of 1) not using env. var correctly under windows and 2) using bad filenames (unix vs. windows).
Very tricky installer problem.

GillesDuvert:
OK, this was nightmarish.
For the record, plplot uses getenv( PLPLOT_DRV_DIR ) to find where **driver description** and driver executables are **located** (and abort if the driver text description is not found). This when the DYNAMIC drivers is enabled of course. The loading of the driver .so is done on linux and Mac in this directory too.
But on windows, the routine used to **load** the drivers will look **only** on the system/user `PATH` or (even better) in the same directory as the executable. **Not** on PLPLOT_DRV_DIR.
So to get the dynamic drivers on Windows, now mandatory as we have improved them for 3D support, the `build_gdl.sh` copies the drivers in _the same directory as `gdl.exe`_ , so:

- In case of a `gdl` compiled by the user, using `build_gdl.sh`, the location of the driver **description** is built-in, the driver copy is in install/bin, everything is OK.
- For a gdl **installed from the installer**, the installer defines or updates PLPLOT_DRV_DIR in the table of system variables to have plplot find the **driver description** and not crash. The driver itself will be loaded from the gnudatalangage/bin directory.

Meaning that a user build of gdl on windows CANNOT BE RELOCATED easily, better create an installer using` build_gdl.sh pack`

PR #1358, opened 2022-08-10
Solve 1206

GillesDuvert:
should be OK with 16 bit aware versions of IM and GM.


alaingdl:
Just for the record.

IM was the initial choice for JPG, PNG ... in GDL long long time ago. Some years ago IM was strongly "unstable" (interface and behavior (basics stuffs ok for version x but broken for version x+1 ...). (We don't have to be the end-users testers of internal changes in IM.) This is why we adopted GM, much more stable, reliable.

If we decide to "eject" one of the two, please keep in mind this  very important argument.

GillesDuvert:
We need to choose, since maintainance is indeed a problem. I would vote for the one that has simplest 16 bit support

Issue #1368, opened 2022-09-01
no more iconify button in Window (u2204 + xfce4)


alaingdl:
furthermore the list of opened windows is no more visible in the "desk" 

all was OK few weeks ago ...

A.

(old behavior, was OK for me)
![old_4buttons](https://user-images.githubusercontent.com/38264661/187953064-b431e1e3-fae7-488b-a4c2-246746cdb437.png)

(new current behavior : not OK for me)
![new_3buttons](https://user-images.githubusercontent.com/38264661/187953058-8f0dc289-827d-4ab6-beea-2a2c9f41622e.png)

GillesDuvert:
wxWidgets either allow the 'old  behavior' , but you **lose focus in your terminal** if you enable that, or make the new behavior and the focus stays in the command line.
This may well depend on the window manager, and probably wxWidgets could do better.
I chose to have the focus stay at the command line.
If anybody knows how to programmatically put back the focus in the command terminal, now with all those new fancy display managers, just tell me.

alaingdl:
In the current Git version, compiled with the script (then with WX on), this is back to OK.
Then I think we can put that to _close_

@GillesDuvert I don't know what is you mention (_focus in your terminal_)

PR #1385, opened 2022-10-25
Speed plots

GillesDuvert:
Should provide a speedup for all plot, oplot and plots command. Corrected a few discrepancies with IDL (/continue in PLOTS, COLOR option cannot be an array for PLOT and OPLOT, etc)

**This is sensitive, please exercise this update as there may be some bugs introduced by the large refactoring of this function.**


alaingdl:
I tested this branch compiled & installed using the script. No special problem. A message 
```
The following tests FAILED:
	101 - test_hdf5.pro (Failed)
```
but it is not related to this branch.

* Problem 1 : severe
Locally (U2204) and remotely (Debian 11), when just making `GDL> plot, findgen(10)`
a window opens but no plot is done. Second plot is OK. An oplot of first plot is done (does appear on the display).
(I have no g/idl_startup)

* Problem 2 : severe
Another problem I already mentioned is that the created window does not appear any more in the "tableau de bord" of XFCE window manager #1368 . Then the opened windows are hidden by others windows

HTH

PR #1427, opened 2022-12-08
Add hdf5 group member related functions

ogressel:


src/hdf5_fun.cpp
```diff
@@ -1075,6 +1075,76 @@ hid_t
...
+    e->AssureLongScalarKW(2, index);
+
+    /* use 'H5Giterate()' to get member name length */
+    size_t name_len;
+    if ( H5Giterate( loc_id, name.c_str(), &index, get_len, &name_len ) < 0 )
+       { string msg; e->Throw(hdf5_error_message(msg)); }
+    index--;
+
+    /* allocate string buffer (remains allocated) */
+    char* member_name=static_cast<char*>(calloc(name_len+1,sizeof(char)));
```
slayoo, line 1138: Thanks @ogressel! One question: is it ensured that DStringGDL destructor will call free on the member_name? I'm trying to check it in `src/datatypes.hpp` (looking at `Data_`, looking for `free`) but cannot find an answer...

ogressel: @slayoo Thanks for pointing this out. I adopted the idiom for creating a string object from a non-constant string from elsewhere, but was unaware that the "(remains allocated)" is in fact unintended behavior.

Issue #1433, opened 2022-12-15
Regression in basic PLOT + xrange


alaingdl:
I discovered that incidentally because I am exploring block by block a time series with thousands of points.

in current git version, if I do :  offset=2400 & plot, b, xrange=[offset, offset+24*24-1]
the x-axis does start at 2200 ! It should start at 2400 ! (as it did till before Dec 6, 2022)

```
GDL> plot, randomu(seed, 100000) , xrange=[2400, 2400+24*24-1]
GDL> print, !x.crange
       2200.0000       3000.0000
```

IDL and GDL prior Dec 06 give : `2400.0000 3000.0000`

I notice that we have to rework and merge two old fashioned tests which should detect such problems :
_test_plotting_ranges.pro_ & _test_plot_ranges.pro_
I can do that.

alaingdl:
In PR #1434 I add some cases but they are not working with current git version ...

GillesDuvert:
ALL should be OK now with https://github.com/gnudatalanguage/gdl/pull/1437

Issue #1432, opened 2022-12-14
plots reports erroneously "Not enough valid and unique points specified" 


brandy125:
This simple plot generates an error message different to IDL where it works normally:
```
GDL> a=[[0,1,2],[2,3,4]]
GDL> plot,findgen(10)
GDL> plots,a
% PLOTS: Not enough valid and unique points specified.
% Execution halted at: $MAIN$          
GDL> 
```

alaingdl:
Thanks for this bug report, confirmed here.

See from here, this bug appeared after Sep 26 and before Oct 18. HTH 
Unfortunately most of the plotting_*pp & gdl*stream.?pp have been changed at that time !

Could we consider to have some plotting tests in Z mode to detect such problems ?

A.

alaingdl:
In fact TEST_PLOT_USERSYM is also broken since this time

GillesDuvert:
ALL should be OK now with #1437

GillesDuvert:
... And thanks for the tests!

GillesDuvert:
Argh corrected the TEST_PLOT_USERSYM problem but not the "Not enough valid and unique points specified".
One problem, one issue...

GillesDuvert:
thanks for your patience!

brandy125:
Thanks to you!!!

PR #1437, opened 2022-12-18
Patch for #1433 #1432 and test for wrong install of drivers.

GillesDuvert:
(edited): initially motivated by #1433, but also solve the problem with **test_usersym** and #1432

#1433 : hm, the algorithm to get the same automatic ranges as IDL is bit more… complicated than expected.

Indeed, we need to insure that the !X or !Y.CRANGE, computed by FLOOR and CEIL on the (double precision) [X|Y]RANGE values do not 'jump' to the next interval due to numeric precision problems. I note that IDL behaves differently in this respect when the  [X|Y]RANGE is explicitely set to double precision: b=randomu(seed, 100000) & offset=2399.999d & plot, b, xrange=[offset, offset+24*24-1] will set lower bound to 2200 b=randomu(seed, 100000) & offset=2399.999 & plot, b, xrange=[offset, offset+24*24-1] will set lower bound to 2400

also, found a bug for PLOT,[0] where X.CRANGE was [0,0] instead of [0,1]

### DIFF
--- a/src/plotting.cpp
+++ b/src/plotting.cpp
@@ -278,6 +278,11 @@ namespace lib
         max = 0.98 * val_ref;
         min = 1.02 * val_ref;
       }
+      if (max==min) { //case "PLOT, [0]"
+        max=1;
+        min=0;
+      }
+      range = max - min;
     }
 
     if (range >= 0) {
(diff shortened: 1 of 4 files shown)


alaingdl:
Hi. Thanks to the correction.

On my side, before this PR, with the script, when I run "./scripts/build_gdl.sh test" this test not fails
But when I call it at the GDL> prompt or, in the testsuite dir with ../install/bin/gdl -e test_plot_ranges
no problem (no errors)
 
I do not see/understand why !
What is the trick ?

GillesDuvert:
> Hi. Thanks to the correction.
> 
> On my side, before this PR, with the script, when I run "./scripts/build_gdl.sh test" this test not fails But when I call it at the GDL> prompt or, in the testsuite dir with ../install/bin/gdl -e test_plot_ranges no problem (no errors)
> 
> I do not see/understand why ! What is the trick ?

I have seen many times cases where a test works when called directly, but fails when called in 'make test' (or "./scripts/build_gdl.sh test" that amounts to the same I hope). No clue..

.. Although several libraries, and especially the plplot drivers, must be installed at a specific location before use. Using "make test" one may use an installed driver that is not the one just compiled by "make".
This because GDL does not look for dependencies in relative (to the location of the executable) locations but in absolute ("/usr/local/share/gnadatalanguage..") locations.
This is annoying and I believe "build_gdl.sh test" should test the **installed** gdl (following  "build_gdl.sh install") and not the built gdl as it does today.

alaingdl:
I take some time to test and all seems to be OK now :smile: 

2 points : 
- you mention 
 `b=randomu(seed, 100000) & offset=2399.999d & plot, b, xrange=[offset, offset+2424-1]` will set lower bound to **2200** but I cannot reproduce that in IDL 8.8 & GDL  (both cases (f or d) yeld to 2000 -- 5000 . If I add /xstyle, the exact ranges are used, same in IDL & GDL (!X.crange :        2399.9990       4822.9990)

- concerning the tests in CI and the use of "our" driver, maybe it is just related to the fact that the (internal) path to the lib (*.so) are not well set through Cmake ?

GillesDuvert:
@alaingdl
yes, strange, I cannot reproduced it just now, as I wrote it.
But I have
```
IDL> plot, randomu(seed, 100000) , xrange=[2399.999, 2400+24*24-1] ;  -> [2400-3000]
IDL> plot, randomu(seed, 100000) , xrange=[2399.999d, 2400+24*24-1] ;-> [2200-3000]
```
go figure.

PR #1446, opened 2022-12-21
Implement .OUT command

GillesDuvert:


brandy125:
@GillesDuvert Does the .out command not work any more?
Could you check please?

PR #1449, opened 2022-12-22
Add hdf5 group link related functions

ogressel:
This finishes the implementation of the "H5G_..." category of HDF5 functions. As far as I can see, these should all be complete in terms of their functionality, as well.


slayoo:
Thank you, @ogressel!

PR #1454, opened 2022-12-24
Add hdf5 identifier related functions

ogressel:
Only two functions in this category (H5I_...), and trivial to implement, too.


slayoo:
Thank you @ogressel !

Issue #1461, opened 2023-01-01
first basic plot is not done with wxwidgets on U22.04


alaingdl:
This bug is reproducible on U22.04 (gnome or xfce), U22.04 (gnome), and remotely (ssh -X) on debian 10 & 11.
(I can provide on demand _dpkg_ list of any other packages
libwxbase3.0-dev:amd64 3.0.5.1
libwxgtk3.0-gtk3-dev:amd64    3.0.5.1+dfsg-4
plplot-driver-wxwidgets:amd64 5.15.0+dfsg-24build2

This bug was present in package 1.0.1 in Debian, and it still not working well. For me, as long as this annoying bug is not solved, we cannot deliver a 1.0.2

How to have it : 

* compile (git) GDL using the script + install it
* run it with no option (default is wx ON) then try to plot : plot, findgen(10) : window without plot for me (not OK)
(a second call is OK)
* run it with the option (gdl --no-use-wx) then try to plot : plot, findgen(10)  : plot OK for me

Please report here your configuration (Linux version + window manager, thanks) and the status of  this bug. We have to solve it.

slayoo:
I confirm, same behavior on my laptop: no plot after a fresh compilation + "sudo make install"
![image](https://user-images.githubusercontent.com/89685/210176874-5b95934b-8d09-4e42-878c-a2ad31c89efb.png)

(Debian Linux, second call gives a plot)

slayoo:
Adding `--no-use-wx` doesn't change anything - same behaviour

GillesDuvert:
Very annoying as its ok on some distros. Probably due to the plot window not receiving the first draw event from plplot. Stay tuned.

GillesDuvert:
> Adding `--no-use-wx` doesn't change anything - same behaviour

Sylwester you did not compile gdl with option -DX11=ON

slayoo:
```
slayoo@aery:~/devel/gdl/build$ cmake .. -DPYTHON=OFF
-- INFO: will use GNU extensions for STDIO (useful for compressed I/O) since it seems accepted by your c++ compiler.
Use fuzzy detection for PLplot lib. (e.g. in /usr/lib)
-- Using a plplot library without private functions - UGLY workarounds will be used.
-- INFO: We prefer to use GraphicsMagick than ImageMagick
-- Looking for grib_index_new_from_file in /usr/lib/x86_64-linux-gnu/libeccodes.so
[...]

PR #1462, opened 2023-01-02
solve #1461 ?

GillesDuvert:
should permit to avoid erasing the first plot because of subtle differences in size between surrounding frame and draw panel at creation.

Please test.

### DIFF
--- a/src/devicewx.hpp
+++ b/src/devicewx.hpp
@@ -175,23 +175,25 @@ if(hide) {
   plotFrame->Refresh();
   plotFrame->Update();
   plotFrame->Raise();
-  //really show by letting the loop do its magic. Necessary.
-#ifdef __WXMAC__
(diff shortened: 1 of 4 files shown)


alaingdl:
Great ! Clearly a step forward in the right direction. But here the _location_ of the plot (or the surface) is not good on (no numbers visible on xaxis, no effect if xtitle provided ...) When plotted with `gdl --no-use-wx` plots are OK

GillesDuvert:
cannot reproduce (as always)
what configuration?
We'll definitely will stick with X11 on linux, as this becomes a farce.

GillesDuvert:
Basically this comes from the size of the frame not being the size of the draw area, but easier to say than to find the good way out.

GillesDuvert:
No this is NOT what to do.

Issue #1463, opened 2023-01-02
OSX widgets cannot get the focus


GillesDuvert:
For some undocumented problem, GDL widgets with the current version and under MacOSX do not get the focus, so the keyboard events are sent to the terminal and not the widget.
In other terms, you cannot interact with a text widget (or anything needing to catch a keyboard key) for GDL under OSX.
This obviously has to do with the way OSX manages the focus, but impossible to find a correct answer and solution in the meagre documentation of wxWidgets. Probably related to the absence of 'main' wxApp in the OSX version of GDL.
Help badly needed.

GillesDuvert:
solved in #1464

Issue #708, opened 2020-02-05
it is not possible to open a window with a size bigger than the screen size


brandy125:
My monitor has a resolution of 3840 x 2160.

The command  GDL> window,1,xsize=8192,ysize=8192

opens a window with size 3840 x 2160 instead of 8192 x 8192

alaingdl:
It is a feature ;)

I would no say it is a bug, it's the normal way, IDL 7x & 8x gives same behavior (the documentation
[WINDOW_Procedure](https://www.harrisgeospatial.com/docs/WINDOW_Procedure.html) is not clear on that)

Maybe using sliders (?) in Widgets you can circumvent that ?

brandy125:
Hi alaingdl,

 I am working since many years with fvwm2 window manager and since I can remember (since IDL 5.2) it is possible to open a window bigger than the screen size and I really need this.

I would say that this behaviour is still a bug since not even sliders appear. The opened window has no sliders and appears as a window opened with GDL> window,1,xsize=3840, ysize=2160 

Do you agree?

Do you think it is possible to fix this?

brandy125:
Here another behaviour that is not correct and related to this:

GDL> window,1,xsize=8192,ysize=1024,/pix
GDL> plot,findgen(100)
GDL> a=tvrd()
GDL> help,a
A               BYTE      = Array[3840, 1024]

That is not correct.

GillesDuvert:
on a machine with a big screen I have
```
IDL> window,1,xsize=8192,ysize=8192
IDL> a=tvrd()
IDL> help,a
A               BYTE      = Array[3440, 1387]
```
and GDL does the same. So the first issue behaviuor is just like IDL does, that is what GDL is supposed to do. 
But, indeed, the /PIX option changes everything, and /PIX should open a non-mapped window of any size supported by the system, which GDL does not. Yet Another Bug.

brandy125:
I see that beside the BUG there are two different behaviours depending on the used window manager.
On my Suse 15.1 with fvwm installed I get with IDL 7.01:

IDL> window,xsize=8192,ysize=8192
IDL> a=tvrd()
IDL> help,a
A               BYTE      = Array[8192, 8192]
IDL> 

If I start with GDM window manager I get 

IDL> window,xsize=8192,ysize=8192
IDL> a=tvrd()
IDL> help,a
A               BYTE      = Array[3840,2160]
IDL> 

[...]

PR #1464, opened 2023-01-02
may well be the culprit of #1461

GillesDuvert:


GillesDuvert:
please test this (clever?) workaround.

alaingdl:
Same situation see from here (u2204 & Debian 10)

BUT I may have some clues for you : `window & window, 2` gave 2 windows of different sizes
Same if : `plot, findgen(10) & window,2` or `window, 0 & plot, findgen(10) & window,2`

If I call again `window, 0` the new generated window does have the same size than window 2
(and is higher that first window 0)

Another problem : `print, !d.x_size, !d.y_size` gives same size but windows don't have same size !

```
GDL> window, 0
GDL> print, !d.x_size, !d.y_size
         960         540
GDL> window, 1
GDL> print, !d.x_size, !d.y_size
         960         540
```

HTH

alaingdl:
Pfu, long time I don't used that : xwininfo

First window is :   Width: 950  Height: 506
Second window is :  Width: 960 Height: 540

What I did : 
```
GDL> plot, findgen(10)
GDL> window, 1,xs=100, ys=200
GDL> window, 2

and in another window :
$ xwininfo -tree -root | grep GDL
 
        0x300003b "GDL 2": ("gdl" "Gdl")  960x540+5+29  +6+56
        0x300002a "GDL 1": ("gdl" "Gdl")  100x200+5+29  +1815+875
        0x3000013 "GDL 0": ("gdl" "Gdl")  950x506+5+29  +955+56


```

and the same with `gdl --no-use-wx`
```
        0x5000011 "GDL 2": ()  960x540+5+29  +6+56
        0x500000a "GDL 1": ()  100x200+5+29  +1815+875
        0x5000001 "GDL 0": ()  960x540+5+29  +955+56
```

alaingdl:
I spend some time making `cout` in gdlwxstream.cpp, devicewx.hpp plotting_windows.cpp

Comments !
-- the sizes (X,Y) of the windows are OK (values as expected) in every WOpen() I checked
-- I put a sleep before WOpen, clearly the window is "plotted" 2 times
-- When copying the first window with w0=TVRD() and doing a TV, w0 in a wider plot, the whole plot is OK (xaxis are here)

 question : 
-- why when calling just `plot, findgen(10)` without having a window before the code don't go through the 
window() as defined in  plotting_windows.cpp ? (then I began to say the issue comes from plplot ?!)

some #define gdlBORDER_ are defined in gdlwidgets.hpp. Any chance they interact with the first plot (then being overwritten by our values ?)

GillesDuvert:
Hm all this seems related with windows that need to produce a resize event when they are resized (since you need to have your plot resized if you resize manually the window, there is even a special case where the window is larger than the screen size see #708). Unfortunately creating the window in the first place produce sizes events (even a lot of size events on gnome (?)) which 1) may 'clear' the window and 2) give the size of the window+decoration, but not the size of the plot window itself, and decoration sizes are not well known by wxWidgets.

Basically this behaviour has not changed since a few years so I expect this size discrepancy to be there in gdl 0.1.x . Besides, it works the same on OSX and Windows, and that never caused any problem (?)

I guess I'll have to install an Debian 10 VM to check this, as I do not have all those pesky problems. 

BTW, you refer to window size problems, but is the first bug (first plot not seeen) still present?

GillesDuvert:
OK, installed a Debian 10 in a VM.
With exactly this PR's code, indeed the first plot is not seen until one clicks back (gives focus back to) the terminal.

GillesDuvert:
Normally this should solve the 'first plot not seen' problem.
Then the size of windows, if persist, can be solved easily I think.

GillesDuvert:
Actually, not quite.
There are some side-effects.
Basically the wxwidgets eventloop is not always present to handle the resizes, or refresh.

GillesDuvert:
The test fails on ubuntu (no display) but the PR is OK on ALL platforms, solving #1461 but also #1463 .
... And installers are OK too.

alaingdl:
I am not sure to understand what I have to check !
OK, compiled on U22.04.
The first plot is now OK
(but the size of the window for the first plot is wrong)

Now testing on OSX

alaingdl:
On OSX, I got :

```
 In file included from /Users/alaingdl/GDL/gdl-1.0.1-git230105CMake/src/graphicsdevice.cpp:34:
In file included from /Users/alaingdl/GDL/gdl-1.0.1-git230105CMake/src/deviceps.hpp:22:
/Users/alaingdl/GDL/gdl-1.0.1-git230105CMake/src/plotting.hpp:258:54: error: unknown type name 'PROJTYPE'
  DDoubleGDL* GDLgrGetProjectPolygon(GDLGStream * a, PROJTYPE ref, DStructGDL* map, DDoubleGDL *lons, DDoubleGDL *lats, DDoubleGDL *zVal, bool isRadians, bool const doFill, bool const dolines, DLongGDL *&conn);
                                                     ^

alaingdl:
Incredible !! On OSX, this PR, with the script + install : launching is OK, then plotting is OK _and the size of the windows is OK !_ Marvellous !! (but why my U2204 is broken ?!)

GillesDuvert:
Works on Debian 10 (on a VM here), my Mageia (of course), Windows 10 and OSX...

Issue #1465, opened 2023-01-06
A problem with idlneturl__define.pro (spawn, cmd, result, blahblah)


Albom:
An error occurs in **idlneturl__define.pro** file (**line 134**).

Trying:
```idl
spawn,'curl -LI --silent --show-error --include --write-out "\n%{size_header}\n%{content_type}\n%{response_code}\n%{size_download}" http://www.http2demo.io/',result,eee
```
Get in eee:

> curl: (3) URL using bad/illegal format or missing URL

Get in result:

> 0  000 00

Running in a command line (cmd):
curl -LI --silent --show-error --include --write-out "\n%{size_header}\n%{content_type}\n%{response_code}\n%{size_download}" http://www.http2demo.io/

Get:

> HTTP...
> ....
> 
> 392
> text/html
> 200
> 0

curl 7.78.0 (x86_64-pc-win32) is in PATH. Another version of curl was also tried. The result is the same.
OS is Windows 11 Home 21H2.

GillesDuvert:
Hi thanks for the report.
This procedure has never been tested under Windows before and the fact that spawning curl does work is already a feat! 😄 
Probably the spawn C++ code does unfortunate conversions of the 'sent' string, like replacing /  by \ etc on this platform.

Albom:
Thank you! Yes, really. In basic_pro.cpp (Line 1835), there is a such conversion:

```cpp
std::replace(cmd.begin(), cmd.end(), '/', '\\');
```

Therefore, the string  http://www.http2demo.io/ is converted to  http:\\\\www.http2demo.io\\. 

Trying curl in a command line:

> curl -LI --silent --show-error --include --write-out "\n%{size_header}\n%{content_type}\n%{response_code}\n%{size_download}" http:\\\\www.http2demo.io\\

And get:

> curl: (3) URL using bad/illegal format or missing URL
> 
> 0
> 
> 000
> 0

Is this conversion required? Users should know what kind of slash symbols ('/' or '\') to use...

GillesDuvert:
Yes the call passes OK on unix/linux and probably OSX as well.
GDL is supposed to behave as IDL (or better :smile) so in the particular case of  idlneturl:: use GDL must provide a call to curl (external or, in the future, internally using libcurl) that does not make such conversion on windows.
This opens a greater question: does the windows version of gdl's 'spawn' need to make this translation at all? I doubt IDL does it.

Albom:
[...]

PR #1468, opened 2023-01-10
SPAWN: a windows change

GillesDuvert:
normally even on windows, spawn should not change the content of the spawned command.


GillesDuvert:
see #1465

Issue #1471, opened 2023-01-11
problem with the size of the first window with wxwidgets ! (+valgrind / callgrind)


alaingdl:
This is a continuation of Issue #1461 since this issue (no first drawing of plot) is now solved in current Git version (to be the 1.0.2). To my knowledge this is the last real basic bug for a 1.0.3 delivery. As long as SHADE_SURF is not solve it is not smart to have a 1.1, isn't it ?

Call 1 : plot, indgen(10) & Window, 2
Call 2 : Window, 0 & plot, indgen(10) & Window, 2
Both windows should have the same size.

* Using the X11 version I have no problem, locally or on remote (ssh -X)
```
cmake .. -DWXWIDGETS=off -DX11=on -DPYTHON=off -DQHULL=off -DCMAKE_INSTALL_PREFIX=$PWD/install
make install
./install/bin/gdl
```

Call 1 and Call 2 give the same output, and the size of both windows (0 & 2) are the same

* Using WXwidgets I do have problems
```
cmake .. -DX11=off -DPYTHON=off -DQHULL=off -DCMAKE_INSTALL_PREFIX=$PWD/install
make install
./install/bin/gdl
```

Call 1 and call 2 give the same output **but** the size the the 2 windows are different

When running that remotely, it is absolutely clear that the first window is plotted in two steps, the first one has the good size (but with a white/grey background) then is resized to a bad size and a black background.

This problem is maybe related to a side effect within the window manager. I do use `xfwm4` and when I double the size of the top bar, the size of the first window is altered by the same factor.

**About Callgrind**

I also try to use `valgrind --tool=callgrind ./install/bin/gdl` (yes I stored the outputs, so called _callgrind.out.1234_, 1234 is the pid of the run as usual with `valgrind`. To be explore with `kcachegrind`. Since these files are big (5 Mo each) I cannot put them here ) and it seems, following the graph of the internal calls, that Call 1 and Call 2 don't follow the same path to reach the same final status (in both cases, X11 & wx). I do see that as a risk of bugs and future problems for the maintenance of the code.

alaingdl:
I forgot to mention that to use valgrind + callgrind we need to compile : 
`cmake .. -DCMAKE_CXX_FLAGS_RELEASE="-g -pg"  [other flags to cmake]`

GillesDuvert:
I succeed to reproduce on a Debian10 . Obviously some uninitialized size somewhere. Indded, must be solved before 1.0.2

[...]

PR #1473, opened 2023-01-13
Windows sizes (and more) to solve #1471 and many others

GillesDuvert:


alaingdl:
Bravo ! Solve locally the problem :smile: 

(I cannot test remotely this WE, annual turn off of electricity for _maintenance_)

PR #1477, opened 2023-01-16
use the general library whereami for getting the path of the excutable on all platforms.

GillesDuvert:
Blind test.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -196,10 +196,12 @@ ${WX_RC}
 endif(HAVE_LIBWXWIDGETS AND WIN32)
 #local plplot drivers
 add_subdirectory(plplotdriver)
+#whereami
+add_subdirectory(whereami)
 
 add_subdirectory(antlr)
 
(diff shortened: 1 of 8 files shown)


alaingdl:
Thanks ! Tested (yes 32 cores ;) just now with **success** on 3 configurations : debian 10 + script, debian 10 + cmake + cmake_install_path  and OSX + script

Sorry but my VM with a BSD system is down now (TBC tomorrow ?) Better to have feedback from @thierry-FreeBSD and others ! Thanks !

slayoo:
@GillesDuvert, let's perhaps use a git submodule (https://git-scm.com/book/en/v2/Git-Tools-Submodules) for bringing in the `whereami` files into `gdl`. This will make updates easier as it will be clear which revision of which repo we point to. From the point of view of compilation, nothing changes. It's just that when cloning gdl repo, one will need to do `git clone --recurse-submodules` instead of `git clone`. Thanks

GillesDuvert:
Yes I knew about submodules and tried it but did wrong, hence the first failure in compilation in this PR, because the whereami folder did not propagate into github correclty.
Unfortunately I only use NetBeans for all the git stuff. Apparently it does not understand submodules, hence my problem.
I'm very marginally interested in a solution that would force me to use a git line command externally to NetBeans each time I make a git push/pull etc...

slayoo:
@GillesDuvert, this should help to include the submodules while checking out for CI: https://github.com/gnudatalanguage/gdl/pull/1482

PR #1482, opened 2023-01-17
fine-tune checkout action (submodules, fetch-depth)

slayoo:


GillesDuvert:
...But does not actually make the whereami folder a submodule i.e. a pointer to https://github.com/gpakosz/whereami

slayoo:
> ...But does not actually make the whereami folder a submodule i.e. a pointer to https://github.com/gpakosz/whereami

Here's a PR: https://github.com/gnudatalanguage/gdl/pull/1484

PR #1484, opened 2023-01-18
move whereami into a submodule

slayoo:


alaingdl:
seems to have problems here : 

```
CMake Error at src/CMakeLists.txt:200 (add_library):
  Cannot find source file:

    /home/alaingdl/GDL/TestPR1484/gdl/src/whereami/src/whereami.c

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx


CMake Error at src/CMakeLists.txt:200 (add_library):
  No SOURCES given to target: whereami

```

GillesDuvert:
@alaingdl this is looks like an old error during an early phase of the patch. gdl/src/whereami/src does not exist. Perhaps a line inadvertly forgotten in CMakeFiles. I'll check.
@slayoo This is typically what is going to happen if we use submodules.

GillesDuvert:
Ok, works with last @slayoo change.
So we have now a submodule. Good first example.

GillesDuvert (commented):
OK!!!

GillesDuvert:
NOK!!
The [whereami @ ba364cd](https://github.com/gpakosz/whereami/tree/ba364cd54fd431c76c045393b6522b4bff547f50)i folder is not propagated to a clone made or managed by NetBeans!!!

GillesDuvert:
OK, this has to be done manually. By removing the cloned gdl, replacing it with the command
`git clone --recursive https://github.com/gnudatalangage/gdl.git `
Then more or less have Netbeans refresh its paths and files by resetting the "Source Folder" in the project (sufficed in my case).

alaingdl:
(this command is not working here, I have to make : 
`git clone --recursive https://github.com/gnudatalanguage/gdl/` )

OK, now works for me with this `--recursive`, the previous was NOK (u22.04)
TBC on OSX ASAP.

slayoo:
@alaingdl @GillesDuvert when a new submodule is added on GitHub and you update your local clone, git does not by default obtain submodule contents - just metadata. Instead of re-cloning with the `--recursive` option, one can also use `git submodule update` or do `git pull --recurse-submodules` as described here (https://git-scm.com/book/en/v2/Git-Tools-Submodules#_pulling_upstream_changes_from_the_project_remote):
image

alaingdl:
heeuuu, don't work for me (on Debian 10.13; git version 2.20.1)

slayoo:
> heeuuu, don't work for me (on Debian 10.13; git version 2.20.1)

@alaingdl, please elaborate

alaingdl:
Sorry. I tried both commands you mentioned (`git submodule update` or do `git pull --recurse-submodules`), on a clean `git pull` version, but when I try to compile, I got the same error message as before (may be I made a mistake)

slayoo:
@alaingdl, please try `git submodule update --init`

alaingdl:
Thanks, yes, `git submodule update --init` is working.  And `git clone --recurse-submodules  https://github.com/gnudatalanguage/gdl` too.

We have to document that somewhere because the basic commands are no more enough to got it

PR #1491, opened 2023-01-23
add missing-submodule warning to CMakeLists.txt

slayoo:


alaingdl:
1/ pb with gdlde ?!
2/ tested here and it was OK (issue the expected message)
3/ could we add that in the build script (a check or a command ?)

GillesDuvert:
>  3/ could we add that in the build script (a check or a command ?)

build_gdl.sh calls cmake so I guess this is OK ?

Issue #1487, opened 2023-01-22
not quite there for 1.0.2


GillesDuvert:
Seems the first plot, that was OK before #1483 on Mageia 8 + KDE, is not OK anymore.
Apparently GNOME and KDE are not mutually compatible with wxWidgets!

GillesDuvert:
TO BE MORE PRECISE:
- GDL was working on xorg ( KDE and GNOME)
- We found a problem of window sizes (wxWidgets windows, not X11 windows) on systems with Wayland.
- Curing (in our own dumb way since it is a wxWidgets bug) for Wayland makes the problem appear WITHOUT Wayland.

GillesDuvert:
According to https://github.com/wxWidgets/wxWidgets/issues/23041 wxWidgets developed a "quite horrible' mechanism for this same reason with X11 (xorg) that's why it worked with X11 from the start.

PR #1492, opened 2023-01-23
solves #1487 : window sizes are OK in all cases I tested.

GillesDuvert:
This on should give the green light to 1.0.2


alaingdl:
I tested with success locally (u2204) and remotely (Debian 10)

Remotely, the way the Wxwidget window is generated is significantly slower than the X11 one, and seems to be processed in two steps (one grey then one black)

PR #1497, opened 2023-01-26
set CMake C standard flag (for whereami)

slayoo:


alaingdl:
tested with success on CentOS 7 + cmake3
(but the script "build_gdl.sh" is not working because it need to call `cmake3`, not `cmake`)

slayoo:
@alaingdl shall we then merge this one?

alaingdl:
heu, sorry, in fact I tried to merge it ... 

it is ok for me : it compiled for the old system, and the recent systems too

Issue #1474, opened 2023-01-14
current file pointer jumps to the end when file grows after it has been opened


brandy125:
The current file pointer must stay at position 100000 even if the file grows:

```
GDL> openw,1,'eraseme'
GDL> line1=bytarr(1000)
GDL> for i=0,99 do writeu,1,line1
GDL> flush,1
GDL> 
GDL> openr,2,'eraseme'
GDL> line2=bytarr(1000)
GDL> for i=0,99 do readu,2,line2
GDL> print,fstat(1)
{ 1 eraseme   1   0   0   0   0   0   1   1  1673731066 1673731024 1673731024 0 100000 100000  0}
GDL> print,fstat(2)
{ 2 eraseme   1   0   0   0   0   0   1   0  1673731066 1673731024 1673731024 0 100000 100000  0}
```
up to here it is correct. But if we increase the file size then the current pointer must stay at position 100000 as it is in IDL. It must not jump to 200000 which is the size of the file.
```
GDL> for i=0,99 do writeu,1,line
GDL> flush,1
GDL> print,fstat(2)
{ 2 eraseme   1   0   0   0   0   0   1   0 1673731066 1673731087 1673731087 0 200000 200000 0}
```

brandy125:
Can anybody help here? I would need a solution for this since I am working with realtime data right now and this is a "no go".
Thanks!

alaingdl:
1/ reproduced
2/ seems to be related to method Tell(p in `io.cpp` which does not distinguish between reading and writing case

alaingdl:
I changed your code in such a way : 

```
pro test_issues1474, maxi=maxi

if ~keyword_set(maxi) then maxi=99

close,1, 2

openw,1,'eraseme'
line1=bytarr(1000)
for i=0,99 do writeu,1,line1
flush,1

openr,2,'eraseme'
line2=bytarr(1000)
for i=0,maxi do readu,2,line2
f1=fstat(1) & print, 'f1 :', f1.unit, f1.cur_ptr, f1.size
f2=fstat(2) & print, 'f2 :', f2.unit, f2.cur_ptr, f2.size
;print,fstat(1)
;print,fstat(2)

for i=0,99 do writeu,1,line1
flush,1
f1=fstat(1) & print, 'f1 :', f1.unit, f1.cur_ptr, f1.size
f2=fstat(2) & print, 'f2 :', f2.unit, f2.cur_ptr, f2.size
;print,fstat(2)

stop
end

[...]

PR #1502, opened 2023-02-07
tentatively solve #1474

GillesDuvert:


slayoo:
@GillesDuvert @acoulais @brandy125 it would be great to include a test with this PR (one that fails without this change, and passes with the change applied) Thanks!

alaingdl:
The good news : successfully tested (with my script) on Debian 10 & U22.04

The bad news : unable to compile GDL with current Brew with the build script or with CMake. I do have a plplot dylib but not OK :(

GillesDuvert:
This patch does not totally solve #1474.
There are still a number of differences between IDL and GDL when it comes to read/write files that can be modified **externally** (NB: externally is the point. GDL behaves OK when it modifies a file internally.)
As always,  it is best (but time-consuming :smile:  ) to start from the example reported by the user, but explore in a test file all the possible outcomes. Here, the example lacks the fact that **writing** to an externally  modified file does not produce identical results on IDL and GDL.

slayoo:
Merci, @GillesDuvert!

GillesDuvert:
> The good news : successfully tested (with my script) on Debian 10 & U22.04
> 
> The bad news : unable to compile GDL with current Brew with the build script or with CMake. I do have a plplot dylib but not OK :(

Homebrew's last version seems buggy: cannot permit to use a modified formula to rebuild plplot. So: `build_gdl.sh` uses scriptlet "`brew_enable_wxwidgets`" for nothing (formula used is not the modified one but a cached one), and thus no dynamic drivers, and thus no GDL. 
On MacPort side, idem, and would need yet another trick to  recompile plplot with the correct options within MacPorts.
Wonder why they do not provide these options from the start.
We should have asked homebrew and MacPorts for a version of plplot with dynamic drivers and wxwidgets a long time ago.

Besides, on arm64, MacPorts or Homebrew gcc is not working, and we must use clang. But clang somehow makes the creation/destruction of GDL objects very slow (probably an error in the legacy GDL code, who knows?) so GDL is not satisfactorily fast.

This is IMHO a serious concern as GDL contributors, whatever their skills and implication, are not supposed to fight Homebrew, MacPorts and clang, but to add new useful commands to GDL.

GillesDuvert:
On Windows I see that we have found an INTERNAL BUG of MINGW that must be reported!

[100%] Linking CXX executable gdl.exe
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: internal error: aborting at ../../binutils-2.40/ld/ldlang.c:527 in compare_section
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: **please report this bug**

GillesDuvert:
1. bug above reported
2. macOS build solved by #1510

PR #1510, opened 2023-02-10
aimed at allowing gdl build on OSX even on arm64 and OSX 13.2

GillesDuvert:
this simply compiles git::plplot locally instead of using (broken) Homebrew.


GillesDuvert:
the Windows build crahes on a recent debian bug  https://sourceware.org/bugzilla/show_bug.cgi?id=30079

Apparently our builds use debian unstable... a bit risky, no? they say: "unstable".

PR #1515, opened 2023-02-15
Correct surface and shade_surf

GillesDuvert:
(going around plplot limitation)
Using the driver 3D Matrix (again) one can get the 3D surfaces and shade_surf approximatively OK in all cases.


GillesDuvert:
The Windows error found during continuous integration is not directly related to test_issue1474 but to a more profound problem with Tell() under windows that was not found by a test before.

Issue #425, opened 2018-07-17
implement BREAKPOINT


slayoo:
https://www.harrisgeospatial.com/docs/BREAKPOINT.html

slayoo:
Posting here the recent message from SF.net discussion by cosmicbruce:

 

> 
> Dear Alain and developer community:
> 
> I don't find any of (colon operator, *gen keyword, implicit print, issues) important for my use, nor do I know others who extensively use it. But it is in simple examples in on-line materials now, so might be good to address this.
> 
> breakpoint, on the other hand, is really valuable.
> 
> My recommendations (for what they are worth):
> * Fix colon operator, findgen, indgen,.. keywords, implicit print issues -----------> Zero Allocation of Resources (live with it)
> 
>     Publish above as "gdl / idl incompatibility issues" -----> Worth doing: Put on the net where it can be found with a detailed search in < 1 minutes of work. Later, when someone has time, make a nice, as complete as possible list of incompatibilities- this would be a valuable reference.
> 
>     Breakpoint: I feel this is in a different category entirely. For me, IDL / gdl power is strictly and only live-time debugging. For example, the command ".skip", as far as I can tell, is completely unique among all languages I've heard of. So, enhancing such capabilities is important to keeping gdl relevant.
> 
> I would say .step is a basic building block of live-time debugging. Breakpoint allows one to precisely apply .step in a very long program. So, to me, the live-time debugging capabilities of gdl will take a significant step up if this is implemented.
> --------->I would humbly request developers please consider working on this.
> 
> Example: you are not sure if some error in the answer comes from the statement in line 50 of your code, or line 350 or 550. You can easily look in your editor and note these linenumbers, and apply breakpoint. But you have to count statements, not line numbers, to use .step. If there are some 150 statements or so between locations where you want to stop and look at values, .step is not practical because of the inconvenience of converting the number of statements (required by .step) to line numbers (that you see in your editor). Without breakpoint, you need to add stop statements, which you have to remove again later, re-compile, and re-run. So, breakpoint is much faster, more interactive, and more convenient.,
> 
> IDLwave does work with gdl, but of course, you can't use it's nice button for the breakpoint statement. So, increased compatibility with IDLwave will also result.
> 
> Thanks,
> 
> -Bruce
>

GillesDuvert:
[...]

PR #1528, opened 2023-03-01
congrid was complaining if last dimension was 1

GillesDuvert:
see https://github.com/gnudatalanguage/gdl/issues/425#issuecomment-1450411335


slayoo:
Kudos for including a new test case!

alaingdl:
yes !

but I would like some/most tests in : spawn_is_dangerous/  and interactive_tests/ go back to testsuite/ because they are useful, and few are now broken (regression) and we forgot to test them.

For SPAWN, for me it is __not__ a problem since it is usage in given contexts we wrote.
For "interactives" and graphics" we may change some outputs to Z or PNG but it is useful to run them

GillesDuvert:
@acoulais of course the interactive and span_is_dangerous tests are badly needed. Just insure they work on builds without displays, or X11 and that spawn commands are those understood by the OS, windows in particular. See #1465 for a recent example.

PR #1549, opened 2023-03-15
rewrote write tiff

GillesDuvert:
Does not use Tiff library.
Complete except for BIGTIFF and COMPRESS support -- not a big deal for BIGTIFF, just copy and change a few Longs to Long64...
Is an example of things **people can contribute with only their IDL skills**, C++ not needed. must be true in particular for all format readers/ writers.


brandy125:
Hi @GillesDuvert geotiffs are getting bigger and bigger. Did you find already someone who can implement bigtiff?

PR #1553, opened 2023-03-16
Better gaussfit, closes #1548

### LINKED ISSUE #1548: GAUSSFIT don't work well whit nterms > 4
GAUSSFIT in GDL don't  resist in the presence of polynomial (nterm > 4)

I have no idea today how to fix it, sorry.

GillesDuvert:


alaingdl:
It is working way better *but* on my tests the result in case of nterns 5 or 6 are less good that with the original. Please wait a few I can check where the discrepancy comes from

slayoo:
@alaingdl @GillesDuvert let's add a test here!

GillesDuvert:
replaced by #1556

GillesDuvert:
this is closed and replaced by #1556

Issue #1576, opened 2023-04-22
GDL 'Direct' function crash on trinary expressions


GillesDuvert:
THIS IS AN INTERNAL ISSUE --- NOT FOR THE FAINT OF HEART
```
GDL> x=1 & y = 2 & a=0 & b=1 & z=cos(x le y ? a:b)
Segmentation fault (core dumped)
```
Direct functions are those, like 'cos' that work on a single argument (y=cos(x)) and thus an expression like y=cos(x) can be estimated rapidly if x is a known variable (opposed to an expression that has to be computed). Large gain of speed, for many siple math functions, and corresponds to the vast majority of use-cases. Single arguments like that permit to avoid creating a new environment, so the speed gain, and direct access to the addresses of variables inside the core function ('cos()' in the example).

GillesDuvert:
precision: the crash arises when the returned value of `x le y ? a:b` is a variable, not an expression:
GDL> `a=0 & b=0 & c=0 & z=cos( a ? b:2) ;` works, and we note that a ? b:2 gives 2, 2 is a 'value' . cos() is called with `isReference=false`  since a 'value' is not a GDL variable.
 GDL> a=0 & b=0 & c=0 & z=cos( a ? 1:b) ; crash!  cos() is called with `isReference=true`  since `b` is a GDL variable.

GillesDuvert:
Also, we have another silliness:
```
GDL> a=0 & b=0 & c=0 & z=cos( a ? b:!radeg)
% Attempt to write to a readonly variable: !RADEG
% Execution halted at: $MAIN$   
```

GillesDuvert:
closed by patch #1579

PR #1579, opened 2023-04-25
solves #1576

GillesDuvert:


slayoo:
@GillesDuvert Let's add a test case for this one!

GillesDuvert:
Sorry I would prefer to work on the non-trivial parts.

PR #1583, opened 2023-04-27
implement value_locate() in C++ instead of slow pro version before.

GillesDuvert:
Can be optimized further for large searches.


GillesDuvert:
speed gain is incredible.

Issue #830, opened 2020-09-02
Record terminator slash (/) in format not supported


lwymarie:
IDL input and output: 

IDL>       print,'NOTE: Calling sequence to IFSF_LINELIST has changed. If vacuum',$
IDL>             '      wavelengths are required, or other wavelength units, or ',$
IDL>             '      to suppress this message, use ARGSLINELIST structure in ',$
IDL>             '      initialization procedure. Default is now air waves.',$
IDL>             format='(A0/A0/A0/A0)'
NOTE: Calling sequence to IFSF_LINELIST has changed. If vacuum
      wavelengths are required, or other wavelength units, or 
      to suppress this message, use ARGSLINELIST structure in 
      initialization procedure. Default is now air waves.

GDL output is not as expected: 

GDL> print,'NOTE: Calling sequence to IFSF_LINELIST has changed. If vacuum',$
GDL> '      wavelengths are required, or other wavelength units, or ',$
GDL> '      to suppress this message, use ARGSLINELIST structure in ',$
GDL> '      initialization procedure. Default is now air waves.',$
GDL> format='(A0/A0/A0/A0)'
% PRINT: Infinite format loop detected.
% Execution halted at: $MAIN$

GillesDuvert:
Thanks for the report!
in other terms,
`print,"a","b",format='(a/a)'` is badly handled by the GDL interpreter.
The documentation says that '/' and ',' are both format separators, not only ','. 
To be solved...
 Note that `format='(a/,a)' ` works and is a temporary solution.

Issue #369, opened 2018-06-16
migrate to antlr 3 or antlr 4


slayoo:

slayoo:
For the record, there are some references around:
- "Migrating from ANTLR 2 to ANTLR 3":
  https://theantlrguy.atlassian.net/wiki/spaces/ANTLR3/pages/2687070/Migrating+from+ANTLR+2+to+ANTLR+3
- "Antlr 3 to Antlr 4 Migration Guide":
  https://github.com/antlr/antlr4/issues/464
- ANTLR v2 to v3 converter:
  http://www.antlr3.org/download/v2v3-0.2.tar.gz

PR #1588, opened 2023-05-03
WIP Remove many antlr issues

GillesDuvert:
This is a WIP.
All the open issues directly related to ANTLR should be closed by this PR.
~~For the sake of maintainance, the version of 27/09/2013 (just before some complicated optimization introduced by Mark Schellens) has been used as template (upon which all the subsequent small changes have been reported and the open issues are solved). This to avoid a dependency between the ANTLR code and the antlr-generated code (code based on internal values of '_tokenSet_XX' that may be changing with the number and order of tokens generated by antlr). Indeed, nobody can IMHO debug such complicated dependencies. The loss in performance induced by this pseudo-regression has to be estimated (it seems negligible) and weighted against the advantages of having more chances to maintain correctly the code, even going to ANTLR3 or 4 more easily.~~
:warning:  I was obliged to use the last version of gdlc.g because contary to what is stated above, the last optimized code by Marc is twice as fast as the one referred to above.
:warning:  The last remaining problem with the ANTLR code is that it will become very very slow during the Parser phase when there is even a small number of (matched) parenthesis. For example a=(3) is instantaneous, a=((((3)))) is very slow.
:warning:  At the moment the **format.g issues nondeterminism warnings when compiling** by antlr, this is **OK** as the order of branching makes the formats (issue #830)  behave OK, so no further maintainance is needed.


GillesDuvert:
Hmmm... The tests show incredible low speed on a few tests, so there is something fishy with the backport of Marc's optimizations (was to be expected)
---> One should separate the compilation time from the execution time in the tests by redoing every test twice: sometimes a long test is just due to a long Parser time, iself due to the long-standing problem of () vs. [] in our ANTLR.

GillesDuvert:
speed of GH continuous integration tests are very bad in all cases (typically 5 to 10 times slower than on my old HP notebook) because the code is unoptimized, the number of cores is 1 and the machine virtual. So the speed tests are not particularly useful. But in the present case, indeed there IS quite a huge penalty in compilation and execution speed if the ANTLR code is not well-written (last patch https://github.com/gnudatalanguage/gdl/pull/1588/commits/89989879a32dd8d59d0c0084f8945c471bfa4301).
This suggest further speedup can be attained by modifying the ANTLR code, possibly after converting it to ANTLR4 (see #369) . It would be recommended in this case to use an ANTLR code (especially for gdlc.g) without Marc's optimizations, at least without the dedicated code using tokenSet_xx .

PR #1601, opened 2023-05-26
fixed codecov so it actually works

Saynkas:


slayoo:
@Saynkas Thanks! 
The CI status is "red" due to codecov and (unrelated) windows build errors.
How to interpret the codecov failures?

Saynkas:
> @Saynkas Thanks! The CI status is "red" due to codecov and (unrelated) windows build errors. How to interpret the codecov failures?

it seems that codecov doesn't like the overall coverage % dropping below what it was before (not too unexpected, the goal is to go towards 100% coverage, not 0%, and codecov had been out of service for 5 months so the code changed since then)

apparently you can specify a threshold to allow the codecov checks to not fail (eg if the threshold is 10% then the overall % can drop by 10% and not count as 'failed'), and by default it compares to the latest commit's %

should I edit the .yml to account for this, and if so, what threshold value should I use ?

taken from https://docs.codecov.com/docs/commit-status

slayoo:
Thanks for clarifying - IIUC, after merging we do want to block PRs lewering the coverage ratio - merging!

Issue #1608, opened 2023-07-24
compilation error when program contains for-loop with begin and endfor in one line


brandy125:
I thought this was solved already but I am not sure any more.

This simple program does not compile any more:

```
pro test
   for i=0,5 do begin print,i & endfor
end
```

It only compiles when there is a "&" after the "begin":

```
pro test
   for i=0,5 do begin & print,i & endfor
end
```

or when the commands are in different lines:

```
pro test
   for i=0,5 do begin 
      print,i
   endfor
end
```

GillesDuvert:
@brandy125 I'm afraid its a consequence of #1588 as reported in #1599 . Will be solved, even if it means reverting to previous behaviour.

brandy125:
Is see…thanks!

Issue #1599, opened 2023-05-25
ANTLR regression


GillesDuvert:
pro testme
	for i=0,1 do begin  $
		print,i
	endfor
end
does not pass correctly since #1588

Issue #52, opened 2018-04-07
shorthand notation for loop in command line mode 


slayoo:
(https://sourceforge.net/p/gnudatalanguage/bugs/174/)

```
IDL> for i=0,2 do begin print,i
0
1
2
```
```
GDL> for i=0,2 do begin print,i
% Parser syntax error: unexpected end of file
```

Reported on SF.net back in 2009, and later mentioned in 2012 here:
https://sourceforge.net/tracker/index.php?func=detail&aid=3475493&group_id=97659&atid=618683

jtappin:
Maybe with the current flurry of activity on inconsistencies between GDL and IDL now is a good time to wake this one up again.

GillesDuvert:
Hi @jtappin the tag "antlr" means it should be done by someone well at ease with the ANTLR parser... Nobody in sight...

jtappin:
O.K.

PR #1610, opened 2023-08-05
reverted for_block to historical version

GillesDuvert:
solves #1599 and #1608. 
#52 should be reactivated.


GillesDuvert:
hm. a file is not up-to-date

GillesDuvert:
merging.
Will have to find why tests crash on Windows, of course they do not on my (virtual) machine.

Issue #1618, opened 2023-08-24
integer powers problems wrt IDL's


GillesDuvert:
This is a summary of present discrepancies btw. GDL and IDL wrt. the POW (`^`) operator.

Generally speaking, integer powers of something can be computed faster using the so-called exponentiation by squaring instead of using std::pow(). think to a^2=a*a (1 machine instruction) instead of exp(log(a)*2) . Many times GDL manipulates simple integer powers of arrays, so the use of an optimized version of pow for integers is mandatory.

Hence #1350 and #1351.

nevertheless, at this time:
- there appears to be some support for integer powers written by Marc in the early times of GDL, using exponentiation by squaring.
- but it is not at all optimally used since:
```
GDL> a=randomu(33,1E8) & b=3.0 & tic & c=temporary(a)^b & toc
% Time elapsed :       0.22601295 seconds.
GDL> a=randomu(33,1E8) & b=3 & tic & c=temporary(a)^b & toc
% Time elapsed :       0.46429801 seconds.
```
the latter using PowInt instead of pow()

GillesDuvert:
... but the inverse is true too:
```
GDL> a=randomu(33,1E8,/long) & b=3.0 & tic & c2=temporary(a)^b & toc
% Time elapsed :       0.30532098 seconds.
GDL> a=randomu(33,1E8,/long) & b=3 & tic & c1=temporary(a)^b & toc
% Time elapsed :      0.054013968 seconds.
```
In each of the above cases, a (very) different 'pow' function is called, based on the type and size of the operands. This may explain that. I note that in the present case, GDL is additionnally faster than IDL.

GillesDuvert:
Finally one should realize that current GDL uses its PowInt() in some very restricted cases, and the C++ `std::pow()` function almost everywhere, with internal conversion to and from `floats` and `double` since std::pow() exists only for those types.
So there is ample room for different results between IDL and GDL depending on the number of atomic operations that will add rounding errors. I checked:
```
IDL> a=fix(!dpi,type=4) & pow=7 & z=1 &for i=0,pow-1 do z*=a & print,z,a^pow,format='(F40.20)'
               3020.29394531250000000000
               3020.29394531250000000000
GDL> a=fix(!dpi,type=4) & pow=7 & z=1 &for i=0,pow-1 do z*=a & print,z,a^pow,format='(F40.20)'
               3020.29394531250000000000
               3020.29370117187500000000
```
[...]

Issue #1531, opened 2023-03-04
Inheritance


rdimeo:
I have a large IDL application (that currently works in IDL) that I'm trying to get running in GDL. In this application I create a class that inherits the IDL_CONTAINER object. However when I try to create this new object, it crashes when trying to invoke the INIT method of the IDL_CONTAINER object. It cannot find the INIT method of the IDL_CONTAINER object. The test code below is a short description and shows the error whenever you try to create the object TEST_CONT via o = obj_new('test_cont').  Note that this test code works fine in IDL. Any ideas would be greatly appreciated. Thanks in advance!

; ************************** ;
pro test_cont::cleanup
self->IDL_CONTAINER::cleanup
end
; ************************** ;
function test_cont::init
; Initialize superclass
if not self->IDL_CONTAINER::Init() then return,0B
return,1B
end
; ************************** ;
pro test_cont__define
define = {TEST_CONT, INHERITS IDL_CONTAINER}
end

alaingdl:
Thanks you for the very clear bug report.

Unfortunately I am totally ignorant in such way to use IDL :(

I was only able to reproduce your problem : 

```
GDL> .r test_cont.pro 
% Compiled module: TEST_CONT::CLEANUP.
% Compiled module: TEST_CONT::INIT.
% Compiled module: TEST_CONT__DEFINE.
GDL> o = obj_new('test_cont')
% TEST_CONT::INIT: Attempt to call undefined method: GDL_CONTAINER::INIT
% Execution halted at: TEST_CONT::INIT      9 /home/alaingdl/GDL/test_cont.pro
%                      $MAIN$          
```

GillesDuvert:
Hi, as the whole sswidl nasa software works with GDL, (sort of, but that's another topic), GDL is seemingly able to deal with objets in general, as they have tons of them. The only difference is IMHO the use of IDL_CONTAINER inside the functions (e.g.: `self->IDL_CONTAINER::cleanup`, where `self->cleanup` would have worked) .

 GDL creators avoided to use the term IDL in their code, but that's becoming difficult. I suspect there is a mismatch in the GDL code somewhere. This does not look too complicated, but may take time.

Would removing all references to  IDL_CONTAINER:: in a copy of the IDL application may save the day? After all, inheritance concept makes `self->IDL_CONTAINER::cleanup` and `self->cleanup` equivalent in all reasonably well programmed cases?
The example below works
```
function myobj::Init
  return, 1
end
function myobj::get
[...]

Issue #1286, opened 2022-03-03
discussion of preformance issue as seens by TEST_OP_POWER on large arrays


alaingdl:
Please remember that, _today_ TIME_TEST4 is faster in current GDL than in IDL 8.8

But for most of the operations, on a 48-core machine, with : 
` TEST_OP_POWER, 1E8`, performances of IDL are way better (3 to 10 times faster) except for power 3.5

(for 1e7 and 1e6, discrepancies are more limited)

I have no experience with C++11 new capability "valarray" + pow 
 https://www.cplusplus.com/reference/valarray/pow/
by I suspect we have to test that !

A.

```
IDL> TEST_OP_POWER, 1E8
% Compiled module: TEST_OP_POWER.
some tests on POWER mathematical OPerations
First test are on LONG INTEGER
Product (x1 * x2): .......     0.018408060
Equiv. Square (x1 * x1): .     0.015163898
Real Square Op (x^2): ....     0.019263983
Cube Op (x^3): ...........     0.019178867
3.5 Op:(int-->flt)........      0.21213603
^4 Op: ...................     0.022486925
Second test are on FLOAT
Product (x1 * x2):. ......     0.016569853
Equiv. Square (x1 * x1): .     0.014093876
Real Square Op (x^2): ....     0.024669886
Cube Op (x^3): ...........     0.024806976
3.5 Op:(int-->flt)........      0.19396806
^4 Op: ...................     0.026787996

GDL> TEST_OP_POWER, 1E8
% Compiled module: TEST_OP_POWER.
some tests on POWER mathematical OPerations
First test are on LONG INTEGER
Product (x1 * x2): .......      0.15494895
Equiv. Square (x1 * x1): .      0.12177396
Real Square Op (x^2): ....     0.067502975
Cube Op (x^3): ...........      0.22579312
3.5 Op:(int-->flt)........     0.091262102
^4 Op: ...................      0.19624090
Second test are on FLOAT
Product (x1 * x2):. ......      0.15187788
Equiv. Square (x1 * x1): .      0.11982489
Real Square Op (x^2): ....     0.051900148
Cube Op (x^3): ...........      0.15119600
3.5 Op:(int-->flt)........     0.038867950
^4 Op: ...................      0.15120792
```

PR #1620, opened 2023-08-26
Integer power optimizations

GillesDuvert:


GillesDuvert:
closes #1618 #1286 #1530 #1531

PR #1631, opened 2023-09-07
changes to get the simplest way to get valid useable DMG on OSX.

GillesDuvert:
WIP only


GillesDuvert:
not good. libplplot is missing.

Issue #1567, opened 2023-04-11
Postscript graphic dimensions


Barney-9000:
A recent recompile of GDL (28/03/23 -- macOS 11.7.5 intel) now results in postscript graphics that deviate markedly from the (centimetre) dimensions specified in the DEVICE command. For example, a postscript graphic which should be 18x24cm is output by GDL at 15.7 × 24.73 cm. An example from IDL & GDL attached. GDL always was a little bit out, but now the deviation its quite a bit larger. This may be a recurrence of a boxsize related oddity fixed in 2021 #1002. Interestingly, the two plots in the GDL graphic have the correct dimensions and exact relative position, see graphic with the two graphics superposed.

All the best -- and many thanks for keeping GDL alive and kicking!

[IDL.pdf](https://github.com/gnudatalanguage/gdl/files/11199947/IDL.pdf)
[GDL.pdf](https://github.com/gnudatalanguage/gdl/files/11199950/GDL.pdf)
[GDL-IDL-superposed.pdf](https://github.com/gnudatalanguage/gdl/files/11199990/GDL-IDL-superposed.pdf)

ogressel:
I post this here, since it's presumably related. Using `xsize` and `ysize` with `device` results in a substantially distorted aspect ratio. A minimal example is:
```
pro iso, eps=eps

  if keyword_set(eps) then begin
     set_plot, 'ps'
     device, xsize=24., ysize=8., file="./iso.eps", /encap
  endif

  plot, [0],[0], xr=[-1,1], yr=[-1,1], xsty=1, ysty=1, /isotropic, /nodata

  oplot, [-1,1],[1,-1]
  oplot, [-1,1],[-1,1]

  return
end
```

ogressel:
@GillesDuvert : I've traced the aspect-ratio-issue down to commit d6a8e4e55fdfe5111081e715050743378b3605cf from two years ago. Since I'm unfamiliar with PostScript dimensions/units, I was wondering whether you have a clue what might be the cause of the issue...

Barney-9000:
[...]

PR #1638, opened 2023-09-13
PostScript shape,size and position improvements.

GillesDuvert:
this PR concerns exclusively the PostScript driver and PostScript Device

It  should
 - insure exact position and dimension of paper (`plots,[0,0,1,1,0],[0,1,1,0,0],/norm`) on a postscript file whatever the size ratio and landscape/portrait
 - a better THICKness of lines for PostScript output
 - no more 'stretching' of Hershey fonts and Symbols as described in commit https://github.com/gnudatalanguage/gdl/commit/d6a8e4e55fdfe5111081e715050743378b3605cf
 - a better default size for landscape PostScript outputs.  
 - possibly solving #1567 also


GillesDuvert:
The failure of test_tic_toc on OSX is due to the difficulty to estimate what should be the Systime() passed. With the new tictoc, more accurate and faster, the current test procedure is bound to fail sometimes.

PR #1640, opened 2023-09-14
versioned subdirectory in release tarballs. Closes #1639

### LINKED ISSUE #1639: Release tarball should use a versioned subdirectory
I unpacked it and got a nasty surprise when it wrote into my git checkout.  Please use a versioned subdirectory like every other project out there.

slayoo:


opoplawski:
Looks to me I guess.  Is there an artifact to test?  Thanks.

opoplawski (approved):

slayoo:
We will need to wait till the weekly release to check it.
Let me merge then.
Thanks for reporting @opoplawski!

Issue #1662, opened 2023-11-08
bug on OSX "only" in some cases in new fast GAUSSFIT


alaingdl:
No problem up to now on Linux, 2 very different OSX gave the same error (b1)

```
pro TEST_BUG_GAUSSFIT

a = [4.0, 1.0, 2.0, 1.0]

n=100
x = (FINDGEN(n)-(n/2))/4
z = (x -a[1])/a[2]
y = a[0]*exp(-z^2/2)
y=y+0.01*randomn(123, n)

yf=gaussfit(x, y)

plot, x, y
oplot, x, yf, col=color2color(!color.green)

yf1=gaussfit(x, y, b1)
oplot, x, yf1, col=color2color(!color.red)

b=a+0.1
yf2=gaussfit(x, y, c, est=b, nt=4)
oplot, x, yf2, col=color2color(!color.orange)

print, b1
print, c

stop
end
```

alaingdl:
I wrote some tests covering more tests for GAUSSFIT() and the situation is not good !
-- some tests (when called without estimated values) are broken on OSX (but not on Linux)
-- few basic cases are now broken in Linux too

The good point is the code is very fast when it is stable !

alaingdl:
Since now we do have a "test_gaussfit.pro" and a "bench_gaussfit.pro" in Git, and I correct the code to take into account the case when the step in X is not unity (1), and I also improved the initial estimation of the a4/a5/a6 and the result is in #1671  I will close this issue.

Please open a new issue if any trouble with GAUSSFIT()

Sorry but few keywords now are not avaialble in the _fast_ GAUSSFIT, you can use the old one ("gaussfit_mpfit.pro" in src/pro) for that.

On larges tests I made, the new implementation is as fast and as accurate than the IDL one ...

Issue #1661, opened 2023-11-06
regrssion within revised TIC & TOc on OSX intel


alaingdl:
with current GIT version, on OSX 14.0 with intel proc I have : 
```
GDL> tic & TEST_TIC_TOC,/no & toc
Time elapsed in loop  0 :       0.13500881 seconds.
Time elapsed in loop  1 :       0.13453913 seconds.
Time elapsed in loop  2 :       0.13385391 seconds.
Time elapsed in loop  3 :       0.13280582 seconds.
Cumulative Time elapsed :       0.53620768 seconds.
Total Time elapsed      :       0.56858802 seconds.
% TEST_TIC_TOC: ====================================================
% TEST_TIC_TOC: =                                                  =
% TEST_TIC_TOC: =  1 errors encountered during TEST_TIC_TOC tests  =
% TEST_TIC_TOC: =                                                  =
% TEST_TIC_TOC: ====================================================
% Time elapsed : 1.6992762e+09 seconds.
```

2 problems. Last line  is wrong on OSX (OK on Linux). And the new version on OSX triggered a error.
Those 2 problems did not exist on the Mac with the previous version (before PR #1646 )

None of those problems arise on Linux (before and after change in TIC/TOC)

GillesDuvert:
The test is not precise enough , now. On OSX the overhead of loops seems to be greater that 10-2 for the test. Which may be another issue. It is better to check that the variance of the val_times is better than 1e-3.

PR #1663, opened 2023-11-10
Trying to solve GAUSSFIT issue

alaingdl:
This patch should solve #1662. It is working now on Linux and OSX. some code will be discarded ASAP. Few others details solved. I am very sorry but no test_gaussfit up to now :disappointed: , but I have a draft :smiley_cat: 

When we will have more time, it would be good to fix some of the warnings triggered by the use of the -Wall flag, especially in the core code ... (see my small change in `dimension.hpp` : 
```
54c54
<     for(SizeT i=0; i<rank; ++i)
---
>     for(char i=0; i<rank; ++i)
```
)

### DIFF
--- a/src/dimension.hpp
+++ b/src/dimension.hpp
@@ -51,7 +51,7 @@ public:
   dimension(const dimension& dim_)
   { 
     rank = dim_.rank;
-    for(char i=0; i<rank; ++i)
+    for(SizeT i=0; i<rank; ++i)
       { dim[i]=dim_.dim[i];}
     stride[0] = 0; // not set
   }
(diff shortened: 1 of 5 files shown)


alaingdl:
problem only with the TIC/TOC reported in #1661

GillesDuvert:
the changes in `dimension.hpp` should be explained (even if apparently inocuous) (this is legacy touchy code).
the calls of` SizeT nParam = e->NParam(xxx);` in `la_trired_pro`and other functions of `gsl_fun`  is important, even if '`nParam`' is reported unused. by the compiler. In that case, replace, e.g., `nParam = e->NParam(3);` by `e->NParam(3);` avoiding creating a variable.

PR #1678, opened 2023-11-23
Create CITATION.cff (which makes GitHub display a "Cite this repo" button in the "About" panel)

slayoo:


alaingdl:
unclear for me what it will add and how it will be visible, but why not !

slayoo:
Thanks Alain, see the bottom of the screenshot below (taken after your merge)
DeepinScreenshot_select-area_20231124003719

PR #1683, opened 2023-12-11
Faster div and mod and report fpexceptions

GillesDuvert:
I removed signaling FP exception mechanism using sigsetjmp that was preventing parallelization of some operators.
I realized that this code was also ignoring the !EXCEPT value. 
Now we just registers FP exceptions in a gdl-owned static flag integer, that happen between a 'start of register' and 'end of register' instruction ( or by creating a unique_ptr doing the job) . Which must be called only at points we know there could be, and want to check, FP exceptions. See sqrt() or total()
The mechanism obeys to the values of !EXCEPT as per the documentation.
This in turn permits to speedup the divisions and Modulos for GDL variables.

Also, adde some tests that were missing to speed up operators when the operand is of size 1


alaingdl:
@GillesDuvert How could I help to test that ? Just try to compile on various machines and run the tests ?! (I have some time next days)

GillesDuvert:
This is all about speed gains, even if the reporting of errors like divide by zero is better. Hopefully without breaking anything. Mostly on all divide functions ( a/=2 or b=A/c) but also mod() and sqrt() where the old code was preventing parallelism. Plus a few speed improvements that I've tested on my machine (some case of total() and product() over a dimension, and integer divides, using libdivide. Difficult to check indeed.

PR #1703, opened 2023-12-30
Shmmap etc

GillesDuvert:
WIP but can be tested


brandy125:
Very nice!

brandy125:
Thanks @GillesDuvert !!!
I tested it and it works as expected! 
Very nice!

alaingdl:
@GillesDuvert wouhaouh 

I cannot compile on Debian due to relic from the saverestore redefinition of xdr_u_int32_t
(unclear for my why this remain ...)

I cannot link on OSX : 
```
[100%] Linking CXX executable gdl
ld: library not found for -lrt
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

But I can compile & test on u2204 ! And the demo provided by @brandy125 is working find for /long and /double ! (but I succeed to crash by after to another local variable the same global one)

GillesDuvert:
The support is very minimal and risky: there is no mechanism to avoid collisions, the detaching and destroying of shared areas is unsafe at best. @brandy125 for serious work and for the time being you need to have only one process writing and all the others reading. 
I suppose it is best to insure this minmal safety first before adding more option such as memory mapping real disk files  (works with some data files like FITS).

Incidentally, this is a way to share data with IDL as the memory mapping of simple arrays is the same.

alaingdl:
On u2204 and Debian 10, I can reproduce most of the problems in Codecov : 
```
GDL> test_finite
munmap_chunk(): invalid pointer
Abandon (core dumped)

GDL> test_size
[...]
free(): invalid pointer
Abandon (core dumped)


GDL> .r test_isa
...
double free or corruption (out)
Abandon (core dumped)


```

GillesDuvert:
Indeed, last hack had a flaw. Corrected. There will be a possibility of memory leak for rare cases until I set up a better scheme, but this will need a change in the BaseGDL objects.

alaingdl:
@GillesDuvert tested with success on x86_64 for Debian10, u2204 and OSX 12.3 BUT on OSX I need to remove "-lrt" in src/CMakeFiles/gdl.dir/link.txt to got the link

no more crash.

Once the small problem on OSX fixed, I vote for a merge !

alaingdl:
@GillesDuvert Linked perfectly on OSX 12.3 ! Still linking on Debian 12 ;) All tests OK on both :)

It is a GO for me for a 1.0.5 !

GillesDuvert:
OK, probably quite complete. Windows support should be easy afterwards.

Issue #1709, opened 2024-01-05
file_readlink really working?


GillesDuvert:
see https://github.com/gnudatalanguage/gdl/discussions/1708#discussioncomment-8021770

alaingdl:
Unclear for me why the code is broken, if I change around line 2206 in file .cpp
```
       if( readlink(symlinkpath, actualpath, PATH_MAX) != -1)
	  actualpath[lenpath]='\0';
 
```
into
```
	int pid=readlink(symlinkpath, actualpath, PATH_MAX);
	cout << " pid " << actualpath << endl;
```

I got the right answer ... (doing ls -lart /proc/"pid"/ I can see it is the good one)

alaingdl:
In the line after, `lenpath` is 0 (zéro), CQFD ! (QED)
then the logic around is wrong !

alaingdl:
my snippet is wrong but I think I have solved the code and I have few tests in mind

alaingdl:
Should be solved in PR #1713 
Please open a new issue if related problem
I did not had a lot of idea for tests ...

PR #1713, opened 2024-01-05
Tentative anwsers to #1708 (call_external(/cdecl) and to #1709 (file_readlink('/proc/self'))

alaingdl:
This PR should solve  #1708 

Keyword /CDECL in `call_external([...] /cdecl)` should be silently ignored on OSX & Linux.
Please contribute on MSwin

It should also solve #1709
(`file_readlink('/proc/self')` on Linux)
I add some tests dedicated to Linux & OSX, please test & contribute for MSwin & *BSD
Any other ideas welcome !

### DIFF
--- a/src/basic_pro_jmg.cpp
+++ b/src/basic_pro_jmg.cpp
@@ -5,7 +5,7 @@
     copyright            : (C) 2004 by Joel Gales
                          : (C) 2010 by Christoph Fuchs (CALL_EXTERNAL)
     email                : jomoga@users.sourceforge.net
- ***************************************************************************/
+***************************************************************************/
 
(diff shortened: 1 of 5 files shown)


GillesDuvert:
@alaingdl there seem to be 2 "cout<<" left in file.cpp (debug )
Congrats for the quick patch!

alaingdl:
> @alaingdl there seem to be 2 "cout<<" left in file.cpp (debug ) Congrats for the quick patch!

in fact, those two "cout <<" are outside the field of my change, but I  can comment them

alaingdl:
no need to test again ... since only two cout << removed ...

Issue #1717, opened 2024-01-08
The size of the title in a plot window can not be adjusted using the "charsize" keyword


brandy125:
Seems that the size of the title in a plot window can not be adjusted with the "charsize" keyword

```
GDL> plot,findgen(10),title='I can not adjust the size of the title',charsize=5
```

alaingdl:
ha ! see from here, seems to appear between 231020 and 231114. "just" 24 have been changed ...

GillesDuvert:
...And other variations of [XY]charsize are wrong too.

GillesDuvert:
thanks, corrected now!

alaingdl:
Yes, solved. 
But the exact sizes of the plot/titles may help to fix few regressions we can see when running test_coyote
(gdl surface of the plotting region is smaller by 20% and the charsize is ~20% bigger )

GillesDuvert:
No, that's not true on my machine. The sizes are OK within 1% even with test_coyote.

PR #1721, opened 2024-01-09
solving #1717

GillesDuvert:


alaingdl:
totally unclear for me why the window part is not done. Seems that a link is not done at the end of the compilation.
first test is done but not really related to GDL exe it-self ?!

GillesDuvert:
the windows build crash on tests approx 3/4 of the time, but in fact everything is OK, one can even retrieve the installer, install on a fresh mwindows machine, run the tests and they work. So I'm inclined to blame the test-running code and forget about it.

PR #1741, opened 2024-01-30
Allanvar, MedAbsDev + allowing compilation without  Eigen3

alaingdl:
For me it is important than GDL can still be compilable when some optional dependences (X11, wxwidgets, Eigen3, HDF5, ...) are not available. I managed here the case with Eigen3 (eg. : IM/GM packages are broken on my Ubuntu 20.04 then it is convenient to flag them off --never succeed to fix the apt/packages issues--)

And now if the functions based on Eigen are call, a smart message is issue if compilation without Eigen ...

Later I will add CHOLDC/CHOLSOL  but I did not finish rewriting of the tests.

And I also add two useful functions : computing Allan variance in 1D array, and also normalized Median Abs. Dev.
Contributions on Allan var. code very welcome.


alaingdl:
I have to add another change but no time now, please no merge now

GillesDuvert:
@alaingdl please merge at your convenance.

PR #1753, opened 2024-02-19
obtaining version id from git metadata (no more need for version bump commits or release branches)

slayoo:


alaingdl:
Thanks @slayoo 
Hum, I have a problem here, I go : `Version v1.0.2-533-g851d3486`

slayoo:
I'm still working on it...

>  Hum, I have a problem here, I go : Version v1.0.2-533-g851d3486

This means 533 commits after the v1.0.2 tag, and the actual commit is `851d3486`.
It refers to v1.0.2 and not v1.0.4 likely because v1.0.4 tags a commit that does not belong to the `master` branch.
If we merge this PR (after it's finalised), the new releases will not need any version-bump commits, and hence will be created out of the master branch, so the version will be marked as `v1.0.5-1-gXXXXXX` (in case there would be one commit on top of the release). In case, the compilation would be done with some uncommitted  code changes, the version will be marked additionally as `-dirty`.

slayoo:
@pjb7687 @GillesDuvert feedback welcome!
This would help to make releasing new versions of GDL not require creating a separate branch and not requiring committing version string bumps - clicking "Publish Release" at https://github.com/gnudatalanguage/gdl/releases/new should be enough.

slayoo:
Thanks @GillesDuvert 
Created and self-assigned an issue to double-check if the release tarballs are ok after this merge: https://github.com/gnudatalanguage/gdl/issues/1757

PR #1765, opened 2024-03-03
Add udunits xmls to windows distribution

### LINKED ISSUE #1732: Windows installer does not ship UDUNITS XML files
CC: @pjb7687, @jdtreffert, @GillesDuvert, @acoulais   

reported here: https://github.com/gnudatalanguage/gdl/discussions/1723

the installer does not ship XML definition files which are needed to use UDUNITS routines.

errors happen at runtime, e.g. when trying to use `IMSL_CONSTANT`

things that are needed to be done:
- include these XML files with the installer
- add CI test logic that would be running GDL unit tests with the installer-installed GDL

pjb7687:
Should fix #1732, #1723

### DIFF
--- a/scripts/build_gdl.sh
+++ b/scripts/build_gdl.sh
@@ -513,6 +513,12 @@ function install_gdl {
         if [[ -n "${PYTHON_DLL}" ]]; then
             cp -f "${PYTHON_DLL}" bin/
         fi
+
+        if [[ $WITH_UDUNITS2 == "ON" ]]; then
+            log "Copying udunits xml files to install directory..."
+            cp -rf /${mname}/share/udunits share/
+        fi
+
 #        #this ensures that we overcome the bug in plplot version for Windows that does not know about PLPLOT_DRV_DIR env. var.
         log "Copying our drivers to same directory as gdl.."
         if [[ -n "${PLPLOT_DRV_DLL}" ]]; then


pjb7687:
image

Seems to be fixed with this.

slayoo:
@pjb7687 how could we check it on CI if that works?

slayoo:
(and first, that it fails with current `master` code)

GillesDuvert:
@pjb7687 Thanks, relevant files are in the installer . Not obvious how to check if they will be found by the executable when installed on a virgin Windows, though (need a specific PATH?).
BTW, it should be the same problem on OSX, the DMG does not have those files.

pjb7687:
> @pjb7687 how could we check it on CI if that works?

I tried to write a `pro` test code to detect this but weirdly GDL gives exit code 0 when the error occurs, and the error cannot be detected. I am trying out other options I will keep you posted when I make some progress on it

pjb7687:
> @pjb7687 Thanks, relevant files are in the installer . Not obvious how to check if they will be found by the executable when installed on a virgin Windows, though (need a specific PATH?). BTW, it should be the same problem on OSX, the DMG does not have those files.

It is easy to ship the XML files within the DMG, but it is super tricky for the dylibs to point that xmls within the package. On Mac, all the paths are absolute in the dylib and initially I had to patch their rpath (it seems gdl now uses dylibbundler instead which looks nice, but I assume it does basically the same thing). The problem is that in the dylibs, the paths to resources are still absolute, and this won't be easy to patch. Actually that was the main reason I was hesitant to merge the DMG PR... This is not only a problem with udunits but all other dylibs rely on external resources.

pjb7687:
Sorry guys for my delayed answer. I recently had a budget cut by the Korean government (https://www.nature.com/articles/d41586-024-00525-7) and have been super busy writing grants to secure money for my students. The progress would still be very slow, but I will keep you posted once this is done.

And I think I am going to release 1.0.5 after merging this.

GillesDuvert:
> > @pjb7687 Thanks, relevant files are in the installer . Not obvious how to check if they will be found by the executable when installed on a virgin Windows, though (need a specific PATH?). BTW, it should be the same problem on OSX, the DMG does not have those files.
> 
> It is easy to ship the XML files within the DMG, but it is super tricky for the dylibs to point that xmls within the package. On Mac, all the paths are absolute in the dylib and initially I had to patch their rpath (it seems gdl now uses dylibbundler instead which looks nice, but I assume it does basically the same thing). The problem is that in the dylibs, the paths to resources are still absolute, and this won't be easy to patch. Actually that was the main reason I was hesitant to merge the DMG PR... This is not only a problem with udunits but all other dylibs rely on external resources.

I had to switch to dylibbundler at some point

PR #1774, opened 2024-03-07
introducing export-ignore directive for all files to make github-generated tarballs emtpy

slayoo:


GillesDuvert:
@slayoo you can merge I presume?

slayoo:
thanks, let's try it!

Issue #1777, opened 2024-03-10
Problem with "/=" operator in 1-element arrays


jaymurthy:
I'm getting the wrong answer from adxy. I can't attach .sav files so I attach the fits file with the header which I read with mrdfits.
[fNCOB15.fits.gz](https://github.com/gnudatalanguage/gdl/files/14550525/fNCOB15.fits.gz)

adxy,hdr,80.4112,42.1945,x,y
print,x,y
       180.78857       319.06283
where x and y should be 60,60.
I've tracked it to a line in wcssph2xy. 
GDL> print,radeg
       57.295780
GDL> phi /= radeg
GDL> print,phi
       88.402565
GDL> phi /= radeg
GDL> print,phi
       88.402565
Oddly it works when I type it in a new instance of gdl and it works in FL
GDL> phi=88.402565
GDL> radeg=57.295780
GDL> phi /= radeg
GDL> print,phi
      1.54292

I'm using GDL - GNU Data Language, Version v1.0.2-534-g20545ef2 on Sonoma 14.3.1 with the M2 chip.

jaymurthy:
SIMPLE  =                    T / Written by IDL:  Wed Jan 24 15:29:30 2024
BITPIX  =                  -32 / Number of bits per data pixel
NAXIS   =                    2 / Number of data axes
NAXIS1  =                  121 /
NAXIS2  =                  121 /
EXTEND  =                    T / FITS data may contain extensions
DATE    = '2024-01-24'         / Creation UTC (CCCC-MM-DD) date of FITS header
COMMENT FITS (Flexible Image Transport System) format is defined in 'Astronomy
COMMENT and Astrophysics', volume 376, page 359; bibcode 2001A&A...376..359H
COMMENT Diffuse background
CRVAL1  =              80.4112 /
CRVAL2  =              42.1945 /
CDELT1  =            -0.100000 /
CDELT2  =             0.100000 /
CTYPE1  = 'GLON-TAN'           /
CTYPE2  = 'GLAT-TAN'           /
CRPIX1  =              60.0000 /
CRPIX2  =              60.0000 /
HISTORY Written on Wed Jan 24 15:29:30 2024
BUNIT   = 'photon units'       /
END

jaymurthy:
![iTerm2 - gdl 2024-03-10 at 7 56 32 PM](https://github.com/gnudatalanguage/gdl/assets/12696836/e9763f74-487e-4070-bdfe-ed9ef1a399ad)

[...]

PR #1779, opened 2024-03-11
add basic test for #1777 (/= operator issue)

slayoo:


alaingdl:
1: I have a more general test (but not working for negative due to modulo with Int)
2 : more important, I think it is better to name the test following the kind of the main problem
we address in the test (I will rename and merge few other tests related to ForEach)

GillesDuvert:
test_indepth_basic_functions.pro in #1784 not only check this but approx. 6100 other possible cases for all  basic operators.

slayoo:
Thanks @GillesDuvert !

PR #1780, opened 2024-03-11
strip first char of version string for !GDL.RELEASE if it is "v"

slayoo:


alaingdl:
I need few time to do it quietly ! I hope to be able to review and test possible side effects this Monday

slayoo:
thanks Alain

alaingdl:
Since I change `gdl_version.pro` in #1786 it is OK for me to merge that.

slayoo:
Thanks @alaingdl

PR #1784, opened 2024-03-17
Check all basic operators results in all possible cases

GillesDuvert:
added test_indepth_basic_functions.pro and modified code to insure ~1200 tests on OPERATORS are similar to IDL.

### DIFF
--- a/src/basic_op.cpp
+++ b/src/basic_op.cpp
@@ -1730,40 +1730,10 @@ Data_<Sp>* Data_<Sp>::AndOp(BaseGDL* r)
 }
 // different for floats
 
-template<class Sp>
-Data_<Sp>* Data_<Sp>::AndOpInv(BaseGDL* right) { TRACE_ROUTINE(__FUNCTION__,__FILE__,__LINE__)
-  return AndOp(right);
-}
-// for floats
-
 template<>
 Data_<SpDFloat>* Data_<SpDFloat>::AndOp(BaseGDL* r) { TRACE_ROUTINE(__FUNCTION__,__FILE__,__LINE__)
 #include "snippets/basic_op_AndOpCplx.incpp"
 }
-
-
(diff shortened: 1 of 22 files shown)


GillesDuvert:
The `test_indepth_basic_functions.pro` brutally explores 95 different operators defined in files `basic_op_xxx.pro`, for a combination of the following cases: 
1. The 10 different numeric operand types (BYTE to DCOMPLEX)
2. The size of each operand: scalar, array of 1 element, array of > 1 element
3. The respective sizes of the two operands for operators using two operands (like in `c= a + b` ), the code is not the same when `a` is smaller of greater than `b.`
4. The persistence of the operand, like in `c = a + temporary(b)`

Because each of these cases need a special code, for speed and memory optimization (this was written by Marc in the very first version of GDL, at least for 90% of the operators), the number of combinations explored is around 11000 cases.

The result of the procedure is a ~24000 lines text file that can be compared with an equivalent IDL output.
The present PR introduces a number of code patches or additions to insure even obscure operator+operand combination give the same result as IDL.
The GDL and IDL produced files are identical, with the exception of the _signedness_ of a few NaN or Inf results in case or zero division of complex values.

As each operator is also optimized to use parallel processing on operands as soon as the number of elements is greater than `!CPU.TPOOL_MIN_ELTS`, the test_indepth_basic_functions.pro procedure can also be used to check that the parallelized results are identical to the single cpu ones (*). This has been checked too. _Note that testing the equivalence of results between single and many cpu as performed in the  `test_indepth_basic_functions.pro` procedure with IDL will... crash IDL on some complex values, so this part of procedure is deactivated with IDL._

(*) TPOOL (Thread Pool) using functions cannot in general be expected to provide exactly the same result in parallell or not parallell mode: TOTAL() is an example. But operators do.

PR #1785, opened 2024-03-19
close #1781 (TICK_GET error)

GillesDuvert:
oops getLabelingValues was wrongly clearing values.
Also, insure only the 1st axis values are reported

Issue #1725, opened 2024-01-15
LIST() and HASH() don't have the rigth SIZE(/n_dim)


alaingdl:
Thanks to some super simple tests in error in `test_isa.pro`, I got : 

```
GDL> size(hash(),/n_di)
           1
GDL> size(list(),/n_di)
           1

[...]
IDL> size(hash(),/n_di)
           0
IDL> size(list(),/n_di)
           0
```

alaingdl:
This one is probably solved, but the side effect of large blind tests is I found new problems in ISA()

alaingdl:
Should be closed in #1786 
Please open a new issue if related problem emerge ...

PR #1786, opened 2024-03-19
various small tests ...

alaingdl:
OK, this is a merging of various codes in GDL syntax for tests : 

* test_xyztick_get.pro tests [xyz]tick_get for PLOT & SURFACE, with various combinations of fields and ranges
(@GillesDuvert that shows PR #1785 is not the definitive one)
* test_operators.pro tests /=, *=, +=, -=. Might be redundant with test_indepth_basic_functions.pro
* gdl_version.pro does manages the new version schema #1780 (for naming convention in i.e. bench out files)
* test_foreach.pro merges some tests related to FOREACH (SF 273, 630 and test_bug_3288652.pro (removed outside)
* bench_loops.pro

bench_loops_Linux_planck4

### DIFF
--- a/src/basic_fun_jmg.cpp
+++ b/src/basic_fun_jmg.cpp
@@ -207,8 +207,8 @@ namespace lib {
     }
 
     if(type == "OBJREF"){
-      rank=1; // alway array following ISA() doc.
-      //cout << "OBJREF" << endl;
+      // AC 2024/03/21 wrong here rank=1; // alway array following ISA() doc.
+      // cout << "OBJREF" << endl;
       DObjGDL* obj = static_cast<DObjGDL*>(p0);
       DObj objID = (*obj)[0];
       if(objID == 0) res = false; else res = true;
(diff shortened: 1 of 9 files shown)


alaingdl:
* correcting ISA() and SIZE() for HASH() & LIST() (#1725)
* wrong name in the final message of `test_xyztick_get.pro`
*  new entry in `test_isa.pro` for missing /file keyword

GillesDuvert:
`test_operators.pro tests /=, *=, +=, -=. Might be redundant with test_indepth_basic_functions.pro`
yes, operators are there for "completion" but the real test that visit every possible line of code is indeed `test_indepth_basic_functions.pro`.
probably not worth the work to clean those duplicates.
and `test_indepth_basic_functions.pro` is really `test_indepth_all_operators.pro`.

GillesDuvert:
`test_xyztick_get.pro` has been invaluable in realizing AutoTick was wrong, thanks @alaingdl .

PR #1793, opened 2024-03-29
Use new python initialisation, if old one is deprecated

jkohnert:
See [issue](https://github.com/python/cpython/issues/88279), and [PEP 587](https://peps.python.org/pep-0587/).

Since Python version 3.11 the old legacy initialisation functions are marked deprecated, the upgrade path is specified in the PEP. GDL uses those deprecated functions in two locations, this PR updates those locations to use the new API in case Python version is 3.11 or above.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,9 @@ src/gdl
 testsuite/launchtest.c
 nbproject/
 build/
+cmake-build*/
 install/
 gdlde/
 gdlde*.zip
 .vscode
+.idea/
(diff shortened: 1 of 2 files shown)


jkohnert:
I just had another look: The Windows build uses Python-3.12 at the moment, this patch successfully removes the deprecation warnings for using `PySys_SetArgv` during this build. On Arch Linux (I'm using it on my dev machine) those warnings are removed as well.

PR #1794, opened 2024-03-29
Update codecoverage to use actions/checkout@v4

jkohnert:
Fix Node-16 depcrations message for codecoverage.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -18,7 +18,7 @@ jobs:
       ROOT_DIR: ${{ github.workspace }}/..
     steps:
       - name: Checkout GDL
-        uses: actions/checkout@v3.5.2
+        uses: actions/checkout@v4
         with:
           submodules: recursive
           fetch-depth: 0


alaingdl:
Could you please explain a little mode or give a like to document that, thank !
(is that link relevant ? https://github.com/actions/setup-node/issues/850 )

jkohnert:
Pretty much, yes. The codecov job issues a warning:
```log
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3.5.2. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

Reading though this documentation leads to [this page](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-versioned-actions) suggesting v4 (or, v4.2.0, if one wants to be more specific).

I'm not too familiar with github's actions, since we're using gitlab at work, and I use gitlab privately, too, but as far as I can tell,this should be the way to go. 

I later saw, the probably outdated action is used at more places, maybe I should use this PR to update those usages, too. What do you think?

Best regards Jan

slayoo:
example here:
image
(https://github.com/gnudatalanguage/gdl/actions/runs/8485091812)

Thanks @jkohnert !

alaingdl:
Thanks you, I never noticed because I jump immediately to the red cross item !

PR #1799, opened 2024-04-06
to be thoroughly tested, should close #1796

### LINKED ISSUE #1796: Format with $ doesn't work as intended and as it used to
According to IDL documentation if the format code $ occurs anywhere in the format string, the new line implied by the closing parenthesis of the format string is suppressed. Hence the following consecutive print commands should provide the same results, but they don't:

printf, 1, format='($,I0)', 123
printf, 1, format='(I0)', 456

provides 123456 as it should, but 

printf, 1, format='(I0,$)', 123
printf, 1, format='(I0)', 456

provides 
123
456
which is wrong. I guess this bug was introduced during some recent bug fixes in the format interpretation.

GillesDuvert:


alaingdl:
what is the status of this PR (not clear for me, relation with the next one ?)

GillesDuvert:
> what is the status of this PR (not clear for me, relation with the next one ?)

they are independent - this one should probably be merged first.

alaingdl:
We do have to add a test on that ... I will do that ASAP.

In fact the test `test_formats.pro` gives a problem, this problem appeared between 2023/10/20 and 2023/11/14 and I suspect it is due to the line :

```
diff gdl-1.0.3git231114CMake/src/calendar.cpp gdl-1.0.3git231020/src/calendar.cpp
98c98
<     if ((DLong)Z+1 > 0) dow = ((DLong)Z+1) % 7; else dow = ((DLong)Z+1099) % 7; //just translate axis...
---
>     if ((DLong)Z > 0) dow = ((DLong)Z) % 7; else dow = ((DLong)Z+1099) % 7; //just translate axis...
```
@GillesDuvert do you remember the reason of this change ?

IDL and GDL older than 2023/10/20 :
```
print,2.455555555D6,format='(c())'
Sat Dec 25 01:19:12 2010
```
GDL since 2023/11/14
```
GDL> print,2.455555555D6,format='(c())'
Sun Dec 25 01:19:12 2010
```

In fact, I don't know why  `test_formats.pro` is not in the LIST file.

GillesDuvert:
This looks like an error/typo . The previous version was correct, as Julian 2455556L is a saturday, not sunday.
Can you correct directly in GH?

PR #1800, opened 2024-04-06
enable T TR and TL codes. Wonder why they did not work and, more, why…

GillesDuvert:
… the ANTLR code was special for TR and TL. removed this special treatment and did not see any problem arising elsewhere (passes tests). This should be thorougly tested.


alaingdl:
For sure I am OK to review but I need some time, and not sure when during next few days

alaingdl:
@GillesDuvert sorry it seems we have a conflit here (maybe my fault)

GillesDuvert:
oulah, the conflict is very serious, apparently github takes the WRONG source. This happened in the past. Cannot figure how it is possible. GRRRRRRRR I'll be obliged to resubmit after scrutinizing the b**** code.

PR #1816, opened 2024-04-22
new test_trired test; most KeywordSet() with numeric fixed; -DEIGEN3=off now OK; limited cleaning in eigenvalues_solvers.*pp

alaingdl:
* just adding #include <limits> in typedef.cpp is enough to be able to compile "everywhere" when Eigen is turn OFF

* copyright headers of the last files written by Gilles ... (eigenvalues_solvers.*pp)
* very limited cleaning in those files (eg having Ix for the keywords index for easier reading (convention)) 
* better management of Eigen/GSL for these last codes
* When GDL compiled without Eigen ON, nice message as some other codes (LA_LEAST_SQUARES ...)
```
GDL> la_elmhes()
% GDL compiled without Eigen3 : LA_ELMHES not available
       0
```

* All remaining KeywordSet() with numerical values should have been cleaned, except in HDF and NetCDF files (may be I missed the case for 

* new test for TRIRED/TRIQL (`test_trired.pro` added in LIST)

What is not OK now from my point of view: LA_HQR missing, LA_TRIQL missing, still pending bugs in CHOLESKY codes

PR #1788, opened 2024-03-20
add Apple Silicon builds to CI (using macos-14 images); add WITH_OPENMP option to build_gdl.sh

slayoo:

PR #1821, opened 2024-04-24
Close 1809

GillesDuvert:
note this has also temporarily removed:

1. use of MPI on OSX
2. test_obj_new
3. test_xyztick_get

to make build clearer.
It is expected that future PRs will reinstall those options and tests.


GillesDuvert:
OK, OpenMPI IS NOT the culprit to the OSX failure and can be reenabled. Just the first victim, leaving its own forensic log.

alaingdl:
I am OK to release it, and solve later the details for OSX

I don't see how I will be able to merge my PR #1816 :(

slayoo:
@GillesDuvert IIUC, GitHub has made arm64 the default now for `macos-latest`. So we have the same failures as in #1788.
Let me change `macos-latest` to `macos-12` in the workflow files here to get back to previous setup.

GillesDuvert:
> @GillesDuvert IIUC, GitHub has made arm64 the default now for `macos-latest`. So we have the same failures as in #1788. Let me change `macos-latest` to `macos-12` in the workflow files here to get back to previous setup.

@slayoo indeed, this looks way better. In this case we could reenable OpenMP on macos12. MPI is now enabled for 'headless' only, which is probably reasonable, even if it was not really a problem. 

There is still a warning on CI page that I do not understand, as there is no 'node.js 20' stuff left in our code.

> **[build-windows (debug, x86_64, Debug)](https://github.com/gnudatalanguage/gdl/actions/runs/8845263876/job/24288822574)** 
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: 
> actions/setup-python@v4. For more information see: 
> https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

slayoo:
> There is still a warning on CI page that I do not understand, as there is no 'node.js 20' stuff left in our code.

The node.js warnings stem from code outside of our repo - from what's behind some of the workflow steps we reuse from external sources... but where is this particular warning displayed?

GillesDuvert:
annotations box in https://github.com/gnudatalanguage/gdl/actions/runs/8845263876

slayoo:
this should help with the warning: https://github.com/gnudatalanguage/gdl/pull/1824

Issue #1837, opened 2024-05-26
Problem with PRTARR() (was : pb with FITS Binary tables)


jaymurthy:
This started when I installed 1.0.6 and I've tried it with multiple FITS binary tables. Here is the link to one file: https://www.dropbox.com/scl/fi/vh2cq627zsrvwttsqviuo/FIG17.fits?rlkey=cjfye0btesovq573jg6qm9ezz&dl=0

I'm trying to isolate the error but haven't had success yet. It works ok with FL.

GDL> d=mrdfits('publicdata/FIG17.fits',0,hdr)
MRDFITS: Null image, NAXIS=0
GDL> d=mrdfits('publicdata/FIG17.fits',1,hdr)
% MRD_TABLE: Unable to dereference NULL pointer: <Expression> <POINTER   (<NullPointer>)>
% Error occurred at: MRD_TABLE         2320 /Users/jayanth/Dropbox/user/idluser/idllib/IDLAstro/pro/mrdfits.pro
%                    MRDFITS           2736 /Users/jayanth/Dropbox/user/idluser/idllib/IDLAstro/pro/mrdfits.pro
%                    $MAIN$
% Execution halted at: $MAIN$

jaymurthy:
I think it is something to do with line 458 in mrdfits.pro where it calls 
match, keyword, scales_names, mkey_scales, mscales, count = N_mscales
 The problem is because there are no matches (no TSCALnn) in the header.

jaymurthy:
The problem is because   scales = ptrarr(n_mforms, /all) (line ~446 in mrdfits.pro) gives
print,scales

(NullPointer)(NullPointer)

This is what it should be (in gdl 1.0.3 and in fl)
FL> print,scales
(PtrHeapVar1)(PtrHeapVar2)

The nullpointer causes it to break. Note that I replaced < with ( above to get it to show up.

alaingdl:
OK, thanks you @jaymurthy I can reproduce your bug :crying_cat_face: 

now we will try to solved it

alaingdl:
OK, the problem appeared between April 25 and 26, when I try to clean up the keywords (by name/by number)

alaingdl:
OK, I think I got it. In fact PTRARR() has one and only one keyword ... "ALLOCATE_HEAP"

alaingdl:
OK @jaymurthy , I solved it, I have the feeling we had a pending bug in the past ...
My problem now is to do a clean PR !

[...]

PR #1838, opened 2024-05-26
correcting #1837 PTRARR()

alaingdl:
Hum, my mistake

### DIFF
--- a/src/basic_fun.cpp
+++ b/src/basic_fun.cpp
@@ -422,17 +422,19 @@ namespace lib {
       throw GDLException("Array dimensions must be greater than 0");
 
     DPtrGDL* ret;
-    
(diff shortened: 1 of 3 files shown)


slayoo:
@alaingdl, let's add a unit test for this change! What would be a minimal code to show the problem (i.e., failing as of now, succeeding after this change is introduced?)

alaingdl:
@slayoo  yes your are right but I am so short in time next days ... Usually when I start a small test code then I find another detail (or bug) somewhere and so on and never succeed to close ..

@jaymurthy it should help for your problem

interesting that a new test case crashed in OSX ...

alaingdl:
OK, my correction is wrong and we did had a test case ! `test_heap_refcount.pro`
Wait a second !

PR #1840, opened 2024-05-29
rewriting test_elhmes to assess small numerical changes between x86 & M2

alaingdl:
all is in the title ! tolerance is 1e-5


slayoo:
@alaingdl, if not creating a new branch for each PR, you can use the `git rebase` command to bring this branch in sync with the upstream master branch (which contains a squash commit from your previous PR, but does not contain commits that were squashed). To do it, you can try:
```
git fetch upstream
git rebase upstream/master
git push -f
``` 
This way, this PR will have only the relevant commits.
HTH,
S.

Issue #1578, opened 2023-04-22
Build failure with ImageMagick


meerfrau:
Is this due to ImageMagick 7.1.1.7?
```
[ 63%] Building CXX object src/CMakeFiles/gdl.dir/magick_cl.cpp.o
In Datei, eingebunden von ./src/gdl/src/magick_cl.cpp:41:
./src/gdl/src/magick_cl.cpp: In Funktion »BaseGDL* lib::magick_ping(EnvT*)«:
./src/gdl/src/magick_cl.hpp:23:15: Fehler: »class Magick::Image« hat kein Element namens »alpha«
   23 | #define matte alpha
      |               ^~~~~
./src/gdl/src/magick_cl.hpp:23:15: Anmerkung: in Definition des Makros »matte«
   23 | #define matte alpha
      |               ^~~~~
./src/gdl/src/magick_cl.hpp:23:15: Fehler: »class Magick::Image« hat kein Element namens »alpha«
   23 | #define matte alpha
      |               ^~~~~
./src/gdl/src/magick_cl.hpp:23:15: Anmerkung: in Definition des Makros »matte«
   23 | #define matte alpha
      |               ^~~~~
./src/gdl/src/magick_cl.cpp: In Funktion »BaseGDL* lib::magick_create(EnvT*)«:
./src/gdl/src/magick_cl.hpp:23:15: Fehler: »class Magick::Image« hat kein Element namens »alpha«
   23 | #define matte alpha
      |               ^~~~~
./src/gdl/src/magick_cl.hpp:23:15: Anmerkung: in Definition des Makros »matte«
   23 | #define matte alpha
      |               ^~~~~
./src/gdl/src/magick_cl.hpp:23:15: Fehler: »class Magick::Image« hat kein Element namens »alpha«
   23 | #define matte alpha
      |               ^~~~~
./src/gdl/src/magick_cl.hpp:23:15: Anmerkung: in Definition des Makros »matte«
   23 | #define matte alpha
      |               ^~~~~
./src/gdl/src/magick_cl.cpp: In Funktion »BaseGDL* lib::magick_readindexes(EnvT*)«:
./src/gdl/src/magick_cl.cpp:390:48: Fehler: »MagickCore::PixelPacket*« {aka »MagickCore::_PixelPacket*«} kann nicht nach »MagickCore::Quantum*« {aka »float*«} in Initialisierung umgewandelt werden
[...]

PR #1850, opened 2024-07-07
Cmakefix

ChunkyPanda03:
I found a couple of errors while compiling I saw that this project uses imagemagick-7 not imagemagick-6 as shown here #https://github.com/gnudatalanguage/gdl/issues/1578#issuecomment-1519117266 and I think the reason it is not automatically looking for the correct version. I have added it into the the FindImageMagick. Also a slight note that debian 12 has only imagemagick-6 and cmake recommends that you install it. 

test_dynamic_drivers.pro had an error that would prevent the test from setting the resolution which I think was not exiting correctly I now have a few plotting tests failing now but I think this is a personal dependency issue I would like to personally resolve before discussing further. 

For the git ignore these files were added to the working directory by the setup script you can choose whether to keep this change or not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ gdlde/
 gdlde*.zip
 .vscode
 .idea/
+2020.2.zip
+qhull-2020.2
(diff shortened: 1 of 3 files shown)


GillesDuvert:
@ChunkyPanda03 thanks for the PR, just a doubt: you changed set_resolu to set_resolution in  test_dynamic_drivers but there should have been no error there, the abbreviation was long enough to avoid confusion?

ChunkyPanda03:
yeah @GillesDuvert I realized that set_resolution is the same as the default it was likely failing for me because of a dependency issue when building on my machine... still tracking that one down.

Issue #1852, opened 2024-07-11
Build fails with strict-aliasing violations


eli-schwartz:
I tried to build with the following *FLAGS to optimize the build: `-flto=4 -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing`

The -Werror=* flags are important to detect cases where the compiler can try to optimize based on assuming UB cannot happen, and miscompile code that has UB in it. strict-aliasing issues are always bad but LTO can make them even worse.

I got this error:
```
[149/169] /usr/bin/x86_64-pc-linux-gnu-g++ -DHAVE_CONFIG_H -I/var/tmp/portage/dev-lang/gdl-1.0.6/work/gdl-v1.0.6/src -I/usr/include/tirpc -I/usr/include/ImageMagick-7 -I/usr/include/eigen3 -I/var/tmp/portage/dev-lang/gdl-1.0.6/work/gdl-v1.0.6 -I/var/tmp/portage/dev-lang/gdl-1.0.6/work/gdl-v1.0.6/src/plplotdriver -I/var/tmp/portage/dev-lang/gdl-1.0.6/work/gdl-v1.0.6/src/whereami/src -I/var/tmp/portage/dev-lang/gdl-1.0.6/work/gdl-v1.0.6_build  -pipe -march=native -fstack-protector-all -O2 -fdiagnostics-color=always -frecord-gcc-switches -flto=4 -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing  -Wformat -Werror=format-security -fopenmp -std=gnu++11 -fPIE -MD -MT src/CMakeFiles/gdl.dir/saverestore.cpp.o -MF src/CMakeFiles/gdl.dir/saverestore.cpp.o.d -o src/CMakeFiles/gdl.dir/saverestore.cpp.o -c /var/tmp/portage/dev-lang/gdl-1.0.6/work/gdl-v1.0.6/src/saverestore.cpp
FAILED: src/CMakeFiles/gdl.dir/saverestore.cpp.o 
[...]

PR #1860, opened 2024-07-16
Adress problem reported in #1852

GillesDuvert:


GillesDuvert:
hm hdf5 regression...

PR #1870, opened 2024-08-02
a version supporting IDL_IDLBridge objects and methods.

GillesDuvert:
This is an hybrid solution hesitating between command-line tools (gmem_xxx commands) , an idl_idlbridge__init.pro containing methods, some C++ object definition, and a specialized eventloop that probably needs to be generalized.
OK on linux & OSX
At the moment, does not compile under Windows.

inter-process communication solution based on signals will certainly prove its limits, so a second version more robust (and, with somme effort, windows-compatible) should be done using only POSIX messages.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -62,6 +62,7 @@ exists_fun.cpp
 fftw.cpp
 file.cpp
 fmtnode.cpp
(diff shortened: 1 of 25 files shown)


GillesDuvert:
hm the thing works only in debug mode. Not yet ready, sorry.

PR #1897, opened 2024-10-29
All cases of FFT are now supported by fftw.cpp

GillesDuvert:
All cases of FFT are now supported by fftw.cpp
GDL FFTW now ONLY when FFTW library is present. (old slow GSL code removed)

### DIFF
--- a/src/fftw.cpp
+++ b/src/fftw.cpp
@@ -43,13 +43,18 @@ namespace lib {
 	T* res;
 	BaseGDL* data;
 	Guard<BaseGDL> guard_data;
-
+    
 	// if recenter and inverse (direct > 0) we will work on a "de-centered" p0 variant.
 	// and of course not center the result.
 
 	if (recenter && sign == FFTW_BACKWARD) {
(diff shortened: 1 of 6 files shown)


GillesDuvert:
hm thanks to test...

GillesDuvert:
OK. **But now we have a problem:**
According to FFT documentation: "_If the CENTER keyword is also set, then only the dimension given by the DIMENSION keyword is shifted to the center. The other dimensions remain unshifted_."

This is NOT WHAT IDL DOES. (but GDL does, now)
Proof:
`plot,ABS(FFT(FINDGEN(100,10),dim=1,/cen))`
if the 2nd dimension is not shifted (as per the documentation) then the centered peaks should be in linear increasing order. This is true for GDL, but in IDL the peaks have been clearly shifted on the x-axis.

Bug report to IDL? Or reproduce this "feature" in GDL?

GillesDuvert:
see #1898 . Our functionality did not change. good.

Issue #1907, opened 2024-11-07
Test failures on Arch Linux


jkohnert:
Hi,

I'm currently trying to update the package for Arch Linux (as I'm the maintainer in the [aur](https://aur.archlinux.org/packages/gnudatalanguage)).

However, there are failing tests:
```log
159/214 Test #159: test_postscript.pro ................***Failed    1.24 sec
[kohni-mobil:31526] shmem: mmap: an error occurred while determining whether or not /tmp/ompi.kohni-mobil.1000/jf.0/2996699136/shared_mem_cuda_pool.kohni-mobil could be created.
[kohni-mobil:31526] create_and_attach: unable to create shared memory BTL coordinating structure :: size 134217728 
% Compiled module: TEST_POSTSCRIPT.
% Compiled module: GDL_IDL_FL.
% TEST_POSTSCRIPT_BASIC: writing file : GDL_ps_basic_1.ps

*** PLPLOT WARNING ***
Unable to open cmap0 file cmap0_default.pal


*** PLPLOT WARNING ***
Unable to open cmap0 file cmap0_default.pal


*** PLPLOT WARNING ***
Unable to open cmap1 .pal file cmap1_default.pal


*** PLPLOT WARNING ***
Unable to open cmap0 file cmap0_default.pal


*** PLPLOT ERROR, IMMEDIATE EXIT ***
Unable to either (1) open/find or (2) allocate memory for the font file
Program aborted
```

All failing tests mention this error, and as far as I could find, it could to be related to [OpenMPI](https://github.com/open-mpi/ompi/issues/12784), but I'm not yet sure.

I checked, there is enough space in /tmp (16GiB); so space-related problems do not seem to be the culprit. However, Arch currently has OpenMPI 5.0.5.

The failing tests are:
```log
Total Test time (real) = 313.53 sec

The following tests did not run:
         21 - test_bug_3055720.pro (Skipped)
         22 - test_bug_3057511.pro (Skipped)
         23 - test_bug_3057520.pro (Skipped)
         24 - test_bug_3061072.pro (Skipped)
         30 - test_bug_3100945.pro (Skipped)
        145 - test_netcdf.pro (Skipped)

The following tests FAILED:
          2 - test_all_basic_functions.pro (Failed)
         15 - test_bug_2610174.pro (Failed)
         31 - test_bug_3104214.pro (Failed)
         48 - test_bug_3394430.pro (Failed)
         54 - test_bug_3595172.pro (Failed)
[...]

PR #1913, opened 2024-11-20
make plInBuildTree() work in GDL's fork of PLplot

jkohnert:
This is a proposed fix for #1907. 

Note: I deleted plplot_config.h since it is generated by CMake and so IMHO should not be in VCS.

Best, Jan


jkohnert:
Stumbled over two more files that are generated and thus should not be in VCS.

jkohnert:
And another auto-generated file deleted.

Issue #1857, opened 2024-07-13
there is no IDL_IDLBridge equivalent 


ChunkyPanda03:
I would like this feature however I don't think I could implement this

ChunkyPanda03:
we are destined to only use one thread for all of eternity :cry:

ChunkyPanda03:
@GillesDuvert was doing some work on this in purely idl/gdl I keep trying to use spawn but I just can't figure out a way to detach it from the main thread. I just don't think it is possible without writing some intermediary in C/C++ that handles the GDL instance what is your opinion on this?

GillesDuvert:
@ChunkyPanda03 what IDL bridge?... As GDL is opensource, one does not necessary need one.
Concerning SPAWN, I'm sure that /NOWAIT could be supported by GDL on all platforms with some work.
People seem happy running many GDLs in parallel, eventually sharing common memory areas (although mmap is very recent and untested) . I understand that huge time-consuming data pipelines (sometimes unofficially :smile: ) use GDL in such manner.

ChunkyPanda03:
@GillesDuvert apologies I did not clarify IDL had a feature where you can automatically open another IDL interpreter in parallel this is known as IDL_IDLBridge, GDL lacks an equivalent feature (GDL_GDLBridge ?). This feature is great for obtaining multithreaded behavior in IDL. In my case I often need to do large analysis procedures which includes rendering png files which need to be off the main thread or the whole thing could take upwards of a couple days otherwise. This is also necessary in idl for window applications because the window will lock up if anything in the main thread is running.

ChunkyPanda03:
here is a link referencing the idl feature https://www.nv5geospatialsoftware.com/docs/idl_idlbridge.html

brandy125:
Yessss, this would be nice to have!

GillesDuvert:
interesting, and indeed useful --- probably just having GDL listen to external commands in its command loop , and adding a subprocess creation by fork() , plus some script for managing the IDL_IDLBridge object.
Anybody interested? :smile:

ChunkyPanda03:
I sure am

GillesDuvert:
nearly there... :smile:

GillesDuvert:
@ChunkyPanda03 @brandy125  please test #1870 
linux only --- works4me in the simple case I have tested.
**BUT OK in DEBUG mode only -- optimization make interprocess hang** 
IPC needs to be more robust ...

[...]

PR #1917, opened 2024-11-24
Removed SIGCHLD -> SIG_IGN; added wait in gdl2gdl.cpp subprocess handling

jkohnert:
This is my attempt to fix the ECHILD/`wait()` problem causing test errors in test_hdf5.pro. The main culprit was setting SIGCHILD to SIG_IGN in the main process to avoid zombie processes unintentionally breaking SPAWN.

Since `wait()` by default also waits for processes in another thread the proposed fix ist to augment the `g2gAsynchronousReturnTrap()` running `wait()` in a loop, and aditionally do the work done before. I'm not fully sure this doesn't break something else, but at least it fixes the test on linux. Lets have a look on MacOS via pipeline.

Best Jan


jkohnert:
I made another attempt in gdl2gdl.cpp Since the map contains the pid, we know what processes to wait for and can use the WNOHANG option. Stopped children can then be removed from the map.

src/basic_pro.cpp
```diff
@@ -2107,43 +2107,46 @@ static DWORD launch_cmd(BOOL hide, BOOL nowait,
...
 
         // wait until child terminates
-        int status;
-        pid_t wpid = wait(&status);
+        int status = -1;
+        if (waitpid(pid, &status, 0) == -1) {
+          e->Throw(DString("Error waiting for child process: ") + strerror(errno));
+        }
 
+        // if waitpid fails, we still have status in the default of -1, indicating a failure
```
jkohnert: This is from an earlier version with a Warning issued instead of an exception being thrown.
```suggestion
```

src/gdl2gdl.cpp
```diff
@@ -103,13 +104,27 @@ static void g2gPerformCallbackProcedure(pid_t pid) {
...
+        // we're out of luck, wait returns an error, so break the loop and exit
+        std::cerr << "g2gAsynchronousReturnTrap exiting" << std::endl;
+        break;
+      }
+      // remove stopped children from the map
+      if (WIFEXITED(status) || WIFSIGNALED(status))
+        g2gMap.erase(g2gMapIter++);
+        // we need to manually count here, since we're modifying the map in the loop
+      else
+        ++g2gMapIter;
```
jkohnert: It seems the previous implementation expected some process (the child?) to take care of changing this map. If this is the case, this code introduces an error. It only works as intended, if the map is filled properly before the thread executing this function is started and the thread executing this code is the 
*only* one modifying it. Otherwise we would need a mutex and probably some more changes.

GillesDuvert: @jkohnert thanks for you efforts. It seems this version cannot run IDL's "idlbridge_demo" procedure anymore. Will have a look, but would not a quick patch be  to temporarily set SIGCHILD to something before the spawn and ressetting it to SIG_IGN just after? Chances that IDL_IDLBridge and SPAWN are run simultaneously are minimal. A more robust solution can be presented afterwards.

GillesDuvert: thanks @jkohnert !
However in the present state 'idlbridge_demo' (the demo for the [IDL_IDLBridge](https://github.com/gnudatalanguage/gdl/issues/1857#issuecomment-2229223204) new enhancement of GDL) does not work anymore (admitedly due to the use of signals and waits).
This can, at least temporarily, be avoided by setting proper signals hanling in SPAWN

GillesDuvert: Applied above suggested patches to your patch. Works.

Issue #1909, opened 2024-11-09
ROT accuracy issue


dinsmoro:
I noticed that `ROT` doesn't give the right outputs:

```
GDL> a=findgen(3,3)
GDL> print, a
      0.00000      1.00000      2.00000
      3.00000      4.00000      5.00000
      6.00000      7.00000      8.00000
GDL> print, rot(a,180.)
      5.00000      7.00000      6.00000
      1.00000      0.00000      3.00000
      1.00000      0.00000      0.00000
```

I tried some other older `ROT` functions from NASA personnel, but they performed the same which I suspect points to the `poly_2d` call (per `we use the c++ internal POLY_2D code to do the job` comment). I can't find that call in the repo, though.

This issue has popped up before, maybe system-dependent?:
https://cow.physics.wisc.edu/~craigm/idl/archive/msg03150.html

And IDL on a separate computer does give the expected values:

```
IDL> a=findgen(3,3)
IDL> print, a
      0.00000      1.00000      2.00000
      3.00000      4.00000      5.00000
      6.00000      7.00000      8.00000
IDL> print, rot(a,180.)
      8.00000      7.00000      6.00000
      5.00000      4.00000      3.00000
      2.00000      1.00000      7.34788e-16
```

alaingdl:
Thank you for this important bug report. I am sad that such an important bug was not detected before :(
And no test in the testsuite. Sorry.

dinsmoro:
Let me know if I can help - I'll take a stab at `poly_2d` if y'all need the help and you let me know where it is, but I can't guarantee anything since I don't know C++! But I can try!

alaingdl:
((I don't understand why we never realized that this code, which is big a complex, never gave problem before !))

OK, I found one problem in the `math_fun_jmpg.cpp` code, adding round() goes in the good direction.
I have another trouble within permutation of i & j index, TBC

```
          DLong px = (DLong) round(P[0] + P[1] * (DDouble) i + P[2] * (DDouble) i);
          DLong py = (DLong) round(Q[0] + Q[1] * (DDouble) j + Q[2] * (DDouble) j);
```

I also realized that we have no systematic tests related to that part of code (TBD)

GillesDuvert:
I've looked at it. There are indeed at lot of cases where the result is different, @dinsmoro 's one is only the tip of the iceberg.
[...]

PR #1919, opened 2024-11-26
New poly 2d

GillesDuvert:
https://github.com/gnudatalanguage/gdl/issues/1909 proved there was a problem in POLY_2D
furthermore there semed to be a speed issue with this command.

This is a rewriting of POLY_2D with special attention to 'border' values for 0 and linear interpolation ( TBC for cubic interpolation) with massive speed gains.

Added a preliminary test file.

Rewrote CONGRID.PRO to use POLY_2 as IDL does.

Plus a small add in basic_pro.cpp (start to make PREF_SET useful)


GillesDuvert:
Apparently a problem with OSX...

GillesDuvert:
I have acces to a mac OS 15.1 (Sequoia) and all the tests are OK. Do no understand why it does no pass on GH.

PR #1920, opened 2024-11-26
qhull should be present in all distros...

GillesDuvert:


GillesDuvert:
yepee, codecov compiles again.

Issue #1923, opened 2024-11-27
Bug for _F32 and _F64 datatypes in hdf5_fun.cpp


rsiddans:
There appears to be a bug in hdf5_fun.cpp in which GDL_ integer types are associated with various H5T_*_F32 and H5T_*_F64 data types via definitions of type_map like this...
/* GDL_LONG64 */
        { H5T_NATIVE_LLONG, H5T_IEEE_F64BE, H5T_IEEE_F64LE, H5T_INTEL_B64,
          H5T_INTEL_F64, H5T_INTEL_I64, H5T_MIPS_B64, H5T_MIPS_F64,
          H5T_MIPS_I64, H5T_NATIVE_B64, H5T_NATIVE_INT64, H5T_NATIVE_INT_FAST64,
          H5T_NATIVE_INT_LEAST64, H5T_STD_B64BE, H5T_STD_B64LE, H5T_STD_I64BE,
          H5T_STD_I64LE, H5T_UNIX_D64BE, H5T_UNIX_D64LE, H5T_ALPHA_B64,
          H5T_ALPHA_F64, H5T_ALPHA_I64 },

The types ending _F64 should be associated with GDL_DOUBLE instead e.g change code from
          /* GDL_DOUBLE */
       { H5T_NATIVE_DOUBLE },
to
          /* GDL_DOUBLE */
        { H5T_NATIVE_DOUBLE,H5T_IEEE_F64BE,H5T_IEEE_F64LE,H5T_INTEL_F64,H5T_MIPS_F64,H5T_ALPHA_F64},
(and remove the F64 types from the lines defining integer types below)

Similarly types ending F32 should be associated with GDL_FLOAT.

The effect of this bug is for double precision data in an HDF5 to be wrongly returned as an integer data type.
(I guess this only happens when the file being read uses a type other than H5T_NATIVE_FLOAT/DOUBLE)
(I have an e.g. hdf5 file which can be correctly read only after fixing hdf5_fun.cpp as suggested above)

GillesDuvert:
Hello, thanks for the report! The last HDF5-related changes were done by @ogressel , may be he can help? If you can clone gdl on GH and create a pull request all by yourself we will be more than happy to use your contribution to improve GDL.

rsiddans:
OK thanks for quick reply.
I've not used git much and not at all github...
I got the source (to a linux machine) via command line by doing
"git clone --recursive https://github.com/gnudatalanguage/gdl.git" 
[...]

Issue #1922, opened 2024-11-27
Missing NetCDF4 data types


rsiddans:
GDL does not allow read/write o several basic datatypes allowed by netcdf4. E.g. NC_STRING, and unsigned integer types in general (NC_UINT, NC_ULONG, NC_ULONG64). 

Furthermore, GDL reports EC_RANGE error if signed bytes are attempted to be read (whereas IDL will read these into its byte type, leaving the user to interpret the values).

These issues prevent use of GDL for several applications we have which depend on netcdf4 i/o.

I have been able modify my local copy of ncdf_att_cl.cpp and ncdf_var_cl.cpp to allow these to be read/written (although only scalar values/atts in the case of NC_STRING).

This is the first time I have attempted to modify code in an open source project like GDL, so am not sure what would be the best way for me to help get these changes into the repository going forward. Would be happy to share my modified code, but have very limited time for the kind of testing I imagine might be required before it would be appropriate to include the modifications in the repo.

GillesDuvert:
Submitting a PR ( or telling where to find the modified code ) will certainly speed up things --- as nobody else around is apparently "maintaining" the NetCDF-related code. I would suppose the ultimate "testing" would be to get GDL run the, I quote,   " several applications we have which depend on netcdf4 i/o."

rsiddans:
Pull request #1926 has my modified code (also covers issue #1923)

GillesDuvert:
merged and closing

Issue #1744, opened 2024-02-06
H5D_read has problems with variable-length strings


klimpel:
The handling of variable-length strings is not yet implemented in
hdf5_unified_read
For the case that the dataset has just a single element, I implemented it in my local version:
```
     } else if (ourType == GDL_STRING) {
 
-      if (debug) printf("fixed-length string dataset\n");
+      bool isVarLenStr = H5Tis_variable_str(elem_dtype) > 0;
+      if (debug) printf(isVarLenStr ? "variable-length string dataset\n" : "fixed-length string dataset\n");
 
       // string length (terminator included)
       SizeT str_len = H5Tget_size(elem_dtype);
 
       // total number of array elements
       SizeT num_elems=1;
       for(int i=0; i<rank_s; i++) num_elems *= count_s[i];
 
+      if (num_elems == 1 && isVarLenStr) {
+        char* raw = nullptr;
+        hdf5_basic_read( loc_id, datatype, ms_id, fs_id, &raw, e );
+
+        // create GDL variable
+        res = new DStringGDL(raw);
+
+        H5Dvlen_reclaim (ms_id, fs_id, H5P_DEFAULT, &raw);
+
+        return res;
+      }
       // allocate & read raw buffer
       char* raw = (char*) malloc(num_elems*str_len*sizeof(char));
```
Implementation of the array case is probably not difficult either, but I am neither familiar enough with DStringGDL, nor do I have a suitable .hdf5 file ready with which I could test that scenario.

GillesDuvert:
I took the liberty to assign this to @ogressel who has largely improved this code recently.

ogressel:
Thanks, @klimpel . I will have a look, if I find some time. But it's been long enough that I need to re-familiarize myself with the Array-specific code.

PR #1926, opened 2024-11-28
Fixes to netCDF4 and hdf5

rsiddans:
Fixes to add missing NetCDF4 datatypes and fix bug reading some float types in HDF5


rsiddans:
Related to issue #1922 and #1923

rsiddans:
Also note I included the code change suggested in #1744 to hdf_fun.cpp, so that it can read (a single) variable length string.

There will remain limitations in both hdf and netcdf for reading arrays of strings (as opposed to char) - however I think string arrays are unusual to find in netcdf files. (Scalar strings attributes are very common though and these can now be read with these changes.)

rsiddans:
Actually a couple of other changes are included in this code:
- keywords gzip+shuffle are added to ncdf_vardef (allowing internal compression to be enabled)
- there is a fix in map_proj_init.pro to remove a limit of 80 degrees, which should not apply to the default map projection (cylindrical) - In IDL by default map_set gives a map covering the full globe to +/-90 deg. The 80 degree limit is relevant for the mercator projection.

GillesDuvert:
@rsiddans please remove plplot_config.h and plDevs.h from your PR. They should not be here, and I've no obvious mean to remove them .

GillesDuvert:
and thanks!

GillesDuvert:
@rsiddans sorry, I found how to delete the 'offending files' from here.

rsiddans:
> @rsiddans sorry, I found how to delete the 'offending files' from here.

OK thanks - sorry not sure why those files were there & many thanks again for all the help.

GillesDuvert:
@rsiddans additionnal thanks to have spotted the map_set problem.

PR #1945, opened 2024-12-13
Close 1353 and 1928 make /NAN implied for several functions

GillesDuvert:
note alleged !version.release is now 8.5.1, the first one where /NAN is not needed anymore in calls. (or, to say differently, it is not necessary to add /NAN to a command where not-a-namber should be taken into account)


brandy125:
Nice!

GillesDuvert:
not yet merged, troubles with the GH continuous integration system.

PR #1946, opened 2024-12-14
Use ubuntu 22 for the moment in workflows

GillesDuvert:


GillesDuvert:
OK looks good even if there seems to be errors, they are related to the nex use of /NAN, which should not be there (git branch hierarchy error).
Closing and redoing better.

Issue #1881, opened 2024-09-12
OSX compiler (M1, M2?) and unsigned values (explains why test_format do not pass)


GillesDuvert:
Basicall the reason why test_format do not pass is not a format problem.
It is just that 
linux:
```
GDL> r=-1d
GDL> z=fix(r,type=15)
GDL> z
  18446744073709551615
GDL> z[0]=r
GDL> z
  18446744073709551615
```
OSX:
```
GDL> r=-1d    
GDL> z=fix(r,type=15)
GDL> z
                     0
GDL> z[0]=r
GDL> z
                     0
```

GillesDuvert:
related to #1877 ?

GillesDuvert:
probably beacuse https://developer.apple.com/documentation/apple-silicon/addressing-architectural-differences-in-your-macos-code#Audit-Code-that-Contains-Float-to-Int-Conversions

Issue #1877, opened 2024-09-02
Linker warning


jtappin:
I'm not sure if this is something we should be worried about or not, but when linking the current GIT version (and for a few weeks now) on Manjaro (Stable branch) I get a lot of warnings similar to this:
```
In member function ‘InitSize’,
    inlined from ‘__ct_base .constprop’ at /data/software/AUR/gnudatalanguage-git/src/gnudatalanguage-git/src/envt.cpp:321:23:
/data/software/AUR/gnudatalanguage-git/src/gnudatalanguage-git/src/envt.hpp:408:12: warning: argument 1 value ‘18446744073709551615’ exceeds maximum object size 9223372036854775807 [-Walloc-size-larger-than=]
  408 |     eArr = new T[ s]; // constructor called
      |            ^
/usr/include/c++/14.2.1/new: In member function ‘__ct_base .constprop’:
/usr/include/c++/14.2.1/new:133:26: note: in a call to allocation function ‘operator new []’ declared here
  133 | _GLIBCXX_NODISCARD void* operator new[](std::size_t) _GLIBCXX_THROW (std::bad_alloc)
      |         
```
while I haven't seen any adverse consequences, I thought it might be worth raising a flag.

Issue #1874, opened 2024-08-15
string and strtrim wrong on ULONG


GillesDuvert:
```
GDL> a=randomu(seed,/ulong)
GDL> string(a)
 -2007711753
GDL> a
  2287255543
```
!!!!!!!!!!!!!!

Issue #1842, opened 2024-06-06
Possible duplicate: test_byte_conversion.pro and test_bytscl.pro fail on arm64 


GillesDuvert:
tests may be currently disabled on these machines (#1788) but the bugs persist.

PR #1949, opened 2024-12-15
Fix GDL Convert() internal function for compatibility with ARM64

GillesDuvert:
PRINT, PRINTF, STRING, STRTRIM, all formats ... depend on C/C++ casting to different types.
This PR modifies the Convert() function to get similar results on X86_64 and AARCH64 (ARM64, M1,M2...) when needed (conversion of a float/double towards unsigned integer types)
This to solve 99% of #1881 , #1877 and #1842 but see #1948
  
Besides, snprintf format was 'i' and not 'u' in Ulong and Uint --- this made for error #1874

Addditionnally it was found that multithreading in Convert() may have been wrongly activated during some conversions, the code now checks GDL_NTHREADS value everywhere (?). Could potentially speedup things.


GillesDuvert:
As a result, test_byte_conversion, test_bytscl and test_formats were reenabled for ARM64 machines. 
Strangely, these test continue to fail on macos-15 on GH, but they pass OK on a arm64+macos-15 machine I have access. So I deem the PR to be OK.

GillesDuvert:
One thing that will NOT be the same is related to Infinity and NaN conversion to integer, that cannot be solved easily.

PR #1958, opened 2024-12-22
Enamble autoprint for structures

GillesDuvert:
The ANTLR2 interpreter has no means to differentiate
if 'a.proc' is a call to the procedure 'proc' of an object 'a' , or the tag 'proc' of a structure 'a'.
It would always interpret as the 1st option.
This change permits to test if (in restricted cases: VAR and SYSVAR only at the moment) a is indeed a structure, not an OBJ, and thus if "a.proc" is an autoprint command, not a call to procedure 'proc' of object 'a'.

### DIFF
--- a/src/GDLLexer.cpp
+++ b/src/GDLLexer.cpp
@@ -15,6 +15,34 @@
 #include <errno.h>
 
 #include <cstdlib>
+#include <fstream>
+#include <iostream>
+#include <string>
+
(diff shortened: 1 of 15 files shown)


GillesDuvert:
Things are complicated by the fact that variables are not defined at compilation time, so this trick should be worked on. Basically one need to know if we are in interactive of compiling mode.
Removed for the time being.

PR #1996, opened 2025-02-16
Always build plplot as a static library

opoplawski:
I'm working on updating the Fedora gdl package to 1.1.X.  I'm using the bundled plplot as it appears to be nearly impossible not to do so.  The issue I'm facing now is that it's installing libplplot.so - which is problematic as it would conflict with the separate plplot package.  Near as I can tell this happens because when building the python module plplot is built as a shared library.  Also, it seems that there is no particular reason for this.  It appears to build fine with this patch as a static library.  Was there some other reason it was done as a shared library?


GillesDuvert:
Dear orion, clearly no reason to build it as a .so instead of a .a
the only reason for this situation is that we (and I in particular) are not cmake specialists and just stop when the desired result is attained. Here it was by the minimal cut (a lot) and paste (as minimally as possible) from the plplot CMakeLists.txt files to make the bundled plplot excerpts work. I would suffer a lot to find how, for example, have cmake create a libplplot.a instead of a libplplot.so
We'll be more than happy to patch our plplot CMakeLists.txt following your advice.

GillesDuvert:
I'll ignore the QHULL problem under windows as it it yet another silly annoyance.

PR #1999, opened 2025-02-18
Implement read variable-length strings from HDF5 file

ogressel:
this addresses issue #1744 raised by @klimpel


GillesDuvert:
@ogressel @klimpel  I'm going to merge this (the build problems are unrelated to theses changes), is that OK?

ogressel:
> @ogressel @klimpel I'm going to merge this (the build problems are unrelated to theses changes), is that OK?

Fine by me, but it would be great if it could be tested independently as well. The unit test covers a 2D array of vlen strings which should be general enough, though.

ogressel:
I did not mean to stall this, please go ahead with the merge.

PR #2027, opened 2025-04-23
Speedup lexer

### LINKED ISSUE #2012: compilation time taks 10 times longer with version 1.1.2_testing2
The compilation time of some routines take 10 times longer now:

Older version:

```
GDL> !gdl
{
    "RELEASE": "1.1-2-g985d888e-dirty",
    "BUILD_DATE": "Nov 15 2024",
    "EPOCH": 1731639600,
    "GDL_USE_DSFMT": 1,
    "GDL_USE_WX": 1,
    "GDL_POSIX": 1
}
GDL> tic
GDL> .run gcp2phi_abs2
GDL> toc
% Time elapsed: 0.98922586 seconds.
GDL> tic
GDL> .run gcp2phi_abs2
GDL> toc
% Time elapsed: 0.99576092 seconds.
```

Newest version:

```
GDL> !gdl
{
    "RELEASE": "1.1.2_testing2",
    "BUILD_DATE": "Mar 31 2025",
    "EPOCH": 1743390000,
    "GDL_USE_DSFMT": 1,
    "GDL_USE_WX": 1,
    "GDL_POSIX": 1
}
GDL> tic
GDL> .run gcp2phi_abs2
GDL> toc
% Time elapsed: 9.6382549 seconds.
GDL> tic
GDL> .run gcp2phi_abs2
GDL> toc
% Time elapsed: 9.5839789 seconds.
GDL> 
```

GillesDuvert:
Here I propose changes in the ANTLR code that aims at speed up (quite a lot in some cases) the compilation of procedures.
The Parser always start parsing a procedure in STRICTARR mode, the normal, modern, way to write IDL code, using '[' and ']' instead of '(' and ')' to mark array indices. The parser is much faster at parsing the procedure if this rule is respected.
If the parser encounters an problem, at locations where an array index is acceptable, it reparses the whole procedure using a 'relaxed' mode where z=toto(a,b,c) can be in fact z=toto[a,b,c] and not the function toto called with a,b,c as arguments.
parsing speed gain is 3 to 10 (still waaay below IDL's homegrown parser :smile: )

Additionnaly this removes the need of the `--fussy` and `--sloppy` commandline swiches, that are removed.
I've added the `--trace-old-syntax` switch that can permit to find ALL (??)  occurences of the pre-v5 (i.e;, non-strictarr) way of writing IDL code. An example:
```
$ gdl --trace-old-syntax -e ".compile /home/gildas/IDLAstro/coyote/cgpickcolorname.pro"
old syntax at line 523, column 26
% Compiled module: CGPICKCOLORNAME_SELECT_COLOR.
```
closes #2012 
 thanks to @brandy125 for mentioning the speed problem.

Note that all this cannot solve the 'nested parenthesis' nigthmare. See the (added) antlr4 diectory to an example of code that talkes forever with ANTLR2 and is immediate with ANTLR4.

### DIFF
--- a/scripts/build_gdl.sh
+++ b/scripts/build_gdl.sh
@@ -118,12 +118,16 @@ elif [ ${BUILD_OS} == "Linux" ]; then
     ) # JP 2021 Mar 21: SuSE lacks eccodes
 elif [ ${BUILD_OS} == "macOS" ]; then
     BREW_PACKAGES=(
-        llvm@18 libx11 libomp ncurses readline zlib libpng gsl wxwidgets graphicsmagick libtiff libgeotiff netcdf hdf5 fftw proj open-mpi python numpy udunits eigen
-        eccodes glpk shapelib expat gcc@11 qhull dylibbundler cmake
+        libx11 libomp ncurses readline zlib libpng gsl wxwidgets graphicsmagick libtiff libgeotiff netcdf hdf5 fftw proj open-mpi python numpy udunits eigen
+        eccodes glpk shapelib expat gcc qhull dylibbundler cmake
     ) # JP 2021 Mar 21: HDF4 isn't available - not so critical I guess
       # JP 2021 May 25: Added GCC 10 which includes libgfortran, which the numpy tap relies on.
       # J-KL 2022 July 30: GCC 10 didn't work with apple silicon mac. So I replaced it with GCC 11
       # GD added dylibbundler that simplify building correct apps.
+	  # GD 25/04/2025 do not use this gcc for compilation. Do as mantioned by homebrew for libomp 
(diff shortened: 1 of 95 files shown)


alaingdl:
I noticed that this change may  help us https://cliutils.gitlab.io/modern-cmake/chapters/packages/OpenMP.html
I am OK to move to Cmake 3.12 !

Questions :

* How can I help you ? Since it seems to be OK on recent OSX 15.3.2 
* How can I be *sure*  that OpenMP is ON ? (if OPENMP_EXISTS() flag reliable ?)
* Do we have tests illustrating the change in speed with OpenMP ON ?

Issue #2017, opened 2025-04-11
regression triggered by READCOL  (+ double call to .reset crash)


alaingdl:
This bug appeared between v1.1.2_testing-2-g2daaac2c (OK 250224) and v1.1.2_testing-7-ge2f79aac (KO 250319)
and was reproduced on U22.04.5 LTS & Debian 10). Sorry not being able to reduce more this annoying bug.

How to reproduce ? having idlastro + coyote in the path, having a file to be read by READCOL, stop somewhere inside a procedure

```
.r ~/GDL/last-gdl/src/pro/utilities/path_add.pro
path_add, /rec, '~/GDL/last-gdl/src/pro/'
path_add, /rec, '~/IDLetGDL/Librairies/DebianAstroGIT/   ; subdirs with idlastro + coyote ...
```

create a procedure with a STOP inside
calling READCOL inside

Seems to recover after .reset *but* second call to .reset crashed ! (core dumped)

```
GDL> readcol, 'toto.dat',g,s
% READCOL: Format keyword not supplied - All columns assumed floating point
% READCOL: Skipping Line (check_numeric) 1
% READCOL: Skipping Line (check_numeric) 2
% READCOL: Skipping Line (check_numeric) 3
% READCOL: 10 valid lines read
GDL> test
% Compiled module: TEST.
% TEST: Attempt to call undefined procedure: AZAZAZAZ
% Execution halted at: TEST                 3 /home/alain/GDL/Bug_READCOL/test.pro
%                      $MAIN$          
GDL> readcol, 'toto.dat',g,s
% READCOL: Format keyword not supplied - All columns assumed floating point
% Compiled module: CGERRORMSG.

%READCOL: Type conversion error: Unable to convert given STRING to DOUBLE.

Traceback Report from READCOL:

     Type conversion error: Unable to convert given STRING to DOUBLE.


```

Aother way is to suceed to have the message : 
```
GDL> azertyui
Unhandled GDL exception: TEST: Attempt to call undefined procedure: AZERTYUI
```
after that READCOL won't worked :(

alaingdl:
Since `void = cgErrorMsg(/quiet)` is called inside READCOL, it might be a side effect of the way !Error_State is managed in GDL code, but I was not able to follow that in details. Suspects in the c++ code (changed between the two versions) : 
-- dinterpreter.cpp /  dinterpreter.hpp
-- GDLInterpreter.cpp / GDLInterpreter.hpp
-- gdlc.i.g

[...]

Issue #2023, opened 2025-04-17
better message when allocating too large byte ?


alaingdl:
Would it be possible to improve the message when trying to affect/create a byte variable with bad argument (typicially number too large) ? Clearly not urgent but would be smart !

```
GDL> a=1516b
     ^
% Syntax error.

IDL> a=1516b

a=1516
   ^
% Byte constant must be less than 256.
```

PR #2032, opened 2025-04-27
Solve reads from cr or crlf files issue 2026

### LINKED ISSUE #2026: regression(s ?) related to READF, one related to CR char
As detected by `test_readf_with_crlf.pro` and `spawn_is_dangerous/test_readf.pro` we do have regressions in READF (both are OK with IDL 8.8)

```
GDL> TEST_READF
% TESTREADF: % TEST_READF: failed with CR char
% TESTREADF: % TEST_READF: failed with CR char
% TEST_BUG_573: running TEST_BUG_573
% TEST_BUG_573: passing with success TEST_BUG_573
```

Difficult to accurately trace back this problem but code was ok in gdl 1.0.2 (OK on 1 Jan 2023, not OK since  Feb. 10, 2023)

```
GDL> TEST_READF_WITH_CRLF
% TEST_READF_WITH_CRLF: End of file encountered. Internal error: Stream not found.
% Execution halted at: TEST_READF_WITH_CRLF    16 /home/alain/GDL/gdl-1.0.6git250114CMake/testsuite/test_readf_with_crlf.pro
```

This regression did appeared between v1.1-19-g27b7038c (6 Dec 2024) and v1.1-35-gaa9693b8 (14 Jan 2025)

GillesDuvert:
closes #2026 
@alaingdl can you check that it solves #2017 (READCOL)


alaingdl:
Unfortunately, the problem with READCOL is not solved. (the two ohter tests are OK now)

Then I am currently trying to go deeper in READCOL, commenting out the lines which catch error.
I am stopped at line 332
```
 tst = strnumber(var[i],val,hex=hex[i],NAN=nan)
```

Side effect with the large change in the Lexer ? When we ask for a unknown variable at main level 
we have (as expected) : 
```
GDL> print, b
% PRINT: Variable is undefined: B
% Execution halted at: $MAIN$          
```
but inside a procedure : 
```
GDL> print, aa
Unhandled GDL exception: PRINT: Variable is undefined: AA
```
instead of something like :
```
IDL> print, toto
% PRINT: Variable is undefined: TOTO.
% Execution halted at: TEST                3 /home/alain/GDL/Bug_READCOL/test.pro
%                      $MAIN$          
```

alaingdl:
Since the crash in CI is 100% OSX/clang related, then not related to the proposed solution here, which solved #2026 I think it is OK to valid this PR but worked again on OSX specific problems for compilation. Is it OK ?

(yes #2026 is solved)
(no #2017 is not solved but it is another story)
OSX not compiled here

GillesDuvert:
OK Alain, I merge. and #2026 is closed.
I'll try to reproduce the #2017 on my side.
The changes in the interpreter may have indeed some side-effects at the level of error messages since I also tried to solve #2023 with a rather simple solution.

PR #2043, opened 2025-06-11
WIP adding the DLM (and LINKIMAGE) functionality.

GillesDuvert:
With this PR, the user has access to a LINKIMAGE and all DLM-related functionalities with DLL (.so, .dll) files compiled for IDL (no need to recompile) or (and this would be way better in terms of efficiency) directly using GDL internal code.

DLMs found in !DLM_PATH will be charged automatically by GDL, whatever their 'flavor' if the `.dlm` is correctly written. 

HELP,/DLM works. Also DLM_REGISTER.

###  IDL-compatible libraries
user: nothing to do. just use IDL_DLM_PATH to  point to the dlm files or use DLM_REGISTER. Example: fmedian in the NASA ssw suite (**however see footnote**, there is no gain in using fmedian)
programmer: Just follow IDL documentation.  The distribution comprizes the idl_export.h needed. 

###  Libraries directly compatible with GDL (no lengthy changes from C++ GDL variables to IDL C variables and back)
These have to be compiled with a C++ compiler, and must `#include "envt.hpp"`. Each function/procedure to be integrated withing GDL should just be defined as:
```
extern "C" BaseGDL* my_fun( EnvT* e) //function: gives back a gdl variable
extern "C" void my_pro( EnvT* e) //defines a procedure
```
my_pro and my_fun probably need to be **uppercase** 

**HOWEVER:** if the code contains its _own C++ templates_  then there are chances that some template specialization will not be present in the .so (.dll) and the library cannot be linked with GDL. TBC.

To illustrate, POWELL is now a DLM, in `gnudatalanguage/dlm `

### usage
To differentiate the two 'flavours', GDL-native .so (.dll) code must be loaded wih a minimal change wrt. the IDL fashion:

1. for DLMs: the DLM shoud **start** with a line saying `#%GDL_DLM`
2. for LINKIMAGE: add the /NATIVE keyword when calling LINKIMAGE.

### footnote
1. ssw'w fmedian.dlm does not seem faster than IDL's own on my machine (but: fmedian permits non-square "kernels"):
```
IDL> z=dist(512) & tic & a=fmedian(z,17,17) & toc
% Time elapsed: 0.60005403 seconds.
IDL> tic & a=median(z,17) & toc    
% Time elapsed: 0.31774592 seconds.
```
besides, fmedian and median do not produce the same result!!!

2. but both are way slower than GDL's MEDIAN() that should be preferred:
```
GDL> z=dist(512) & tic & b=median(z,17) & toc
% Time elapsed : 0.079225063 seconds.
```

 And yes, GDL's MEDIAN handles non-square kernels provided the x and y widths are odd:
```
GDL> z=dist(512) & tic & b=median(z,[11,17]) & toc
% Time elapsed : 0.050585985 seconds.
```
But use with caution as this is not IDL-compatible.


alaingdl:
I can compile with (apparent) success on Debian 10
`% Loaded DLM: POWELL`

But we cannot compile on OSX (large update of Xcode & LLVM last days). We will try hard !

GillesDuvert:
Well, thanks, that's good news for a start. I'm slowly polishing (!) the thing. (IDL<->GDL Structures are tough.)
Not soon in production, I'm afraid...

liufrancois:
Hi, I'm trying to build the PR on a Mac Mini M2 running macOS 15.5, but I ran into a few issues:

when compiling with Clang(20.1.5):
`make -j` stops at around 30% with no error message.

In `src/export.hpp`, I got an error: `'ulong' was not declared`. I fixed it by adding `typedef unsigned long ulong;`.

In `src/dlm/CMakeLists.txt`, I had to change `brent.cpp` to `brent.hpp` to remove an error.

In `src/basic_pro_jmg.cpp`, I had to remove `RTLD_DEEPBIND` to avoid a compilation error.

I also encountered an error with an `unknown option: --dynamic-list=.../src/dynlist.txt`.


I also tried compiling with GCC(11.5.0), and to reach the same point successfully, I had to disable several components by adding these flags:
`-DX11=off -DWXWIDGETS=off -DPYTHON=off -DREADLINE=off`

Hope this helps with improving compatibility on macOS. Let me know if you'd like me to test anything else.

GillesDuvert:
@liufrancois thanks, I'm aware of issues on OSX and Windows. Quite normal in fact. Will attend these issues once the linux version is OK, which it is not. At that moment the "WIP" in the PR title will disappear. Thanks again.

GillesDuvert:
@liufrancois if the goal is to get a gdl-only dlm to work, aside the problems you solved,  you need to find how to get all symbols in the gdl executable available to the "loaded" .so file. This is  the purpose of the link option that osx/clang does not seem to accept.

PR #2052, opened 2025-07-18
microscopic updates 

alaingdl:
- put 3.13 instead of 3.16 in CmakeList.txt because it is enough and I have several machines running cmake 3.13
- update FindReadline.cmake. This is important to properly manage path to any Readline version (eg locally compile) since Readline 8.3 is problematic with GDL on OSX
- add AppleClang in src/dlm/CmakeList.txt to have DLM compilable on OSX 15.5
- update "test_hdf5" to exit when utilities are not around, but with a /force keyword to run it is needed
- update LIST in testsuite
- update "test_file_lines" to work on IDL too
- add new cases (STRTRIM) in "test_str_functions"
 
 A comment on the tests in LIST : clearly "test_hdf5.pro" triggers a (unclear) problem on current OSX with file containing a "-". Please don't remove tests in LIST, they may help to find other problems !


GillesDuvert:
@alaingdl strange, this PR has no files changed, is it normal?

GillesDuvert:
OK, I commit, (no file changed: no harm) , as this make things clear.

PR #2072, opened 2025-09-10
Update python_print.py

iancrossfield:
update 'print' syntax to Python2 *or* Python3-compatible.


GillesDuvert:
workflow cancelled (as python is not tested)

PR #2077, opened 2025-09-15
Min max for plot is back

GillesDuvert:


GillesDuvert:
Again, this impossibility to get rid of already merged commits!!!!

PR #2082, opened 2025-09-17
More NetCDF functionality

rsiddans:
Mods add some more NetCDF functionality including read string arrays and unsigned int64
Also more datatypes can be written
When writing, overflow of datatype is now allowed (e.g. when writing int value=300 to a byte NetCDF variable the conversion will overflow but this is allowed) - this is to reproduce the behaviour of IDL. 
Also fixes a remaining bug in HDF read of unsigned long64 (partially fixed before)


GillesDuvert:
Many thanks, @rsiddans !

rsiddans:
You're welcome - thanks for merging (+ dealing with the other little issues I raised so quickly)

Issue #2084, opened 2025-09-19
Fail to install gdl on macOS sequoia (arm64)


jianxing-chen:
The computer infomation:
Apple M4max 
Darwin Kernel Version 24.6.0 
arm64

I use the command:
$ sudo port install gnudatalanguage
to install gnudatalangue,
but there are several errors occured as follows:

.......
--->  Installing intltool @0.51.0_7
--->  Activating intltool @0.51.0_7
--->  Cleaning intltool
--->  Fetching archive for pslib
--->  Attempting to fetch pslib-0.4.6_2.darwin_24.arm64.tbz2 from https://pek.cn.packages.macports.org/macports/packages/pslib
--->  Attempting to fetch pslib-0.4.6_2.darwin_24.arm64.tbz2 from https://mirror.sjtu.edu.cn/macports/packages/pslib
--->  Attempting to fetch pslib-0.4.6_2.darwin_24.arm64.tbz2 from https://kmq.jp.packages.macports.org/pslib
--->  Configuring pslib
Error: Failed to configure pslib: consult /opt/local/var/macports/build/pslib-a00eb35f/work/pslib-0.4.6/config.log
Error: Failed to configure pslib: configure failure: command execution failed
Error: See /opt/local/var/macports/logs/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_print_pslib/pslib/main.log for details.
Error: Follow https://guide.macports.org/#project.tickets if you believe there is a bug.
Error: Processing of port gnudatalanguage failed
--->  Some of the ports you installed have notes:
  cmake has the following notes:
    The CMake GUI and Docs are now provided as subports 'cmake-gui' and 'cmake-docs', respectively.
  coreutils has the following notes:
    The tools provided by GNU coreutils are prefixed with the character 'g' by default to distinguish them from the
    BSD commands.
    For example, cp becomes gcp and ls becomes gls.
    
    If you want to use the GNU tools by default, add this directory to the front of your PATH environment variable:
        /opt/local/libexec/gnubin/
  db48 has the following notes:
    The Java and Tcl bindings are now provided by the db48-java and
    db48-tcl subports.
  eigen3 has the following notes:
    This product includes software developed by the University of Chicago, as Operator of Argonne National Laboratory.
  hdf5 has the following notes:
    Mac users may need to set the environment variable "HDF5_USE_FILE_LOCKING" to the five-character string "FALSE"
[...]

PR #2086, opened 2025-09-23
Switch to Eigen/CMake's FindEigen3.

jkohnert:
Attempt to switch to Eigen3's CMake-implementation. This also supports Eigen3's current release version 5.0.0 as long as the used API still works.


GillesDuvert:
@jkohnert thanks, tweaking the CMake code is never easy... The failure of rebuild on macos-15 is unrelated, of course (see #2084)

jkohnert:
@GillesDuvert thanks. I'm still strggling with the Windows build error, tough. I currently don't have a Windows maschine at hand, so it is a bit hard to debug that stuff...

jkohnert:
The Windows test error doesn't really make sense to me:
```log
136/215 Test #136: test_memory.pro ....................***Timeout 3600.02 sec
% Attempt to call undefined procedure: DLM_REGISTER
% Execution halted at: $MAIN$          
% Compiled module: TEST_MEMORY.
```
I had this error in an earlier stage and it went away when re-adding the CXX_FLAG. This optimization flag seems unrelated to the error. My next change to exclude Eigen3 v5.0.0 re-introcuded the error alhough only the new version of Eigen3 now wil report an error. Why is this `DLM_REGISTER` undefined? Hmmm.

GillesDuvert:
@jkohnert DLM_REGISTER is called when GDL starts, indeed nothing directly related.
Eventually, put this on hold, somebody (me?) will test on a windows (virtual) machine at some point.

GillesDuvert:
@jkohnert I see you have changed CMAKE_CXX_STANDARD from 11 to 14 , obviously for the best. (and Eigen:: probably makes the best of C++14)
However historically there has been quite a strong debate about changing the c++ version, with all those (still alive?) old machines running gdl and such. Even the CMake version is stuck to an absurdly old number for these same reasons.
So I'm delaying this PR until I've got an agreement from the rest of the team, stay tuned.

alaingdl:
* about Eigen3 & CMake : I was quite surprise you need a change, the old (current) version is working fine for me

* about Cmake, old & new  (preparing the move to cmake 4)

I am maintaining GDL on some laptops and HPC computers with "old" systems (eg Debian 10, Ubuntu 20.04 LTS ) where I have only old cmake versions (eg 3.13.4 for Debian 10) and it is now working fine for GDL. Please don't broke that !
I know  I can compile newer CMake & so on, but in limited time it is just not a priority for me.
(eg : making careful revisions of some PR is more critical, or dissect some bug)

* about change of c++ version

I am not an expert but I suspect we may have side effects, and also maybe some changes to be made in the code. When I compile GDL on OSX I always received a long list of warnings ... Who has time & competence to clarify that ?

Question : is it difficult to have 2 cmake files, not changing the actual one which is very stable for me, and a new one preparing the move directly into Cmake V 4 ? (new syntax, I assume simpler !)


My feeling it would be great to have a one or two days face to face meeting fully dedicated to exchange and fix code and discuss/prepare a roadmap, we made it ten years ago in Meudon (Paris Observatory) and it was a great time !

jkohnert:
@GillesDuvert and @alaingdl I'm using Arch, and they switched to Eigen3 v5.0.0 recently. This change breaks the CMake-script currently shipped with GDL. So I made the attempt to fix it. And since Eigen ships CMake-Configurations itself, the first attempt was to get rid of this own implementation and just use what upstream provides so we don't reinvent the wheel. This changed configuration than still works with Eigen3 v5.0.0. 

But: Eigen3 v5.0.0 breaks compilation when using C++11, since it depends on C++14 (see: https://gitlab.com/libeigen/eigen/-/releases/5.0.0#breaking-changes). At first I tried to exclude Eigen3 v5.0.0 but than I tried to just adjust the C++ standard used and see if it works. Which it does. And since C++14 is more like an "augmentation" to C++11 and AFAIK doesn't break things that are working in C++11, that change should probably be fine. I saw a bunch of additional warnings, though, however: moving forward in the standard seems to be desirable, anyway. :smile:

jkohnert:
@alaingdl Debian 10 reached end-of-life more than a year ago, even for LTS suport, see: https://endoflife.date/debian I don't think it's desiable to keep the code it is just to support operation system versions that are declared unsupported upstream.

GillesDuvert:
I'm very much in favor of keeping the head (!) above the water, so using new standards when they are needed. 
I believe most of the changes can be retrofitted to those old and obsolete systems with some tweaking and `#define` everywhere but I would prefer these backports to be developed in another repository. And why EOL systems would need to be upgraded to the last GDL version appear questionable.
That said, @jkohnert I'm not sure that your PR has been compiled (on ubuntu 22.04 @ github) in C++14, due to the absence of any warning in the compilation? If its is, that's sort of miraculous.

jkohnert:
@GillesDuvert I compiled the code including this patch using my current Arch system. Compilation breaks due to Eigen3 v5.0.0 not supporting C++11 when using C++11 in CMakeLists.txt; compilation works fine when setting C++14 in the CMakeLists.txt. Also all tests run fine when using C++14 in CMakeLists.txt. I have to admint, though, that the actual standard used is currently invisible in the pipline's build logs. So I can't tell for sure whether C++14 was actually used in the Ununtu-pipeline-build. And, of course, I know "it's working on my maschine" isn't a valid argument for merging this PR.

AFAIR it is possible to get more build output to be able to check what compile command are used exacly, so I can check that.

It is worth noting that all changes work using currently defined CMake-versions, vo v4.X is needed. Even Eigen3's CMake implemenation only requires CMake v3.0.

We could also use the existing own FindEigen3-implementation for v1.1.X and simply return a better error message for Eigen3 v5.0.0, which obviously isn't "too old". :smile: This would probably require a different PR.

Or we could probably split this PR for upcoming v1.1 versions: use Eigen3's FindEigen3 implementaion and only exclude v5.0.0, so we can stick on C++11 for v1.1. That would only need to revert the latest commit to this PR and make another PR with that last commit.

Best, Jan

jkohnert:
Setting verbosity for makefile in cmake confirms the code is build in C++14 standard when the setting is active in CMakeLists.txt, see https://github.com/gnudatalanguage/gdl/actions/runs/18036500131/job/51324556556?pr=2086#step:4:600

This again resembles the local build I get. Locally, I currently get two warnings:
```log
[101/239] /usr/bin/c++ -DH5_BUILT_AS_DYNAMIC_LIB -DHAVE_CONFIG_H -DWXUSINGDLL -D_FILE_OFFSET_BITS=64 -D_GNU_SOURCE -D_LARGEFILE_SOURCE -D_POSIX_C_SOURCE=200809L -D__WXGTK3__ -D__WXGTK__ -I/home/jankoh/projects/gdl/src -I/usr/include/tirpc -I/usr/lib/wx/include/gtk3-unicode-3.2 -I/usr/include/wx-3.2 -I/usr/include/GraphicsMagick -I/opt/hdf4/include -I/usr/include/python3.13 -I/usr/lib/python3.13/site-packages/numpy/_core/include -I/usr/include/freetype2 -I/home/jankoh/projects/gdl -I/home/jankoh/projects/gdl/src/antlr -I/home/jankoh/projects/gdl/cmake-build-debug/src -I/home/jankoh/projects/gdl/src/whereami/src -I/home/jankoh/projects/gdl/cmake-build-debug -isystem /usr/include/eigen3 -fopenmp -g -std=gnu++14 -fPIE -fdiagnostics-color=always -MD -MT src/CMakeFiles/gdl.dir/basic_op_sub.cpp.o -MF src/CMakeFiles/gdl.dir/basic_op_sub.cpp.o.d -o src/CMakeFiles/gdl.dir/basic_op_sub.cpp.o -c /home/jankoh/projects/gdl/src/basic_op_sub.cpp
In Datei, eingebunden von /usr/include/eigen3/Eigen/Core:349,
                 von /home/jankoh/projects/gdl/src/includefirst.hpp:76,
                 von /home/jankoh/projects/gdl/src/basegdl.hpp:21,
                 von /home/jankoh/projects/gdl/src/str.hpp:25,
                 von /home/jankoh/projects/gdl/src/GDLParser.hpp:16,
                 von /home/jankoh/projects/gdl/src/GDLLexer.hpp:14,
                 von /home/jankoh/projects/gdl/src/dinterpreter.hpp:42,
                 von /home/jankoh/projects/gdl/src/basic_op_sub.cpp:27:
/usr/include/eigen3/Eigen/src/Core/CwiseNullaryOp.h: In instantiation of »static const Eigen::DenseBase<Derived>::ConstantReturnType Eigen::DenseBase<Derived>::Constant(Eigen::Index, Eigen::Index, const Scalar&) [with Derived = Eigen::Array<std::complex<float>, -1, 1>; ConstantReturnType = Eigen::CwiseNullaryOp<Eigen::internal::scalar_constant_op<std::complex<float> >, Eigen::Array<std::complex<float>, -1, 1> >; Eigen::Index = long int; Scalar = std::complex<float>]«:
/usr/include/eigen3/Eigen/src/Core/ArrayBase.h:125:69:   required from »Derived& Eigen::ArrayBase<Derived>::operator-=(const Scalar&) [with Derived = Eigen::Map<Eigen::Array<std::complex<float>, -1, 1>, 16, Eigen::Stride<0, 0> >; Scalar = std::complex<float>]«
  125 |     internal::call_assignment(this->derived(), PlainObject::Constant(rows(), cols(), other),
      |                                                ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
/home/jankoh/projects/gdl/src/basic_op_sub.cpp:312:9:   required from »Data_<Sp2>* Data_<Sp>::SubS(BaseGDL*) [with Sp = SpDComplex]«
  312 |   mThis -= s;
      |   ~~~~~~^~~~
/home/jankoh/projects/gdl/src/instantiate_templates.hpp:33:16:   required from here
   33 | template class Data_< SpDComplex>;
      |                ^~~~~~~~~~~~~~~~~~
/usr/include/eigen3/Eigen/src/Core/CwiseNullaryOp.h:186:1: Anmerkung: das ABI der Strukturübergabe mit »complex float«-Element hat sich in GCC 4.4 geändert
  186 | DenseBase<Derived>::Constant(Index rows, Index cols, const Scalar& value) {
      | ^~~~~~~~~~~~~~~~~~
```

This is obviously unrelated to the standard change, and just a hint.

So IMHO using C++14 seems pretty straight forward. More problems are likely to occur when using C++17 or newer standard versions, since there are breaking changes w.r.t. C++11

I'll revert the change making the makefile verbose, now.

Best, Jan.

opoplawski:
eigen3 5.0.0 has landed in Fedora Rawhide so gdl builds have started to fail.  Looking forward to seeing this finished and released.  Thanks!

PR #2087, opened 2025-09-23
HDF5 can now read compound (structure) arrays

rsiddans:
HDF5 can now read compound (structure) arrays (before could only read scalar compound variables).
Also minor changes to handling of overflow of byte types when writing NetCDF.

### DIFF
--- a/src/hdf5_fun.cpp
+++ b/src/hdf5_fun.cpp
@@ -918,21 +918,43 @@ namespace lib {
 
        if (debug) printf("compound dataset\n");
 
-       if(data_rank>0)
-          e->Throw("Only scalar dataspaces supported for compound datasets.");
-
-       DStructDesc* cmp_desc = new DStructDesc("$truct");
-       DStructGDL* res = new DStructGDL(cmp_desc);
-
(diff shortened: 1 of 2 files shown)


GillesDuvert:
@rsiddans thanks, will merge soon, however, I wonder if testsuite/test_hdf5.pro could be patched to test these changes?

rsiddans:
> @rsiddans thanks, will merge soon, however, I wonder if testsuite/test_hdf5.pro could be patched to test these changes?

Good point - I just looked - it should be easy enough as the test file "hdf5-struct-test.h5" already has a compound array which cannot be read without this update. I think the test program does not report error on the file now because it does not attempt to read the compound array (pro test_hdf5_comp only reads "a_compund_dataset", not "a_compund_dataset_array") . 

However if I run test_hdf5 it crashes out before it gets to test_hdf5_comp, so I suppose this test program needs other fixes...
"% Execution halted at: TEST_HDF5_GROUP    614 /misc/users/rsiddans/gdl/20250917/rsmods/testsuite/test_hdf5.pro"

PR #2103, opened 2025-10-29
Fix using wxWidgets on macOS

### LINKED ISSUE #2102: wxWidgets not initialized properly on macOS causing unfocusable window
On macOS (15.7.1), the app does not show an icon in the dock, and the window opens in the background. When trying to select it via Mission Control, it cannot be focused. It seems this issue is due to `wxEntry` not being called in initialization. When initializing the standard way, it works as expected, eg:

```diff
diff --git a/src/gdl.cpp b/src/gdl.cpp
index e581a7a7..0c6dd7a7 100644
--- a/src/gdl.cpp
+++ b/src/gdl.cpp
@@ -222,7 +222,22 @@ namespace MyPaths {
 }
 }
 
+#ifdef HAVE_LIBWXWIDGETS
+class MyApp : public wxApp
+{
+public:
+    virtual bool OnInit();
+};
+wxIMPLEMENT_APP(MyApp);
+int mymain(int argc, char *argv[]);
+bool MyApp::OnInit()
+{
+  return !mymain(argc, argv);
+}
+int mymain(int argc, char *argv[])
+#else
 int main(int argc, char *argv[])
+#endif
 {
 
   // indicates if the user wants to see the welcome message
diff --git a/src/gdlwidget.cpp b/src/gdlwidget.cpp
index 3794922d..d30958f2 100644
--- a/src/gdlwidget.cpp
+++ b/src/gdlwidget.cpp
@@ -653,7 +653,7 @@ extern "C" { void CPSEnableForegroundOperation( ProcessSerialNumber* psn ); }
         return new wxAppGDL;
     }
     wxAppInitializer  wxTheAppInitializer((wxAppInitializerFunction) wxCreateApp);
-#else
+#elif 0
     wxApp& wxGetApp() { return *static_cast<wxApp*>(wxApp::GetInstance()); }
     wxAppConsole *wxCreateApp()
     {
```

mohd-akram:
Fixes #2102

### DIFF
--- a/src/gdlwidget.cpp
+++ b/src/gdlwidget.cpp
@@ -1413,6 +1413,7 @@ bool GDLWidget::InitWx() {
   // program was called from the command line (and isn't a bundle)
   try{
      wxInitialize();
+     wxGetApp().CallOnInit();
   } catch (...) {return false;}
  //avoid using if no Display is present!
   wxDisplay d;


GillesDuvert:
@mohd-akram thanks a lot! You seem to have a good background in wxWidgets, that we do not have. Initializing correctly the widgets system on linux, OSX and Windows, and having them behave, is still a mystery for me, there are no examples for the particular case of GDL where the main app /eventloop  is not wxwidgets but some custom internal gdl event loop. I suspect many misunderstandings in our code, if you have some time for a review in the future, please do not hesitate.

mohd-akram:
> @mohd-akram thanks a lot! You seem to have a good background in wxWidgets, that we do not have. Initializing correctly the widgets system on linux, OSX and Windows, and having them behave, is still a mystery for me, there are no examples for the particular case of GDL where the main app /eventloop is not wxwidgets but some custom internal gdl event loop. I suspect many misunderstandings in our code, if you have some time for a review in the future, please do not hesitate.

I wish :) I stumbled into this solution [here](https://stackoverflow.com/questions/208373/wxwidgets-how-to-initialize-wxapp-without-using-macros-and-without-entering-the) (then found that it was used in deprecated code [here](https://github.com/gnudatalanguage/gdl/blob/417f76e6a8ade2fc5260c08f40ff58a2e7bf60aa/src/plplot/drivers/deprecated_wxwidgets.cpp#L1357)). Previously I was trying to incorporate `wxEntry` which required some minor refactoring to use, but found this one-liner which achieves the same thing. The `Yield()` approach seems to work on macOS (I haven't tried extensively), so maybe that can be done for all platforms? That seems like the correct way to do things, but I'm not an expert. I think that would then allow using `wxEntry` as I don't think it would work otherwise in conjunction with the custom loop since it blocks forever.

GillesDuvert:
@mohd-akram Unfortunately, the "Yield()" solution never worked on Windows, and partially on linux. May be I should test again after this PR. Will merge ASAP. Thanks again.

Issue #2105, opened 2025-10-31
Relaxed Structure Assignment


briardew:
Forgive me if this is known or covered by another issue, but relaxed structure assignment (see the section in [https://www.nv5geospatialsoftware.com/docs/Structures.html](https://www.nv5geospatialsoftware.com/docs/Structures.html)) does not work in GDL.

Example code:
```
x0 = {name:'', value:''}
xa = replicate(x0, 1)
xa[0] = {name:'test', value:['test']}
```
Works fine in IDL, throws a "conflicting data structures" error in GDL.

briardew:
I'm actually not sure this is relaxed structure assignment, but it certainly works in IDL and not in GDL and is causing me some headaches. Sorry, I'm not an IDL user, just trying to convert some code.

GillesDuvert:
looks like another problem due to GDL having problems with 1-element array being the same as no array at all (see #2083) (or not). 
Looks like a regression otherwise we would have heard about that earlier.

GillesDuvert:
not a regression, actually. This behaviour is here from the beginnings, and was not reported earlier (thanks).
This is an unmentioned effect of "relaxed structure assignment" as the result of  `xa[0] = {name:'test', value:['test']}` which should have been: 
```
IDL> help,xa
** Structure <274e7a88>, 2 tags, length=32, data length=32, refs=2:
   NAME            STRING    'test'
   VALUE           STRING    Array[1]
```
is instead
```
IDL> help,xa
** Structure <274e7a88>, 2 tags, length=32, data length=32, refs=2:
   NAME            STRING    'test'
   VALUE           STRING    'test'
```
which is what relaxed structure assignment does.

GillesDuvert:
Need to revert [dd37ea0](https://github.com/gnudatalanguage/gdl/commit/dd37ea05b5230af2e670b861802b949ae1d2801f) as the patch for this issue  (https://github.com/gnudatalanguage/gdl/issues/2105) is not what should be done.

PR #2118, opened 2025-11-15
Closes #2083 and #2105

### LINKED ISSUE #2083: GDL adds dimension when extracting substructure from array of structures
GDL behaves differently from IDL when extracting a sub-structure from an array of structures.
GDL adds a dimension (length 1) to the extracted array of sub-structures.
This can break heritage IDL code (via wrong indexing of the extracted array of sub-structures).

E.g.
a=replicate({x:{y:1}},3)

GDL> help,a.x
<Expression>    STRUCT    = -> <Anonymous> Array[1, 3]
IDL> help,a.x
<Expression>    STRUCT    = -> <Anonymous> Array[3]

This probably happens because of the oddity in IDL that x appears like an array with 1 element despite it clearly being defined to be scalar.
("help,a" shows a.x is "STRUCT    -> <Anonymous> Array[1]")

However GDL is not fully consistent - it behaves like IDL for extracting tags within the substructure:

GDL> help,a.x.y
<Expression>    INT       = Array[3]
(IDL gives the same)

Note that GDL also behaves the same as IDL if x really is declared as an array e.g. in case
a=replicate({x:replicate({y:1},4)},3)
GDL> help,a.x
<Expression>    STRUCT    = -> <Anonymous> Array[4, 3]
(IDL gives the same)

GillesDuvert:


GillesDuvert:
warning errors in #issues references above. read: #2083 and #2105

PR #2119, opened 2025-11-16
Small changes to avoid warnings from clang compiler

GillesDuvert:


GillesDuvert:
separate test show this compiles and runs excellently on "Sequoia" OSX 15.5 ARM64.

PR #2134, opened 2025-12-01
add subdir for MPFIT procedures to be visible in testsuite/

alaingdl:


alaingdl:
@jtappin I rewrite the code "gaussian_function.pro" to be closer to the refrence. The associated test "test_gaussian_function.pro" is now working fine in GDL and indepandly in IDL

PR #2143, opened 2025-12-12
add Ubuntu-ARM CI jobs; mark three tests as XFAIL for ARM arch

slayoo:


slayoo:
compilation on ARM OK with the following 3 tests failing:
```
 The following tests FAILED:
	 65 - test_byte_conversion.pro (Failed)
	100 - test_formats.pro (Failed)
make: *** [Makefile:71: test] Error 8
```

GillesDuvert:
@slayoo  the 3 tests will fail on ARM machines because rounding see https://developer.apple.com/documentation/apple-silicon/addressing-architectural-differences-in-your-macos-code#Audit-Code-that-Contains-Float-to-Int-Conversions

I would support changing theses tests on arm machines to reflect the fact that float-cast-overflow behaviour is not the same on IA and ARM.

slayoo:
Hi @GillesDuvert!

> @slayoo the 3 tests will fail on ARM machines because rounding see https://developer.apple.com/documentation/apple-silicon/addressing-architectural-differences-in-your-macos-code#Audit-Code-that-Contains-Float-to-Int-Conversions
> 
> I would support changing theses tests on arm machines to reflect the fact that float-cast-overflow behaviour is not the same on IA and ARM.

here's an attempt to mark these as "expected failure": https://github.com/gnudatalanguage/gdl/pull/2143/commits/f410732c2a9bc459b148ac56315da94c377e0639

PR #2144, opened 2025-12-12
CI: bump macOS Intel runners from macOS 13 (announced to be discontinued) to macOS 15

slayoo:

PR #2145, opened 2025-12-13
Drop wx-3.0 include dir

opoplawski:
This fixes building with wx 3.2 in Fedora rawhide


slayoo:
seems to solve #2146, and helps with #2144 
thanks!

GillesDuvert:
see #2147 , this #include was unintentional and apparently inserted by NetBeans. Removed directly in the GITHUB code instead, but thanks for the quick report:

PR #2147, opened 2025-12-13
Fix include

jkohnert:
Recent PR for #2141 breaks build on current Archlinux, since wx header is in another subdir than "wx-3.0", Stripped unnneded path in include and moved include after the `#ifdef`. 

Need to look into the apply build error...

### DIFF
--- a/src/gdlwxstream.cpp
+++ b/src/gdlwxstream.cpp
@@ -15,14 +15,12 @@
  *                                                                         *
  ***************************************************************************/
 
-#include <wx-3.0/wx/graphics.h>
-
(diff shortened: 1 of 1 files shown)


GillesDuvert:
@jkohnert thanks a lot, but this is surprising: I never added `#include <wx-3.0/wx/graphics.h>` . I can only suppose it has been added automatically (I use an IDE, NetBeans). This is definitely something I must change in my NetBeans setup. Actually, gdlwxstream.cpp does not need any "`include <wx...`". I'll patch this directly in the github code.

PR #2152, opened 2025-12-15
avoid failures of rebuild on github's osx or windows

GillesDuvert:
After painful tests, these failures were due to GitHub ("Travis") changes that need to be analyzed each time by a knowledgeable person. The symptoms are:

1. GDL can be compiled / tested without errors on individual (personal) machines
2. same GDL version rebuils stops prematurely/ lots of errors when rebuild automatically on Travis (the "Checks button above)

In this case the PR contents are not culprit and the PR is candidate to merging, irrespective of the "Checks" failures. AND the conditions of rebuild on Travis (use of CACHE, workflow .yml code .... ) must be checked by  a knowledgeable person (who?).

### DIFF
--- a/scripts/build_gdl.sh
+++ b/scripts/build_gdl.sh
@@ -369,6 +369,7 @@ function prep_packages {
             exit 1
         fi
         brew update-reset
+		export SDKROOT=$(xcrun --show-sdk-path) #solves IA64 brew problems?
         log "Installing packages: ${BREW_PACKAGES[@]}"
         if [ ${DRY_RUN} == "true" ]; then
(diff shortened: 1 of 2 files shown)


GillesDuvert:
Windows rebuild was failing because of CACHE.
Travis CACHE seems to be used only with the windows build?
Cache should be deleted and checks replayed BEFORE any attempt at solving what may be an unexisting problem.

GillesDuvert:
OSX rebuilds WILL most certainly need some changes to build_gdl.sh anytime the Travis virtual machine are updated, so ANY rebuild may fail at spurious times for reasons not related to changes in GDL code.

Issue #2114, opened 2025-11-13
Compilation Fails on Github's OSX


GillesDuvert:
See https://github.com/gnudatalanguage/gdl/actions/runs/19301035710/job/55195296130 , full log.
There is a lot of warnings but also 2 errors about system includes not found.
All this points towards a confusion arising during the build_gdl.sh setup of the machine (build_gdl.sh prep) where the include path is not properly configured by the homebrew installs.
Probably because of this message:

```
Using `clang`, `clang++`, etc., requires a CLT installation at `/Library/Developer/CommandLineTools`
If you don't want to install the CLT, you can write appropriate configuration files pointing to your SDK at ~/.config/clang.
```

**Hope somebody friendly with OSX will help.** Until then I ignore the OSX errors.
It seems that this behaviour appears only on a 'fresh' GitHub OSX machine, not on user's Macs where the necessary OSX SDK has been manually installed.

PR #2156, opened 2025-12-17
test hypothesis

GillesDuvert:
…for the culprit for OSX failure of compilation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@
 #  (at your option) any later version.
 #
 
-cmake_minimum_required(VERSION 3.13.3 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.16 FATAL_ERROR)
 
 if (NOT EXISTS "${CMAKE_SOURCE_DIR}/src/whereami/.git" )
   message(FATAL_ERROR "The src/whereami git submodule is not initialised.\n Please run `git submodule update --init`")


alaingdl:
not sure to understand the goal, cmake is in version  4.2.1 on my mac

and test again the script : not working at all
(but compilation OK with cmake ..)

GillesDuvert:
@alaingdl  The code-related failures on OSX and Windows are visible only just now, after PR #2152 has been merged.

We now can find bugs on these architectures. That was not possible since one or 2 month befor my issue #2114.

I have looked carefully to all the rebuilds in the last 2 months. As nobody proposed any help on the Windows and OSX failures of rebuild I was obliged to look into that myself, when I have no access to such many different machines and the web forums suggestions are always irrelevant.

The OSX rebuild on Travis fail for both Intel and Arm architectures when `cmake_minimum_required(VERSION 3.13.3 FATAL_ERROR)` (due to lack of finding NETCDF)
If one replaces by the (older version) `cmake_minimum_required(VERSION 3.16 FATAL_ERROR)`
then "at least" the OSX Intel build is OK. **meaning that netcdf libraries are found by 3.16 and not by 3.13.3** .

Strangely the ARM build still has the failure.
And, more strangely, the Windows build find netcdf and builds GDL, but all the tests using netcdf (shoud probably be factorized in one big netcdf test) fail.

There is something rotten about our findNetCdf.cmake on OSX (and now Windows?) (I see a netcdf-cxx homebrew that we do not require, and is not needed on OSX-intel anyway) and probably some adjustement in the netcdf-related code is neceesary for Windows. 

A simple solution to avoid the burden of incomplete reconstruction on Travis OSX and Windows is to remove netcdf from the GDL build prerequisites. I would like to see green flags when we publish new GDL versions.

GillesDuvert:
Apparently this specific problem is not appearing anymore. Closing.

Issue #701, opened 2020-01-14
READS in error with complex


GillesDuvert:
3 special cases for complex that are not yet OK with GDL
```
 f1=complex(12,13) & f2=complex(18,19)
 f=complexarr(2) & inarray=' ( 12,13) (18,19) ' & reads,inarray,f ; 2 complex written correctly: real and imaginary part readable
 if total(atan(f-[f1,f2],/phas)) ne 0.0 then errors++

```

```
f1=complex(12,13) & f2=complex(18,0) & f3=complex(19,0)
f=complexarr(3) & inarray=' ( 12,13) 18 19' & reads,inarray,f ; 3 complex mixed writing: 1st real and imaginary part, rest is real only
if total(atan(f-[f1,f2,f3],/phas)) ne 0.0 then errors++

```
 ```
f1=complex(12,0) & f2=complex(13,0)
 f=complexarr(2) & inarray=' 12,13 ' & reads,inarray,f
 if total(atan(f-[f1,f2],/phas)) ne 0.0 then errors++


```

GillesDuvert:
solving this is a bit tricky but part of an ongoing work that has been delayed for more pressing business.

alaingdl:
I already wrote an extension to _test\_reads.pro_

Do you plan to work on it now, or may I try ?

GillesDuvert:
OK @alaingdl I have just a small, unrelated, patch "under the elbow". The main changes in  https://github.com/gnudatalanguage/gdl/pull/719/commits/2d6203f3c42b2459bbd3e620bcb70d4c098ceb84 are already there.
Please proceed! Thanks!

alaingdl:
This missing feature is now available. A large number of test cases has been added to _test\_reads.pro_

Even it should be solved now, please noticed that some cases can be very tricky, and the code should resist to a mixing of various amount of separators (space, tabs, comma ...) Please report as accurate as possible if problem. (this is also why so much examples are provided in _test\_reads.pro_)

Please open a *new* bug report if problem found.

PR #2166, opened 2026-01-15
Patch a READS bug.

GillesDuvert:
reads would exit too soon from its accept_comma_and_complex_default_format_string() function if whitespaces were present on the string line before an ending linefeed. This would produce an stream truncation problem.


GillesDuvert:
Bug was located in a C routine added to comply with #701 but that IMHO is slow and complicated (already 2 patches on it since creation) and there should probably be a better, more C++, solution.

PR #2168, opened 2026-01-21
Implement LAMBDA and LAMBDAP

GillesDuvert:
3 differences with IDL's LAMBDAS:
 1. lambda text must always be a string (between " "). Needs a change in the ANTLR compiler to do otherwise.
 2. lambdas are here but their use is limited since GDL variables are not yet objects that would know about the MAP, REDUCE and FILTER methods.
 3. Using a lambda in QROMB() needs to have defined it before:
```
GDL> compile_opt idl2
GDL> result = QROMB("LAMBDA(x:x^3 + (x-1)^2 + 3)", -4, 4)
% Function not found: LAMBDA(X:X^3 + (X-1)^2 + 3)
% Execution halted at: $MAIN$          
GDL> zzz=LAMBDA("x:x^3 + (x-1)^2 + 3")
GDL> result = QROMB(zzz, -4, 4)
GDL> result
       74.666664
```
Obviously having lambdas more inlined, thus changing the Parser, would be great.

### DIFF
--- a/src/GDLInterpreter.cpp
+++ b/src/GDLInterpreter.cpp
@@ -415,15 +415,15 @@ GDLInterpreter::GDLInterpreter()
 	}
 	catch ( GDLException& e) {
 		
-		// errors while typing commands in inner loop (called via Control-C, STOP, error) should be ignored.
-		if (e.Interpreter()->IsInnerInterpreterLoop()) {
-		  e.Interpreter()->SetInnerInterpeterLoop(false);
-		  _retTree=NULL;
-		  debugMode=DEBUG_CLEAR;
-		  // we cannot throw, here, it caused #2017 
-		  Warning(e.getMessage());  //just warn that there was a (typing ?) error see #1855
-		return RC_OK;
-		}
+				// errors while typing commands in inner loop (called via Control-C, STOP, error) should be ignored.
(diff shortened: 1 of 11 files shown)


alaingdl:
This is really smart, and I can compile it with gcc 8, but I can NOT compile with gcc 11, due to object.cpp

 31%] Building CXX object src/CMakeFiles/gdl.dir/objects.cpp.o
/home/coulais/GDL/gdl-1.1.3git260123CMake/src/objects.cpp: In function ‘int ProIx(const string&)’:
/home/coulais/GDL/gdl-1.1.3git260123CMake/src/objects.cpp:1048:48: error: incomplete type ‘DInterpreter’ used in nested name specifier
 1048 |   EnvT* requestedScope = (EnvT*) DInterpreter::CallStackBack();
      |                                                ^~~~~~~~~~~~~
/home/coulais/GDL/gdl-1.1.3git260123CMake/src/objects.cpp: In function ‘int FunIx(const string&)’:
/home/coulais/GDL/gdl-1.1.3git260123CMake/src/objects.cpp:1071:50: error: incomplete type ‘DInterpreter’ used in nested name specifier
 1071 |     EnvT* requestedScope = (EnvT*) DInterpreter::CallStackBack();

Issue #2164, opened 2026-01-12
READ_CSV seems to consistently skip first line


rcasini:
[MODEL_SiO2.csv](https://github.com/user-attachments/files/24562327/MODEL_SiO2.csv)

I'm trying to run a code that I wrote (and properly runs) in IDL. The code needs to ingest a model from file, but the READ_CSV in GDL gives different results from IDL, consistently returning a structure that has a dimension shorter by 1 than the expected one, which should correspond to the total number of entries (lines) in the model file. I'm adding a sample model file, so you can try reproduce the issue using the command

DUMP=READ_CSV('MODEL_SiO2.csv')

and confirm that DUMP is a structure with five field arrays of size 7 instead of 8 as it should be based on the content of the model file. The missing information is always the one in the first line. 

I'm using GDL v1.1.3, freshly compiled on OpenSUSE TW.

Thank you!
Roberto

alaingdl:
I can reproduce your bug

I gave a look at the code, unclear for me why this first line is skipped

I am not confortable with the way the file is parsed,
reading several times "line", and the file is not closed at the end

I am not sure the block below
; IDL skips blank lines after this position.
is not redone later in

```
  for iline=0L,restalire-1 do begin
    READF, lun, line
    if (STRLEN(STRTRIM(line,2)) eq 0) then continue
 ```
and line `     if (numrec gt 0 and count ge numrec) then break `
is unclear to me

just to say I am reluctant to make change to this code in an hurry

rcasini:
Thank you, Alain.

The issue seems to be that the first block you mentioned in your message correctly reads the first line, but then its content is dumped into a header variable (by the instruction header=subline), whereas the structure to be returned is built iteratively in the following block that you also pasted in your message. However, because the reading location in the file hasn't been set back to the top, the first line is now lost, and the structure will only contain the content from line 2 down.

alaingdl:
First line was not put in the pointer, the in the structure. I found a way to a temporary workaround

I can send it to you by email ... Don't hesitate to email me !

I think the code needs a few cleanup and some tests before I put it in the Git !

alaingdl:
I have a student now, when he will have finish a improved FILE_DELETE version, I will ask him to have a look on that !

rcasini:
> First line was not put in the pointer, the in the structure. I found a way to a temporary workaround
> 
> I can send it to you by email ... Don't hesitate to email me !
> 
[...]

PR #2171, opened 2026-01-27
revised version of READ_CSV()

alaingdl:
* Revised version of READ_CSV() coming with a test (test_csv.pro)
* small change in io.hpp (advice from g++ -Wreorder)
* new small function to test type conversion
* rewritten TEST_VOIGT to really check the numbers !


alaingdl:
No surprise, let's go ! Close #2164 for basic CSV files

We have to rework on FIX() and FILE_DELETE, arriving soon

PR #2183, opened 2026-03-23
Fixed WIDGET_LIST scrolling not being auto-activated when it should

frankmri:
This fixes the scroll bars not getting shown when list size is beyond the widget size (at least when the list widget is instantiated). 

I'm submitting it for your review
Best


GillesDuvert:
@frankmri thanks for your contribution! Not tested on all platforms yet, but should be ok.

PR #2184, opened 2026-03-26
DLM work

GillesDuvert:
WIP
Huge amount of corrections to the first version of the DLM support, as trying to get the CDF dlm work "as is" with GDL showed that many IDL_XXX functions description in the documentation were either misread or not sufficiently documented.


GillesDuvert:
replaced by #2191

PR #2210, opened 2026-07-19
Removing INVERT() memory leaks and close #2206

### LINKED ISSUE #2206: INVERT(Matrix) leaks memory
By default, gdl uses the eigen3 template library to implement matrix invert(/DOUBLE) where matrix = DBLARR(N,N)
My code iterates on squared covariance matrices (N < 1200) removing outliers 1 by 1, so matrices are smaller until abs(residuals) < 3 sigma.

It works fast on idl... but it consumes 25gb ram with gdl 1.0.4 (ubuntu 24.4) or latest release 1.1.3 compiled on my machine with eigen3.
When I use cmake -DEIGEN3=OFF, the same program is faster (gsl) and takes 3gb max to run.

Workaround: disable eigen3 by compiling with -DEIGEN3=OFF, should it be the default again, until this huge memory leak is fixed.

PS: I talked with @GillesDuvert that looked at eigen_invert_fun and he wrote a simple reproducer.

Keep up the good work,
FOSS rocks!

Goodbye & good luck,
Laurent

GillesDuvert:
#2206 reports the eigen invert() flavor to create memory leaks.  I noted that the GSL "flavor" also had memory leaks in some cases.

This PR insures that all memory leaks are removed (hopefully, the number of cases makes this touchy), remove some uneccessary allocations/conversions at the expense of a few if/else, factorize, and correct a logic error related to the use of the /DOUBLE keyword. Return from the gsl variant is not similar to what IDL returns in the case of a not inversible matrix (not corrected: seems harmless . To correct, take example from the eigen variant).

The EIGEN:: variant is (for me) 10-30% more efficient than GSL for ALL matrices sizes (this may not be true once the gsl version is modified to not copy the input data in a preallocated gsl structure!) . Both are more efficient than IDL's. Really huge matrices not tested (max 3000x3000 : 15 seconds)

### DIFF
--- a/src/matrix_invert.cpp
+++ b/src/matrix_invert.cpp
@@ -111,6 +111,75 @@ namespace lib {
     bool hasDouble=e->KeywordSet(DOUBLEIx);
     
     matrix_input_check_dims(e);
+    
+    BaseGDL* p0 = e->GetParDefined( 0);
+    SizeT nEl = p0->N_Elements();
+
+    // only one element matrix: no need of eigen or gsl.
+    if (nEl == 1) {
+      SizeT nParam = e->NParam();
+      if (nParam == 2) e->AssureGlobalPar(1);
+      DLong singular=0;
+      if (p0->Type() == GDL_COMPLEXDBL) {
+        DComplexDblGDL* res = static_cast<DComplexDblGDL*>
+            (p0->Convert2(GDL_COMPLEXDBL, BaseGDL::COPY));
+        double a, b, deno;
+        a = real((*res)[0]);
+        b = imag((*res)[0]);
+        deno = a * a + b*b;
+        if (deno == 0.0) {
+          singular = 1;
+          (*res)[0] = DComplexDbl(0., 0.);
+        } else {
+          (*res)[0] = DComplexDbl(a / deno, -b / deno);
+        }
+        if (nParam == 2) e->SetPar(1, new DLongGDL(singular));
+        return res;
+      }
+
+      if (p0->Type() == GDL_COMPLEX) {
+        DComplexGDL* res = static_cast<DComplexGDL*>
+            (p0->Convert2(GDL_COMPLEX, BaseGDL::COPY));
+        float a, b, deno;
+        a = real((*res)[0]);
+        b = imag((*res)[0]);
+        deno = a * a + b*b;
+        if (deno == 0.0) {
+          singular = 1;
+          (*res)[0] = DComplex(0., 0.);
+        } else {
+          (*res)[0] = DComplex(a / deno, -b / deno);
+        }
+        if (nParam == 2) e->SetPar(1, new DLongGDL(singular));
+        return res;
+      }
+
+
+      if ((p0->Type() == GDL_DOUBLE) || hasDouble) {
+        DDoubleGDL* res = static_cast<DDoubleGDL*>
+            (p0->Convert2(GDL_DOUBLE, BaseGDL::COPY));
+        if ((*res)[0] == 0.0) {
+          singular = 1;
+        } else {
+          double unity = 1.0;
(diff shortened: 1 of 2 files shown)


bourgesl:
I tested your patch with my code and INVERT looks better (/GSL and default ie EIGEN) ie low memory footprint... ~ 950mb for gsl but 3gb for eigen  (low slope ie few mem leaks are remaining)... 
I will stay with /GSL to have the lowest memory footprint for now (faster for me too).
Congratulations!!

Finally eigen leaked 40gb, then crashed my linux...
[20260720_170215.jpg](https://github.com/user-attachments/assets/d2db259c-d05e-4842-a2f7-e49dc640eda1)

bourgesl:
Here is my shell script to use heap profiler with jemalloc
```
#!/bin/bash
source env.sh

# Notes: enable and activate profiling, generate a profile every ~1GB of additional memory allocated.
#   The profile-dump interval is expressed somewhat cryptically as $\log_2$\ of the memory quantity:
#   here 1GB == 30 bits.
export MALLOC_CONF="prof:true,prof_active:true,lg_prof_interval:30,prof_prefix:jeprof.out"

LOCAL_PRELOAD=/home/bourgesl/apps/jemalloc/lib/libjemalloc.so.2

export LD_PRELOAD=${LOCAL_PRELOAD}

set -eux;

echo "script_make_jsdc_v5"
gdl -e "script_make_jsdc_v5 " | tee run.log

```
And the conversion command to turn heap files to SVG call graph:
```
#!/bin/bash
export JEPROF='/home/bourgesl/apps/jemalloc/bin/jeprof'
DIR=/home/bourgesl/dev/SearchCalServer-gh/SearchCal/alx/config/JMDC/jsdc_soft

# The basic form of the command:
${JEPROF} --svg /usr/bin/gdl $DIR/MY_FILE.heap > MY_FILE-heap.svg

```
for example:
`/home/bourgesl/apps/jemalloc/bin/jeprof --svg /home/bourgesl/dev/gdl/dist/bin/gdl ./jeprof.out.128118.2.i2.heap > jeprof.svg
`

GillesDuvert:
@bourgesl INVERT has as many cases as there are byte,int,long,long64,floats,double,complex...etc plus the overhead of the /DOUBLE option and the 2 flavors.
According to your return it is tempting to remove completely the use of eigen for INVERT, with the added bonus of a simpler and more maintainable code.

bourgesl:
At least this memory leak crashed my laptop few times (48gb allocated) so it is a critical/ blocker to me.
As recommended, please disable eigen by default, ie adopt gsl (working fine) until the eigen wrapper is made stable again.
🙏
PS: Jemalloc prof=leak
https://github.com/jemalloc/jemalloc/wiki/Use-Case%3A-Leak-Checking

MALLOC_CONF=prof_leak:true,lg_prof_sample:0,prof_final:true \

See the previous run-prof.sh
https://github.com/gnudatalanguage/gdl/pull/2210#issuecomment-5023848684

alaingdl:
GDL> #invert
Internal FUNCTION : res=INVERT([2 Args],DOUBLE,GSL,EIGEN)

You already can select between both just by activating the keyword  ...
 Ok it is not IDL complaint, but it was useful long time ago

Compiling GDL without Eigen is as simple as doing : `cmake . -DEIGEN3=off`

HTH

Alain

bourgesl:
Yes I built my own gdl but linux packages (ubuntu), github releases or default build should set EIGEN=OFF too as these releases are not stable, please

GillesDuvert:
OK GSL is now preferred. And faster, now, avoiding data copies as much as possible. Use **gsl_matrix_wiew** !

One lesson: Eigen:: makes a lot of assumptions and memory management that can (and do in this case) severely disrupt our memory management --- because we use Eigen:: too naively and should have read the small print.

Issue #2212, opened 2026-07-22
regression in message when function not found : %  FCALLNode::Eval - AutoObj


alaingdl:
(The process is OK for procedures !)

Instead of having `% Function not found:`  we got : `%  FCALLNode::Eval - AutoObj`

I suspect just a detail in last PR from  @GillesDuvert  ...

Few days ago we had 

```
  GDL - GNU Data Language, Version v1.1.3-21-g1e1fc0ae-dirty

GDL> z=refer()
% Function not found: REFER
% Execution halted at: $MAIN$          
GDL> refer
% Attempt to call undefined procedure: REFER
% Execution halted at: $MAIN$          
```

Now we have : 
```
    "RELEASE": "1.1.3-33-g21bbb4af-dirty",

GDL> a=refer()
%  FCALLNode::Eval - AutoObj
% Execution halted at: $MAIN$          
GDL> refer
% Attempt to call undefined procedure: REFER
% Execution halted at: $MAIN$          
```

GillesDuvert:
now, according to idl's documentation, an object named "refer" can be created with toto=refer()
so not only gdl looks for refer.pro but also refer__define.pro
obviously the message should be more friendly.
But this is not a bug or a regression.

alaingdl:
Ok !! I did not know ! Sorry.
I tested with IDL 8.5 and 9.1 and the messages are the same : 

```
IDL> refer
% Attempt to call undefined procedure: 'REFER'.
% Execution halted at: $MAIN$          
IDL> a=refer()

a=refer()
         ^
% Syntax error.

```
Not sure such message is very useful for beginners ! Clearly previous message by GDL is better.

Issue #2200, opened 2026-06-11
Windows build broken on weekly build, 1.1.2 latest doesn't have a windows version


drmclem:
Hi

Downloaded the weekly build version and it fails on entry point not found libaws-cpp-sdk-s3.dll - the latest release version doesn't have a windows build - last version which works is GDL - GNU Data Language, Version v1.1-dirty. 

<img width="548" height="224" alt="Image" src="https://github.com/user-attachments/assets/f881b4a8-5d08-4b58-8bd9-8e4575a61284" />

Matthew

GillesDuvert:
@drmclem thanks for the report, this seems pretty strange as gdl does absolutely not need any interface to Amazon Web Services (and if it does, then this dependency must be removed at once!)
(This seen from my linux point-of-view of course)
Probably a system library used by GDL or the windows installer needs it. Afraid I cannot help you on a Windows subject, hopefully somebody else will have an answer.

drmclem:
these are all being installed from the msi installed from the weekly assets

<img width="942" height="767" alt="Image" src="https://github.com/user-attachments/assets/d9fa33d3-c59d-4b5c-bedb-ad6826762e9b" />

Also these files are in the .zip file as well as the gdlsetup...msi

fawltylanguage:
HDF5 dependency?

GillesDuvert:
probably just download it from the web? I found https://www.dllme.com/dll/files/aws-cpp-sdk-s3

drmclem:
Hi all - I just wanted to update on some info on this one and windows building in general. Managed to set up and build gdl with windows on Win 11 but have several issues with mixed package types which may explain the failure of windows build asset from earlier. In short, 

- yes the quick one - libaws is a dependency on hdf5

The longer one - 

- When a default MSYS2 is installed it presents 4 working environments and by default opens the ucrt runtime linkage, but the build scripts and instructions on the WIKI don't correctly pick this up instead defaulting to the MSVCRT versions so you can end up with odd packages being collected for the install. When running the UCRT64 environment, everything is under /ucrt64 not /mingw64 but the scripts copy from the deep link - /msys64/mingw64 ... which is not the default).

The MSYS2 does now seem to provide useful environment variables to help with this - I have made the necessary mods to the build script and tested it on my build system for the ucrt and it all seems to work better now (as an aside - I wanted the UCRT version to use with the CALL_EXTERNAL function)

I have update the build_gdl.sh to use the provided environment variables 
```
[...]

Issue #2177, opened 2026-02-24
test_rounding is failing on ppc64le and s390x


opoplawski:
with 1.1.3 Fedora rawhide.  

ppc64le:
```
164/210 Test #166: test_rounding.pro ..................***Failed    0.72 sec
% Compiled module: DLM_REGISTER.
% Compiled module: TEST_ROUNDING.
% Compiled module: BANNER_FOR_TESTSUITE.
% Compiled module: GDL_IDL_FL.
% TEST_ROUNDING_TYPE: 
===========================================================
% TEST_ROUNDING_TYPE: testing /ROUND
% Compiled module: GIVE_LIST_NUMERIC.
% Compiled module: ERRORS_ADD.
% TEST_ROUNDING_TYPE: Error on operation : bad value type BYTE
Neg. side
% TEST_ROUNDING_TYPE: Error on operation : bad value type UINT
Neg. side
% TEST_ROUNDING_TYPE: Error on operation : bad value type ULONG
Neg. side
% TEST_ROUNDING_TYPE: Error on operation : bad value type ULONG64
Neg. side
% TEST_ROUNDING_TYPE: =                                                        =
% TEST_ROUNDING_TYPE: =  4 errors encountered during TEST_ROUNDING_TYPE tests  =
% TEST_ROUNDING_TYPE: =                                                        =
% Compiled module: ERRORS_CUMUL.
% TEST_ROUNDING_TYPE: testing /CEIL
% TEST_ROUNDING_TYPE: Error on operation : bad value type BYTE
Neg. side
% TEST_ROUNDING_TYPE: Error on operation : bad value type UINT
Neg. side
% TEST_ROUNDING_TYPE: Error on operation : bad value type ULONG
Neg. side
% TEST_ROUNDING_TYPE: Error on operation : bad value type ULONG64
Neg. side
% TEST_ROUNDING_TYPE: =                                                        =
% TEST_ROUNDING_TYPE: =  4 errors encountered during TEST_ROUNDING_TYPE tests  =
% TEST_ROUNDING_TYPE: =                                                        =
% TEST_ROUNDING_TYPE: testing /FLOOR
% TEST_ROUNDING_TYPE: Error on operation : bad value type BYTE
Neg. side
% TEST_ROUNDING_TYPE: Error on operation : bad value type UINT
Neg. side
% TEST_ROUNDING_TYPE: Error on operation : bad value type ULONG
Neg. side
[...]

Issue #2176, opened 2026-02-24
test_plot_ranges fails on ppc64le


opoplawski:
Updating the fedora package to 1.1.3 seeing this failure only on ppc64le:
```
153/210 Test #147: test_plot_ranges.pro ...............***Failed    1.81 sec
% Compiled module: DLM_REGISTER.
% Compiled module: TEST_PLOT_RANGES.
% Compiled module: ROUTINE_NAME.
% Compiled module: CHECK_IF_DEVICE_IS_OK.
% Compiled module: BANNER_FOR_TESTSUITE.
% Compiled module: GDL_IDL_FL.
% TEST_PLOT_YRANGES : 
  NO errors encountered during TEST_PLOT_YRANGES **POSITIVE** tests  
% Compiled module: ERRORS_CUMUL.
% TEST_PLOT_YRANGES : 
  NO errors encountered during TEST_PLOT_YRANGES **NEGATIVE** tests  
% Compiled module: ERRORS_ADD.
% TEST_PLOTTING_RANGES: Error on operation : error case 4 on !Y.crange
% TEST_PLOTTING_RANGES: Error on operation : error case 4 on !Y.crange
% TEST_PLOTTING_RANGES: Error on operation : error case 4 on !Y.crange
% TEST_PLOTTING_RANGES: Error on operation : error case 4 on !Y.crange
% TEST_PLOTTING_RANGES: Error on operation : error case 4 on !Y.crange
% TEST_PLOTTING_RANGES: 
  5 errors encountered during TEST_PLOTTING_RANGES tests  
% TEST_PLOT_RANGES: ========================================================
% TEST_PLOT_RANGES: =                                                      =
% TEST_PLOT_RANGES: =  5 errors encountered during TEST_PLOT_RANGES tests  =
% TEST_PLOT_RANGES: =                                                      =
% TEST_PLOT_RANGES: ========================================================
```

PR #2216, opened 2026-07-30
Remove real-time search for routines names and more

GillesDuvert:
1. GDL is constantly looking for values associated to names: variable names, key names, and routine names. It unfortunately does it with a naive loop on the corresponding unsorted string vector. Worse, it does it each time a routine is called, a variable is passed etc. I have modified the code so that the routine names are in an unordered set, and, when searched, the index of the name is the index of the corresponding Pro() or Fun() entry. This makes it quite faster.
2. Another point of slowness was the possibility in a C++ routine to look to a keyword using its name, not its index. This is now discouraged (Marc Schellens already warned about this a long time ago). In _almost all cases_, using a static int as the return of e->KeywordIx() insures that all further calls to the routine will not look again for the (same) index. All the code using non-static such ints has been reveiewed and converted to static when possible (almost always). The value for` !ERR ` was frequently searched in the code, now a specialzed function removes this need.
3. In contrary, it is very difficult to instrument the GDL code to avoid naive O(N) name searches for keywords and variables (say: sort the name array and perform O(log(N)) binary searches), especially as the variables passed by GDL to a function are in an order that mangles keyword-relted variables and parameter variables. A partial solution has been implemented  but tests have shown no improvement at all due to the overhead of adding sorted keys etc. So this has to stay as it is now.
4. In passing a number of other little improvements have been acted:

- parser optimization in primary_expr()
- test of wheter a variable in a loop may be the loop index, and warning if it is modified inside the loop was slowing the execution due to the name-to-value search discussed above: the test is removed, unless the new `--warn-loop` switch is set when calling gdl. 
- crashes with `.RESET` or` .RESET_SESSION` are cured.
- Idem for `WIDGET_CONTROL,/RESET`
- operators overload and internal objects definition (overload.cpp) has been reviewed and simplified. LIST and related now are inheriting correct obejcts (CONTAINER). Undocumented (but useful) List::Set has been added.
- uninitialized elements in the plplot wxwidget driver have been set to 0, removing the annoying warnings when using Valgrind. Possibly more such warnings could be cured similarly.
- #2212 cured

### DIFF
--- a/scripts/build_gdl.sh
+++ b/scripts/build_gdl.sh
@@ -187,9 +187,14 @@ function build_msys2_package {
 }
 
 function find_architecture {
+
     if [ ${Platform:-`uname -m`} == "x86_64" ]; then
+		declare -x MSYSTEM="UCRT64"
+		declare -x MSYSTEM_CARCH="x86_64"
+		declare -x MSYSTEM_CHOST="x86_64-w64-mingw32"
+		declare -x MSYSTEM_PREFIX="/ucrt64"
         export mname="mingw64"
-        export MSYSTEM="MINGW64"
+#        export MSYSTEM="MINGW64"
         export arch="x86_64"
     else
         export mname="mingw32"
--- a/scripts/build_gdl.sh
+++ b/scripts/build_gdl.sh
@@ -187,14 +187,9 @@ function build_msys2_package {
 }
 
 function find_architecture {
-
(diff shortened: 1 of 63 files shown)


GillesDuvert:
victory: reference version: all tests passed on all platforms.
But we keep in mind #2200 #2177 #2176

GillesDuvert:
successful Windows build is obtained by removing the cache.
This should be done each time a build fails for non-gdl reasons.

Issue #1643, opened 2023-09-21
regression in 1.0.3 with the wxwidget driver when oplot ?


alaingdl:
On my laptop, Linux U22.04 up-to-date, with GDL compiled and install with the script, when running the following code, only the *last* red line is drawn. No problem with the X11 driver.

```
alaingdl@planck4:~/GDL$ more test_plot_loop.pro

pro TEST_PLOT_LOOP

x=findgen(10)

for ii=0, 4 do begin
   plot, ii+x
   oplot, 0.1+ii+x, thick=2, color=COLOR2COLOR(!color.red)
   wait, 1
endfor

end
```

(I realized because of new bench_gaussit.pro code)

GillesDuvert:
Typically a  synchro/refresh problem, only visible with wxWidgets due to the absence of a server running in a separate thread  or process (to be compared with X11, where Flush() is rarely needed).

GillesDuvert:
actually the problem was present since 1.0.1 . Should be solved now.

alaingdl:
Carefully tested on my laptop, unfortunately no changes :disappointed:

GillesDuvert:
rats!

alaingdl:
OK, we have a problem here :(

-- when on local, without this patch, the red line is NOT plotted
-- on remote (VPN), with or without `container->Update(); //solve 1643` the red line is plotted, but with this line the time need for the plot is about 3 times more !

PR #2226, opened 2026-08-29
improve largely graphic speed with wxwidgets while keeping #1643 and close #2199

### LINKED ISSUE #2199: gdl map_grid and map_continents painfully slow to x window
Map functions including map_continents seem to basically work well in the latest version - thanks to all involved in this.

I just notice that when running interactively to an x-window the drawing of the continent shapes is painfully slow (also grid lines are slow but not so many of them).

E.g. Just doing this...
map_set,_EXTRA=extra
map_grid ; this takes ~2s
map_continents ; this takes ~62s (!)

I also notice that this is all done very quickly if a different device is used (e.g. ps) or if the window is simply iconised while map_continents is run.
The improvement is so significant that I've modifed map_continents to (if !d.name eq 'X') iconise the window, then draw the lines, the re-display the window. Like that it takes 0.2s to draw the map.

However, I suspect there may be a better way to do this, without having to iconise/reshow the window.
E.g. is it possible plplot actually works via a buffer, such that it might be possible to hold off what is shown in the xwindow until all the lines have been drawn to the buffer ?

GillesDuvert:


alaingdl:
I tested it on a remote computer (ssh -X), and it is not OK for the graphics : if I just do GDL> tv, dist(120)
The window does appear
The TV does appear
But after that I never gain back the prompt (and load of my local Xorg is about 100%)
(yes it is compiled with WX)

GillesDuvert:
> I tested it on a remote computer (ssh -X), and it is not OK for the graphics : if I just do GDL> tv, dist(120) The window does appear The TV does appear But after that I never gain back the prompt (and load of my local Xorg is about 100%) (yes it is compiled with WX)

what kind of computer?

This PR was tested OK in "direct" mode and through VNC on a OSX machine. No tests have been done on Windows (I've no such access) and obviously given the, I repeat, strong obsolescence of X11, no tests for the X11 plot version (gdl -X) via ssh -X.

But in this latter case (ssh -X + gdl -X) there was no code change related to the X device and the X driver in this PR, so your tests may not be directly related to this PR.

PR #2239, opened 2026-09-19
Fix licensing of `types.h` and `xdr.h`

### LINKED ISSUE #2235: [1.1.3] Inappropriate licenses for types.h and xdr.h
Both files have a license that is considered non-free. Specifically, they come with the following clause:

https://github.com/gnudatalanguage/gdl/blob/1d90439b42fa8550045cc5bc81656b5965ae16cc/src/xdr.h#L5-L8

The restriction to distribute them only as part of a project makes them non-free because it is a limitation of distribution. 

There exist versions of the file re-licensed (with Suns approval) to BSD, which could replace `types.h` and `xdr.h`, for example in [ChromiumOS](https://chromium.googlesource.com/chromiumos/third_party/ntirpc/+/refs/heads/duplex-10-clioq/tirpc/rpc), so in they could be replaced. They are only used on Win32.

This was reported ad [Debian#1147366](https://bugs.debian.org/1147366).

olebole:
This PR takes both files as BSD-3-Clause licensed from krb5, https://github.com/krb5/krb5 and adjusts them for the specifics of GDL.

In `types.h`, this is mainly removing the @@-macros (the file was taken from `types.hin` in krb5); for `xdr.h` the change is the inclusion of the `xdr_int64_t` and `xdr_int64_t` prototypes and the limitation of `MAX_NETOBJ_SZ` to 1024 (originally 2048).

Closes: #2235 

Could someone test this on win32?


GillesDuvert:
@olebole  thanks, but as you can see in the continuous integration on GH, the build-windows fails at the compilation level, confirming my comment that all versions of xdr implementations are not equivalent.

However, the BSD re-licensed xdr example you gave seems promising and I hope that including it as a header-only library (and adapt GDL to it, and confirm it works... etc) will suppress yet another dependency of GDL. Stay tuned.

olebole:
I accidently forgot to re-add the `caddr_t` typedef. After inserting it, the windows part compiles well.

